PR #2, opened 2024-06-19
Create extension with placeholder module

ebrahimebrahim:
Closes #1 

* Use extension wizard to start the extension
* Create a scripted module "OpenLIFU Home" to serve as the starting point for the custom app
* Update README

Still need to 
- [x] Test that this can be installed
- [x] Test that this can be bundled with a custom application with its main module set as home module
- [x] Update README with instructions for building

PR #5, opened 2024-07-02
Add python dependency install functionality

ebrahimebrahim:
Adds a button to check for and install python dependencies.

The function `check_and_install_python_requirements` can be used in the future to run the check at the time that it's needed, so that the user doesn't have to use that button.

- [x] Merge #2 before reviewing this as this is branched from there.
- [x] Merge https://github.com/OpenwaterHealth/OpenLIFU-python/pull/23 and then update the commit hash in `python-requirements.txt` before merging this.
- [x] Improve the method of checking whether `openlifu` exists (instead of attempting import, use importlib's `find_spec`)
- [x] Check (without prompting) upon module entry whether the dependencies exist. Make the button text change from "install" to "re-install" if they exist.

### DIFF
--- a/OpenLIFUHome/Resources/UI/OpenLIFUHome.ui
+++ b/OpenLIFUHome/Resources/UI/OpenLIFUHome.ui
@@ -188,9 +188,6 @@
       </item>
       <item row="1" column="0">
        <widget class="QPushButton" name="installPythonReqsButton">
-        <property name="text">
-         <string>Install Python Requirements</string>
-        </property>
        </widget>
       </item>
      </layout>
--- a/OpenLIFUHome/Resources/UI/OpenLIFUHome.ui
+++ b/OpenLIFUHome/Resources/UI/OpenLIFUHome.ui
@@ -186,6 +186,13 @@
         </property>
        </widget>
       </item>
+      <item row="1" column="0">
+       <widget class="QPushButton" name="installPythonReqsButton">
+        <property name="text">
+         <string>Install Python Requirements</string>
+        </property>
+       </widget>
+      </item>
      </layout>
     </widget>
    </item>
--- a/OpenLIFUHome/Resources/python-requirements.txt
+++ b/OpenLIFUHome/Resources/python-requirements.txt
@@ -1,0 +1,1 @@
+git+https://github.com/OpenwaterHealth/OpenLIFU-python.git@fe76a8d8ed4e0759a86c63e51651de47c40b0061
--- a/OpenLIFUHome/CMakeLists.txt
+++ b/OpenLIFUHome/CMakeLists.txt
@@ -9,6 +9,7 @@ set(MODULE_PYTHON_SCRIPTS
 set(MODULE_PYTHON_RESOURCES
   Resources/Icons/${MODULE_NAME}.png
   Resources/UI/${MODULE_NAME}.ui
+  Resources/python-requirements.txt
   )
 
 #-----------------------------------------------------------------------------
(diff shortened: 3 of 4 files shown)


OpenLIFUHome/Resources/python-requirements.txt
```diff
@@ -0,0 +1 @@
+git+https://github.com/OpenwaterHealth/OpenLIFU-python.git@5c0eb79a1d0ca3984436e5d0f6c4588f15ab29bc
```
ebrahimebrahim, line 1: Update this commit hash once https://github.com/OpenwaterHealth/OpenLIFU-python/pull/23 is merged

OpenLIFUHome/Resources/UI/OpenLIFUHome.ui
```diff
@@ -186,6 +186,13 @@
         </property>
        </widget>
       </item>
+      <item row="1" column="0">
+       <widget class="QPushButton" name="installPythonReqsButton">
```
sjh26, line 190: Did you intend to put this button inside of the "Advanced" tab.  I think it should be up near the top, since in the extension version, you will need to click this in order to use the modules.

One thing I have done in another app is to check for the installed packages when entering the module, and then use that result to show or set relevant text on the install button, and after the install the button is hidden or set with different text (re-install vs install)

To make the custom app look nice i think we want show/hide, with an additional re-install button in the advanced section.

ebrahimebrahim: Yes I meant to put it inside "Advanced". 

I didn't want to check for installed packages when entering the module, because sometimes the import of openlifu takes a moment (https://github.com/OpenwaterHealth/OpenLIFU-python/issues/8) 

I was imagining that whenever the user does something that requires the openlifu package, that would trigger the check and install function to get called. So that's why I am hiding away the button under "Advanced" -- because it wouldn't normally be needed.

I like the idea of changing the button text to "re-install" appropriately, but again I didn't want to attempt `import openlifu` unnecessarily when it can cause a moment of lag

What do you think?

I could change the way in which the check is done. Instead of checking by attempting `import openlifu`, I could check using pip itself, something like `slicer.util._executePythonModule("pip", ["show", "openlifu"])` ... IDK if there is a better way to check if something is installed

ebrahimebrahim: Yes I meant to put it inside "Advanced". 

I didn't want to check for installed packages when entering the module, because sometimes the import of openlifu takes a moment (https://github.com/OpenwaterHealth/OpenLIFU-python/issues/8) 

I was imagining that whenever the user does something that requires the openlifu package, that would trigger the check and install function to get called. So that's why I am hiding away the button under "Advanced" -- because it wouldn't normally be needed.

I like the idea of changing the button text to "re-install" appropriately, but again I didn't want to attempt `import openlifu` unnecessarily when it can cause a moment of lag

What do you think?

I could change the way in which the check is done. Instead of checking by attempting `import openlifu`, I could check using pip itself, something like `slicer.util._executePythonModule("pip", ["show", "openlifu"])` ... IDK if there is a better way to check if something is installed

sjh26: If you are planning to trigger the check and install through other actions, that fine and we can leave the button in Advanced.

For other methods to check if the module exists, maybe https://docs.python.org/3/library/importlib.html#importlib.util.find_spec  (from https://stackoverflow.com/questions/14050281/how-to-check-if-a-python-module-exists-without-importing-it)

ebrahimebrahim: Nice, this will be a better way. I will do this and request review again when it is ready

sjh26 (approved):
tested, lgtm

ebrahimebrahim:
(all that changed since the review was the commit hash of openlifu python)

PR #50, opened 2024-07-16
Add session loading UI

ebrahimebrahim:
Before merging this:
- [x] Make it so that when you click "Load" while a subject is selected, the item is also expanded in the tree view. (feedback from @peterhollender)
- [x] Complete and merge https://github.com/OpenwaterHealth/OpenLIFU-python/pull/81
- [x] Update the version of openlifu used here to one that includes those changes
- ~~[ ] Make sure targets are handled correctly -- I had misunderstood that the the target point stored in a session is actually the center and that the protocol's focal pattern will have be taken into account to determine the actual set of points that will be focused on.~~ (This will actually be handled by some upcoming refactoring)
- ~~[ ] Move the generic standalone utility functions (commented on [below](https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/50#discussion_r1700867971)) to an appropriate place.~~ (#55 will handle this)

### DIFF
--- a/OpenLIFUHome/OpenLIFUHome.py
+++ b/OpenLIFUHome/OpenLIFUHome.py
@@ -209,7 +209,20 @@ def get_xx2mm_scale_factor(length_unit:str) -> float:
     openlifu = import_openlifu_with_check()
     return openlifu.util.units.getsiscale(length_unit, 'distance') / openlifu.util.units.getsiscale('mm', 'distance')
 
-
+def linear_to_affine(matrix, translation=None):
+    """Convert linear 3x3 transform to an affine 4x4 with
+    the given translation vector (the default being no translation)"""
+    if translation is None:
+        translation = np.zeros(3)
+    if matrix.shape != (3, 3):
+        raise ValueError("The input numpy array must be of shape (3, 3).")
+    return np.concatenate(
+        [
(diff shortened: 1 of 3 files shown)


OpenLIFUHome/OpenLIFUHome.py
```diff
@@ -49,60 +57,11 @@ def __init__(self, parent):
...
         )
 
-        # Additional initialization step after application startup is complete
-        slicer.app.connect("startupCompleted()", registerSampleData)
-
-
 #
-# Register sample data sets in Sample Data module
+# General utilities that probably should go elsewhere?
+# TODO Move some of these functions to more suitable places, e.g. some need to be where all modules can access them.
```
ebrahimebrahim, line 62: This section here contains a whole bunch of generically useful function, many of which should really be accessible to all modules in this extension. Where is a good place to put them?

brad-t-moore: Let's follow up with @sjh26 but in the meantime this isn't a blocker for this PR.

brad-t-moore (approved):
LGTM.  I reviewed the code.  I'm working setting up a local dev environment to build and test, but for this PR I think we're good as-is.

PR #54, opened 2024-08-07
Protocol loading

ebrahimebrahim:
- [ ] Merge #50 as this is branched from it.


ebrahimebrahim:
Protocol configuration will be its own independent module and it is also quite a low priority because we can very much get by without it for the treatment workflow. Hence closing this for now -- we can refer back to it later as a a UI mockup suggesting how the protocol configuration module could be designed. Here is a screenshot of what this looked like for future reference:

image

PR #63, opened 2024-08-20
Create modules

sadhana-r:


OpenLIFULib/OpenLIFULib/__pycache__/__init__.cpython-39.pyc
ebrahimebrahim, line 1: Should not commit any pycache files. Feel free to add `__pycache__` to `.gitignore`

ebrahimebrahim (changes requested):
Can you clean up the commit history?
- None of my commits from #50 should be showing up duplicated here.
- Some of your commits show up twice.
- There is a merge commit (probably part of the reason some of the commits show up twice). Let's keep the history linear as much as possible.
- Typo in one of the commit messages "rcreated"
- Follow the format of using imperative mood and sentence case, just to keep consistency. So e.g. "removed template UI code from all modules" becomes "Remove template UI code from all modules".

Issue #56, opened 2024-08-14
Create all the modules


ebrahimebrahim:
Here are the modules we will have:
- OpenLIFU Home (already exists)
- OpenLIFU Data
- OpenLIFU Pre-Planning
- OpenLIFU Transducer Tracking
- OpenLIFU Sonication Planning
- OpenLIFU Sonication Control
- OpenLIFU Protocol Configuration

Here are the things to do:
- Create all the modules, as blank modules for now. We can just put a QLabel that says "module under construction" or something (some of them may remain under construction for a while)
- Move all the database and session loading stuff that is introduced in #50 into the OpenLIFU Data Module and out of OpenLIFUHome
- Put buttons in OpenLIFUHome that take you to all the other modules. We can group Data, pre-planning, transducer tracking, sonication planning, and sonication control into one set. Then Protocol configuration can be off by itself as a separate thing.

sadhana-r:
When building the Slicer extension on Windows, there is a 260 character OS "max path limit". Therefore, few of the the module folder names had to be adjusted:

- OpenLIFUTransducerTracking --> OpenLIFUTransducerTracker
- OpenLIFUSonicationPlanning -->  OpenLIFUSonicationPlanner
- OpenLIFUProtocolConfiguration --> OpenLIFUProtocolConfig

Note: The module names that show up in Slicer have not be adjusted.

ebrahimebrahim:
Ah I always forget about that max path limit thing... the rename seems good to me

> Note: The module names that show up in Slicer have not be adjusted.

What did you mean by this?

sadhana-r:
> > Note: The module names that show up in Slicer have not be adjusted.
> 
> What did you mean by this?

For example, within the Slicer UI, the module is still called "OpenLIFU Transducer Tracking" since that sounds nicer but the source code folders/files are called OpenLIFUTransducerTracker" to reduce the character count.

ebrahimebrahim:
Closed by #67

PR #64, opened 2024-08-20
Create modules squashed

### LINKED ISSUE #55: Move common utilities to a "lib" module
There is [a pile of utilities](https://github.com/OpenwaterHealth/SlicerOpenLIFU/blob/1ded5237e8221b63a8906dff091395fb212f35c2/OpenLIFUHome/OpenLIFUHome.py#L60-L225) that I put into the OpenLIFUHome module that should really go to a common location where they can be accessed by all SlicerOpenLIFU modules.

How to do this? Create an `OpenLIFULib` python module or package (not a slicer module), which can be imported in modules
The common functionality can go into something like `Modules/Scripted/OpenLIFULib/OpenLIFULib/__init__.py` to be imported by `OpenLIFUHome/OpenLIFUHome.py`

sadhana-r:
Closes #55 

Leaves a couple of items left to resolve in #56


.gitignore
```diff
@@ -1 +1,10 @@
 build/
+OpenLIFUHome/__pycache__
+OpenLIFULib/OpenLIFULib/__pycache__
+OpenLIFUPrePlanning/__pycache__
+OpenLIFUProtocolConfig/__pycache__
+OpenLIFUSonicationControl/__pycache__
+OpenLIFUSonicationPlanner/__pycache__
+OpenLIFUTransducerTracker/__pycache__
```
ebrahimebrahim, line 8: You can just add a line

```
__pycache__/
```

and that should cover all of them

ebrahimebrahim (approved):
One small change to the `.gitignore` and then you can go ahead and merge. Thank you!

OpenLIFUHome/OpenLIFUHome.py
```diff
@@ -24,6 +18,8 @@
     vtkMRMLTransformNode,
 )
 
+import OpenLIFULib
+
 if TYPE_CHECKING:
     import openlifu # This import is deferred to later runtime, but it is done here for IDE and static analysis purposes
     import openlifu.db
```
ebrahimebrahim, line 25: It will also be nice if you can add a copy of these lines to the top of OpenLIFULib, for the sake of the type checker

ebrahimebrahim: (I guess you don't need the `openlifu.db` part in `OpenLIFULib`)

PR #66, opened 2024-08-21
Add openlifu object list and protocol loading

### LINKED ISSUE #57: Add openlifu object list and protocol loading
Add the ability to load a protocol and populate it into a list of OpenLIFU objects that is maintained in the Slicer scene. This includes:

- a button "load protocol" in the OpenLIFUData module that opens a file browser and allows one to load a protocol from a json file
- a _list of loaded openlifu objects_: maintain a list of loaded openlifu objects in the parameter node of OpenLIFUHome or OpenLIFUData
- when protocol is loaded as an `openlifu.Protocol` object, that object is added to the list of loaded openlifu objects
- a listview in the OpenLIFUData module shows in some way what are the loaded objects

ebrahimebrahim:
Close #57 

- [x] update openlifu python requirement to the version that has the changes introduced  in https://github.com/OpenwaterHealth/OpenLIFU-python/pull/86


ebrahimebrahim:
@sadhana-r When testing this remember to use the button to re-install python requirements, since the version was updated to enable this functionality

sadhana-r (approved):
Tested out the different functionality and it looks good!

PR #69, opened 2024-08-26
Add functionality for loading volumes and fiducials (#59)

### LINKED ISSUE #59: Add button for adding data manually
We will use Slicer's "add data" dialog when we want to manually load a volume into the scene.

~~Add a button to the openlifu data module that raises the add data dialog.~~ (button is already added in #66)

This is truly an extraneous button because it would be disabled in the custom application (where guided mode is always enabled) and the usual slicer "add data" button is always there when you are using the extension. The point is to group this button alongside the "load transducer" and "load protocol" buttons in the OpenLIFUData module to convey that this is how you would load a volume.

If it's possible to raise the "add data" dialog in a way that only allows it to load volumes then that would be nice but it's not super necessary since this is just part of the manual workflow.

### LINKED ISSUE #68: Move non-home modules to subcategory
Move all the modules except for OpenLIFUHome to a submenu so that the OpenLIFU category doesn't look so crowded and you can easily find the home module

sadhana-r:
Close #68 
Close #59

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -10,6 +10,7 @@ from slicer.i18n import tr as _
 from slicer.i18n import translate
 from slicer.ScriptedLoadableModule import *
 from slicer.util import VTKObservationMixin
+from slicer import app
 
 from slicer.parameterNodeWrapper import parameterNodeWrapper, parameterNodeSerializer, Serializer, ValidatedSerializer, validators
 from slicer import (
@@ -200,6 +201,8 @@ class OpenLIFUDataWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         self.ui.loadedObjectsView.setColumnWidth(0, 200)
         self.ui.loadedObjectsView.setColumnWidth(1, 200)
         self.ui.loadProtocolButton.clicked.connect(self.onLoadProtocolPressed)
+        self.ui.loadVolumeButton.clicked.connect(self.onLoadVolumePressed)
+        self.ui.loadFiducialsButton.clicked.connect(self.onLoadFiducialsPressed)
         self.addObserver(self.logic.getParameterNode().parameterNode, vtk.vtkCommand.ModifiedEvent, self.onParameterNodeModified)
 
         # ====================================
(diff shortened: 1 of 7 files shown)


ebrahimebrahim:
Nice; let me know when it's ready for review! I tested it for fun just to try out `slicer.util.openAddFiducialDialog()` and that didn't seem to work -- I wonder what's up with that

sadhana-r:
The ioManager functions are bugged as they have not been updated to use the new file type name for Markups. On Windows no memory leaks are detected when loading volumes and loading fiducials (after working around the bugged function).

OpenLIFUData/OpenLIFUData.py
```diff
@@ -10,6 +10,7 @@
 from slicer.i18n import translate
 from slicer.ScriptedLoadableModule import *
 from slicer.util import VTKObservationMixin
+from slicer import app
```
ebrahimebrahim, line 13: unused import

OpenLIFUData/OpenLIFUData.py
```diff
@@ -273,6 +276,16 @@ def onLoadProtocolPressed(self) -> None:
...
             self.logic.load_protocol(filepath)
 
+    def onLoadVolumePressed(self) -> None:
+        """ Call slicer dialog to load volumes into the scene"""
+        return slicer.util.openAddVolumeDialog()
+    
+    def onLoadFiducialsPressed(self) -> None:
+        """ Call slicer dialog to load fiducials into the scene"""
+        ioManager = slicer.app.ioManager()
+        return ioManager.openDialog("MarkupsFile", slicer.qSlicerFileDialog.Read)
```
ebrahimebrahim, line 286: can you add a code comment here indicating that we should use `slicer.util.whateverItWasCalled` but that we are doing this workaround for such and such reason

ebrahimebrahim (changes requested):
Great, thank you!
A couple of tiny changes and then it should be good to go

(The memory leak that I'm experiencing with the volume loading function is something I'll grapple with for a bit and I'll open a separate bug report for that if it turns out to be a bug)

ebrahimebrahim (approved):
excellent, good to go now!

PR #70, opened 2024-08-27
Enable SlicerOpenLIFUTransducer loading (#58)

### LINKED ISSUE #58: Add ability to load transducer from file
Depends on #57

Add ability to load transducer from file. Includes:
- a `SlicerOpenLIFUTransducer` class that consists of an `openlifu.Transducer`, a `vtkMRMLModelNode`, and a `vtkMRMLTransformNode`
- button to load transducer from json file
- when transducer is loaded, create a `SlicerOpenLIFUTransducer` and add it to the list of loaded openlifu objects, such that it is displayed in the listview
- refactor the session loading functionality that was introduced in #50 such that it uses the transducer loading code that is introduced here, rather than duplicating transducer loading code

### LINKED ISSUE #65: Show volumes and fiducials in the loaded objects view
Depends on #57 

Volume nodes and fiducial nodes that are in the scene should show up in the listview of loaded openlifu objects.

Volumes will show up as having type "Volume" and fiducial nodes can show up as having type ~~"Target"~~ "Points"

This means something has to get called whenever a volume or fiducial node is added to the scene.
(something like `slicer.mrmlScene.AddObserver(slicer.vtkMRMLScene.NodeAddedEvent, ...)`, there is an example in the slicer script repository) in order to update the list.
And similarly when nodes are are removed from the scene.

A few questions we may want to answer:
- How do we want to treat fiducial nodes that contain more than one point in their point list? I think we can just treat them as possible "targets" for use in openlifu, but raise errors when there is not exactly one point in the list.
- Do we want to use fiducial nodes for anything else in the SlicerOpenLIFU world? E.g. there is a concept of "markers" perhaps to be used for registration. If we are planning to have those then maybe fiducial nodes can be described as "Point" rather than "Target"...

ebrahimebrahim:
Close #58 
Close #65

- [x] Enable serialization of transformers
- [x] Enable the loading of transducers from json file
- [x] Refactor existing transducer loading code under `load_session` to use this transducer loading code
- [x] Make loaded transducers show up in the loaded objects view
- [x] Refactor session loading code to use the transducers in the loaded openlifu objects list rather than maintaining its own. (So there should be a way to add and remove objects from this list (by ID) that does the proper cleanup)
- [x] Make sure something reasonable happens when the user deletes one or both of a `SlicerOpenLIFUTransducer`'s nodes.
- [x] Invalidate session when session's transducer gets wiped out by manual deletion of an affiliated node
- [x] Move error display decorators to widget methods so they are more easily guaranteed to not repeat in a call stack
- [x] Look into why the M2 transducer loads with the wrong spatial scale
- [x] Move things to OpenLIFULib that I think would probably be used by other modules.

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -422,6 +422,14 @@ class OpenLIFUDataWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
                 [protocol.protocol.name, protocol.protocol.id, "Protocol"]
             ))
             self.loadedObjectsItemModel.appendRow(row)
+        for transducer_slicer in parameter_node.loaded_transducers.values():
+            transducer_slicer : SlicerOpenLIFUTransducer
+            transducer_openlifu : "openlifu.Transducer" = transducer_slicer.transducer.transducer
+            row = list(map(
+                OpenLIFULib.create_noneditable_QStandardItem,
+                [transducer_openlifu.name, transducer_openlifu.id, "Transducer"]
+            ))
+            self.loadedObjectsItemModel.appendRow(row)
 
     def onParameterNodeModified(self, caller, event) -> None:
         self.updateLoadedObjectsView()
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -11,30 +11,27 @@ from slicer.i18n import translate
 from slicer.ScriptedLoadableModule import *
 from slicer.util import VTKObservationMixin
 
-from slicer.parameterNodeWrapper import (
-    parameterNodeWrapper,
-    parameterNodeSerializer,
-    parameterPack,
-    Serializer,
-    ValidatedSerializer,
-    validators,
-)
+from slicer.parameterNodeWrapper import parameterNodeWrapper
 from slicer import (
     vtkMRMLScriptedModuleNode,
     vtkMRMLScalarVolumeNode,
     vtkMRMLMarkupsFiducialNode,
-    vtkMRMLModelNode,
-    vtkMRMLTransformNode,
 )
 
-import OpenLIFULib
 from OpenLIFULib import (
-    import_openlifu_with_check as openlifu_lz, # "openlifu_lz" stands for "openlifu lazy import"
+    openlifu_lz,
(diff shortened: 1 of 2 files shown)


ebrahimebrahim:
Here are some things you can check for when reviewing this:

- Load a couple of transducers (there are two in the example data) by using "load transducer" and browsing to the json file. It should show up in the view of loaded objects.
- Try manually deleting a model node or transform node of a transducer. It should realize that the transducer is now invalid, show you a message, and remove the transducer from the list of loaded transducers. It's okay if some orphaned nodes are left in the slicer scene, it's just no longer recognized as an existing SlicerOpenLIFU packet.
- Try loading a transducer that was already loaded. It should prompt if you want to replace it.
- Load a session. The associated transducer should load and show up in the list of loaded transducers.
- Try loading the example transducer, then load the example session (which has that same transducer). You should get a prompt indicating that the loaded transducer is about to get replaced
- Try loading the example session and then try loading the transducer associated with it manually -- it shouldn't let you.
- Try loading the example session, and then delete one of the transducer-related slicer nodes associated with that session. It should realize that the session became invalid.

sadhana-r:
The above tests worked. It's not clear what happens when a session is made "invalid". Shouldn't all associated objects be deleted? For example:

-  When a session becomes invalid (due to deleting of transducer), the volume and fiducial still remain loaded in the scene and show up in OpenLIFU objects - they should also be removed? If the user than re-loads the session, the volume and fiducials are duplicated as OpenLIFUObjects with the same name/different ID. However, adding the same volume through "Load Volume" updates the name (mni_{2}) and ID. This behaviour should be standardized. 
- Similarly, when a volume is deleted, the session becomes invalid but the transducer and target are still listed an "OpenLIFUobjects".  Loading the session after this, reloads the transducer (after prompting the user) but creates a duplicate target (same name/different ID). 
- Removing a fiducial/target does not invalidate/unload the session. 

Other thoughts:
- Can the Loaded OpenLIFU objects table can be made visible outside of the collabsible "OpenLIFU Objects" drop down. So that if a user only loads a session without wanting to manually select individual files, the table is still visible without having to scroll down. 
- In the Loaded OpenLIFUObjects table, it may be helpful to add a column "Session ID" to see which objects are associated with a session and which ones aren't.

sadhana-r (approved):

PR #71, opened 2024-08-29
Implement sonication planning controls

### LINKED ISSUE #60: Implement sonication planning module controls
Depends on #56 and #57.
Depends on #58 as well but could still get started without it.

Implement some of the functionality in OpenLIFUSonicationPlanning:
- populate dropdowns with protocol, transducer, volume, and target
- add a "plan" button that is just a dummy for now

### LINKED ISSUE #73: Put type column before ID
In the loaded objects view, put the _Type_ column before the _ID_ column.

sadhana-r:
Close #60
Close #73

### DIFF
--- a/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
+++ b/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
@@ -156,7 +156,7 @@ class OpenLIFUSonicationControlWidget(ScriptedLoadableModuleWidget, VTKObservati
             self._parameterNodeGuiTag = self._parameterNode.connectGui(self.ui)
 
     def onRunClicked(self):
-        print("Run sonication control")
+        print("Placeholder text: Running sonication")
 
         # Enable Abort button
         self.ui.abortPushButton.setEnabled(True)
--- a/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
+++ b/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
@@ -97,6 +97,10 @@ class OpenLIFUSonicationControlWidget(ScriptedLoadableModuleWidget, VTKObservati
         self.addObserver(slicer.mrmlScene, slicer.mrmlScene.EndCloseEvent, self.onSceneEndClose)
 
         # Buttons
+        # Disable the Abort button. Can only be clicked once the user hits 'Run'
+        self.ui.abortPushButton.setDisabled(True)
+        self.ui.runPushButton.clicked.connect(self.onRunClicked)
+    
     
         # Make sure parameter node is initialized (needed for module reload)
         self.initializeParameterNode()
@@ -150,6 +154,15 @@ class OpenLIFUSonicationControlWidget(ScriptedLoadableModuleWidget, VTKObservati
             # Note: in the .ui file, a Qt dynamic property called "SlicerParameterName" is set on each
             # ui element that needs connection.
             self._parameterNodeGuiTag = self._parameterNode.connectGui(self.ui)
+
+    def onRunClicked(self):
+        print("Run sonication control")
+
+        # Enable Abort button
+        self.ui.abortPushButton.setEnabled(True)
+        
+        #TODO: Once sonication control has been implemented,
+        # disbale 'Run' button during run and then disable 'Abort' again once completed. 
         
 # OpenLIFUSonicationControlLogic
 #
--- a/OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
+++ b/OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
@@ -184,21 +184,19 @@ class OpenLIFUSonicationPlannerWidget(ScriptedLoadableModuleWidget, VTKObservati
         if len(dataLogicParameterNode.loaded_protocols) == 0:
             self.ui.ProtocolComboBox.addItems(["Select a Protocol"])
         else:
-            for protocol in dataLogicParameterNode.loaded_protocols.keys():
+            for protocol in dataLogicParameterNode.loaded_protocols.values():
                 # Better to use ID or name here? Showing both for now
-                self.ui.ProtocolComboBox.addItems([protocol])
-                # self.ui.ProtocolComboBox.addItems(["{} (ID: {})".format(protocol.protocol.name,protocol.protocol.id)]) 
+                self.ui.ProtocolComboBox.addItems(["{} (ID: {})".format(protocol.protocol.name,protocol.protocol.id)]) 
     
         # Update transducer combo box
         self.ui.TransducerComboBox.clear()
         if len(dataLogicParameterNode.loaded_transducers) == 0:
             self.ui.TransducerComboBox.addItems(["Select a Transducer"]) 
         else:
-            for transducer in dataLogicParameterNode.loaded_transducers.keys():
-                # transducer_openlifu = transducer.transducer.transducer
+            for transducer in dataLogicParameterNode.loaded_transducers.values():
+                transducer_openlifu = transducer.transducer.transducer
                 # Better to use ID or name here? Showing both for now
-                self.ui.TransducerComboBox.addItems([transducer])
-                # self.ui.TransducerComboBox.addItems(["{} (ID: {})".format(transducer_openlifu.name,transducer_openlifu.id)]) 
+                self.ui.TransducerComboBox.addItems(["{} (ID: {})".format(transducer_openlifu.name,transducer_openlifu.id)]) 
         
 
     def onDataParameterNodeModified(self,caller, event) -> None:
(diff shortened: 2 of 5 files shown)


ebrahimebrahim:
```
git fetch
git rebase origin/main
```

should get your branch rebased to the new main branch -- LMK if there's any issues with that! I know rebase can get hairy, I can help

ebrahimebrahim:
for the commit swapping the columns I made an issue can be referenced for it: https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/73

ebrahimebrahim:
For those first three commits, can you add references to issue numbers?

IDK what is the best way to do it but I usually do an interactive rebase and set those commits to "reword" -- LMK if you'd like any help with that

OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
```diff
@@ -152,6 +171,94 @@ def setParameterNode(self, inputParameterNode: Optional[OpenLIFUSonicationPlanne
...
+    
+        # Update transducer combo box
+        self.ui.TransducerComboBox.clear()
+        if len(dataLogicParameterNode.loaded_transducers) == 0:
+            self.ui.TransducerComboBox.addItems(["Select a Transducer"]) 
+            self.ui.TransducerComboBox.setDisabled(True)
+        else:
+            self.ui.TransducerComboBox.setEnabled(True)
+            for transducer in dataLogicParameterNode.loaded_transducers.values():
+                transducer_openlifu = transducer.transducer.transducer
```
ebrahimebrahim, line 224: i am sorry you had to write transducer.transducer.transducer :)

OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
```diff
@@ -152,6 +171,94 @@ def setParameterNode(self, inputParameterNode: Optional[OpenLIFUSonicationPlanne
...
+    
+
+    def checkCanPlan(self, caller = None, event = None) -> None:
+
+        # If all the needed objects/nodes are loaded within the Slicer scene, all of the combo boxes will be enabled
+        # This means that the plan button can be enabled
+        if self.ui.TargetComboBox.enabled and self.ui.ProtocolComboBox.enabled and self.ui.VolumeComboBox.enabled and self.ui.TargetComboBox.enabled:
+            self.ui.PlanPushButton.enabled = True
+        else:
+            self.ui.PlanPushButton.enabled = False
```
ebrahimebrahim, line 185: when the plan button becomes disabled can we set a tooltip text on it explaining why it is disabled / what the user would have to do to enable it? and unset that tooltip (or set it to something else that is descriptive of what pressing the button would do) when the button becomes enabled again

OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
```diff
@@ -150,6 +154,15 @@ def setParameterNode(self, inputParameterNode: Optional[OpenLIFUSonicationContro
             # Note: in the .ui file, a Qt dynamic property called "SlicerParameterName" is set on each
             # ui element that needs connection.
             self._parameterNodeGuiTag = self._parameterNode.connectGui(self.ui)
+
+    def onRunClicked(self):
+        print("Run sonication control")
```
ebrahimebrahim, line 159: it should not say "control"

maybe also would be better to say "placeholder" somehwere in the printed text

OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
```diff
@@ -55,6 +59,9 @@ class OpenLIFUSonicationPlannerParameterNode:
 
     """
 
+    activeVolume: vtkMRMLScalarVolumeNode
+    activeTarget: vtkMRMLMarkupsFiducialNode
```
ebrahimebrahim, line 63: are these parameters getting used anywhere? i didn't see them getting used. if not then we should remove them

(i think we probably do not need to use the parameter node for this module at all)

OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
```diff
@@ -152,6 +171,94 @@ def setParameterNode(self, inputParameterNode: Optional[OpenLIFUSonicationPlanne
...
+        self.updateComboBoxOptions()
+
+    def onPlanClicked(self):
+        print("Run sonication planning")
+
+        dataLogicParameterNode = self.OpenLIFUDataLogic.getParameterNode()
+        activeTransducer = list(dataLogicParameterNode.loaded_transducers.values())[self.ui.TransducerComboBox.currentIndex]
+        activeProtocol = list(dataLogicParameterNode.loaded_protocols.values())[self.ui.ProtocolComboBox.currentIndex]
+        activeVolume = list(slicer.util.getNodesByClass('vtkMRMLScalarVolumeNode'))[self.ui.VolumeComboBox.currentIndex]
+        activeTarget = list(slicer.util.getNodesByClass('vtkMRMLMarkupsFiducialNode'))[self.ui.TargetComboBox.currentIndex]
```
ebrahimebrahim, line 258: here it looks like we are relying on the assumption that `slicer.util.getNodesByClass` will always return the nodes listed in the same order. it's probably true, but this is not an official guarantee so we probably should not do that. better to maintain a mapping between QComboBox indices and mrml nodes. it could be an object of type `Dict[int,slicer.vtkMRMLNode]` that we maintain for each combobox and re-set on every `updateComboBoxOptions`.

then if we do that for volumes and targets we might as well also do it for transducers and protocols.

so `OpenLIFUSonicationPlannerWidget` would maintain some attributes1 like
```
transducer_dict : Dict[int,SlicerOpenLIFUTransducer]
protocol_dict : Dict[int,SlicerOpenLIFUProtocol]
volume_dict : Dict[int,vtkMRMLScalarVolumeNode]
target_dict : Dict[int,vtkMRMLMarkupsFiducialNode]
```

this is just a suggestion but some way or other we should stop assuming that `slicer.util.getNodesByClass` will always return nodes in the same order.

a different approach could be to use the fact that QComboBox can hold "data" in the items: https://doc.qt.io/qt-6/qcombobox.html#currentData-prop
So maybe you could stick the actual nodes in there as the data ... this would be easier if it works

OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
```diff
@@ -152,6 +171,94 @@ def setParameterNode(self, inputParameterNode: Optional[OpenLIFUSonicationPlanne
             # Note: in the .ui file, a Qt dynamic property called "SlicerParameterName" is set on each
             # ui element that needs connection.
             self._parameterNodeGuiTag = self._parameterNode.connectGui(self.ui)
+            self.addObserver(self._parameterNode, vtk.vtkCommand.ModifiedEvent, self.checkCanPlan)
```
ebrahimebrahim, line 174: is this observer needed?

OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
```diff
@@ -152,6 +171,94 @@ def setParameterNode(self, inputParameterNode: Optional[OpenLIFUSonicationPlanne
...
+        self.checkCanPlan()
```
ebrahimebrahim, line 175: does this belong in `setParameterNode` or up in `setup`?

ebrahimebrahim (changes requested):
Very nice work! Left some comments and questions

OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
sadhana-r: I modified the code to store the data within the dropdown items

OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
sadhana-r: i moved it to `setpu` and also call it after `updatecombobox`

OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
```diff
@@ -152,6 +169,93 @@ def setParameterNode(self, inputParameterNode: Optional[OpenLIFUSonicationPlanne
...
+            self.ui.VolumeComboBox.setDisabled(True)
+        else:
+            self.ui.VolumeComboBox.setEnabled(True)
+            for volume_node in slicer.util.getNodesByClass('vtkMRMLScalarVolumeNode'):
+                self.ui.VolumeComboBox.addItem("{} (ID: {})".format(volume_node.GetName(),volume_node.GetID()), volume_node)
+          
+        # Update target combo box 
+        self.ui.TargetComboBox.clear()  
+        if len(slicer.util.getNodesByClass('vtkMRMLMarkupsFiducialNode')) == 0:
+            self.ui.TargetComboBox.addItem(["Select a Target"])
```
ebrahimebrahim, line 236: this currently shows up as a string represetnation of the list
image

ebrahimebrahim (changes requested):
Looks good and nice clean code

One small thing left

OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
sadhana-r: fixed!

ebrahimebrahim (approved):

PR #104, opened 2024-09-18
Enable new subject creatiion

sadhana-r:


OpenLIFUData/OpenLIFUData.py
```diff
@@ -736,6 +813,30 @@ def clear_plan(self,  clean_up_scene:bool = True) -> None:
...
+        """ Adds new subject to loaded openlifu database.
+
+        Args:
+            subject_name: name of subject to be added (str)
+            subject_id: id of subject to be added (str)
+        """
+
+        newOpenLIFUSubject = openlifu_lz().db.subject.Subject()
+        newOpenLIFUSubject.name = subject_name
+        newOpenLIFUSubject.id = subject_id
```
ebrahimebrahim, line 828: not too important but you should be able to construct it in one go:
```suggestion
        newOpenLIFUSubject = openlifu_lz().db.subject.Subject(
            name = subject_name,
            id = subject_id,
        )
```

OpenLIFUData/Resources/UI/OpenLIFUData.ui
```diff
@@ -26,12 +26,12 @@
       </item>
       <item>
        <widget class="ctkPathLineEdit" name="databaseDirectoryLineEdit">
+        <property name="filters">
+         <set>ctkPathLineEdit::Dirs</set>
```
ebrahimebrahim, line 30: I see you are turning the `<string>` tag back to `<set>`

See 97c134dc1407e7ac9c8aa4557ebad9fb6b92a7c2 where I changed it to `<string>`

I'm not fully sure what's going on with this, maybe it's automatically changed by your version of Qt designer? My version of Qt designer seems to hate the `<set>` and ends up automatically dropping the important `ctkPathLineEdit::Dirs` ... not sure what is up with that. Did you Qt designer complain at all about `<string>` or did it just silently modify the tag?

OpenLIFUData/OpenLIFUData.py
```diff
@@ -736,6 +813,30 @@ def clear_plan(self,  clean_up_scene:bool = True) -> None:
...
+        subject_ids = self.db.get_subject_ids()
+
+        if newOpenLIFUSubject.id in subject_ids:
+            if not slicer.util.confirmYesNoDisplay(
+                f"Subject with ID {newOpenLIFUSubject.id} already exists in the database. Overwrite subject?",
+                "Subject already exists"
+            ):
+                return
+
+        self.db.write_subject(newOpenLIFUSubject, on_conflict = 'overwrite')
```
ebrahimebrahim, line 839: This doesn't work in the version of openlifu that is indicated in the `python-requirements.txt` file. You will need to update the version in `python-requirements.txt` so that installing the python requirements using the button in the home module makes everything work. (if you go with the latest I think it will become `on_conflict=OnConlfictOpts.OVERWRITE` or something like that... you will see)

OpenLIFUData/OpenLIFUData.py
```diff
@@ -216,6 +271,28 @@ def on_item_double_clicked(self, index : qt.QModelIndex):
         # Make sure parameter node is initialized (needed for module reload)
         self.initializeParameterNode()
 
+    @display_errors
+    def onAddNewSubjectClicked(self, checked:bool) -> None:
+
+        if not self.logic.db:
+            slicer.util.errorDisplay("Database needs to be loaded first")
```
ebrahimebrahim, line 278: Suggest disabling the button when there is no databse rather than allowing it to be pressed and displaying this error.

Similar to `update_sessionLoadButton_enabled`, you can have a `update_newSubjectButton_enabled` which checks logic.db and updates enabledness and sets a tooltip

`update_newSubjectButton_enabled` could be called at the end of `onLoadDatabaseClicked`

ebrahimebrahim (changes requested):
Really nice work, just some minor changes

I like how the QDialog class turned out!

OpenLIFUData/OpenLIFUData.py
```diff
@@ -73,6 +74,53 @@ class OpenLIFUDataParameterNode:
     loaded_transducers : "Dict[str,SlicerOpenLIFUTransducer]"
     loaded_plan : "Optional[SlicerOpenLIFUPlan]"
 
+
+class AddNewSubjectDialog(qt.QDialog):
+    """ Add new subject dialog """
+
+    def __init__(self, database, parent="mainWindow"):
```
ebrahimebrahim, line 81: oh, also now you should be able to remove `database` as an argument here!

OpenLIFUData/Resources/UI/OpenLIFUData.ui
sadhana-r: My Qt designer must have automatically modified this cause I didn't explicitly edit this or see any errors. I'll change it back to '<string>' for now and keep an eye out in future updates to see if this happens again.

OpenLIFUData/OpenLIFUData.py
sadhana-r: thanks for catching this!

ebrahimebrahim (approved):
perfect, ty!!

PR #107, opened 2024-09-25
Refactor session handling

### LINKED ISSUE #77: Refactor session handling to use a "SlicerOpenLIFUSession"
Make SlicerOpenLIFUSession concept and refactor existing session loading functionality to it

Once a session is loaded, it adds to the list of loaded openlifu objects a SlicerOpenLIFUSession. At most one can be there at a time.

A SlicerOpenLIFUSession consists of an openlifu.Session, an associated SlicerOpenLIFUTransducer, volume node, and list of fiducial nodes.

[Relevant slide](https://docs.google.com/presentation/d/1yH7Nh0XqOb4q_BCZdTR4L2R6TGPShxaMDx2XHCmvxSg/edit#slide=id.g2f2d4b7456b_0_238)

### LINKED ISSUE #106: loaded objects view doesn't update at data module widget setup
Bug can be witnessed by getting an openlifu object loaded into the scene before opening the openlifu data module (e.g. by loading a mrml scene that had an openlifu object in it) and then opening the openlifu data module.

### LINKED ISSUE #109: Load protocol alongside session
After the change https://github.com/OpenwaterHealth/OpenLIFU-python/issues/93 we can now load a session's protocol when we load a session.

The protocol should also be considered to be session-affiliated information, so removing it would cause the active session to become invalid.

ebrahimebrahim:
Close #106
Close #109 
Close #77 

The session should now be an object that shows up in the loaded objects view

There can be at most one loaded session at a time

Protocols are now part of session and get loaded alongside them

When there is an active session, the session "owns" certain things: a volume, ~~some targets~~, a transducer, a protocol. When a session is unloaded in order for a new one to be loaded, those things are cleared out. 

This should work with `db-extended-example-v04.zip` in [our data folder](https://drive.google.com/drive/folders/1bldoP9VxE9SHyAB0bVAJ2l6WUuTxz3Yu).

---

Remaining tasks after PR review:

- [x] Expose the "clean up scene" option when informing the user that the session is being unloaded.

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -704,26 +704,12 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
 
         self.clear_session()
 
-        # Identify volume path. This will be simply db.get_volume_filepath after https://github.com/OpenwaterHealth/OpenLIFU-python/pull/123
-        volume_filename_maybe = Path(self.db.get_volume_filename(subject_id, session_openlifu.volume_id))
-        volume_file_candidates = volume_filename_maybe.parent.glob(
-            volume_filename_maybe.name.split('.')[0] + '.*'
-        )
-        volume_files = [
-            volume_path
-            for volume_path in volume_file_candidates
-            if slicer.app.coreIOManager().fileType(volume_path) == 'VolumeFile'
-        ]
-        if len(volume_files) < 1:
-            raise FileNotFoundError(f"Could not find a volume file for subject {subject_id}, session {session_id}.")
-        if len(volume_files) > 1:
-            raise FileNotFoundError(f"Found multiple candidate volume files for subject {subject_id}, session {session_id}.")
-        volume_path = volume_files[0]
+        volume_info = self.db.get_volume_info(session_openlifu.subject_id, session_openlifu.volume_id)
 
         # Create the SlicerOpenLIFU session object; this handles loading volume and targets
         new_session = SlicerOpenLIFUSession.initialize_from_openlifu_session(
             session_openlifu,
-            volume_path,
+            volume_path = volume_info["data_abspath"],
         )
 
         # === Load transducer ===
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -614,6 +614,8 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
 
     def get_subject(self, subject_id:str) -> "openlifu.db.subject.Subject":
         """Get the Subject with a given ID"""
+        if self.db is None:
+            raise RuntimeError("Unable to fetch subject info because there is no loaded database.")
         try:
             return self._subjects[subject_id] # use the in-memory Subject if it is in memory
         except KeyError:
@@ -622,6 +624,8 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
 
     def get_sessions(self, subject_id:str) -> "List[openlifu.db.session.Session]":
         """Get the collection of Sessions associated with a given subject ID"""
+        if self.db is None:
+            raise RuntimeError("Unable to fetch session info because there is no loaded database.")
         return [
             self.db.load_session(
                 self.get_subject(subject_id),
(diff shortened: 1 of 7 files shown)


ebrahimebrahim:
(I just rebased to main after merging #104)

sadhana-r:
Session object loading/unloading works great!

> When there is an active session, the session "owns" certain things: a volume, some targets, a transducer, a protocol. When a session is unloaded in order for a new one to be loaded, those things are cleared out.
> 

A few clarifications/questions:

- When I load a session and then delete a transducer or volume, the session gets unloaded.  But the sessions' volumes and points still show up in the loaded objects view and data module. If the session "owns" them, shouldn't they get removed?
- Deleting targets does not invalidate/unload a session
- Do we plan to have protocols show up in the Slicer data module eventually? Right now, if the loaded objects gets crowded fron unwated objects from previously unloaded objects (assuming the current functionality of point 1 is correct) then objects can be manually deleted from the data module. But theres no way to unload/remove a protocol.

sadhana-r:
Another thing I noticed is that when a session is first loaded, the volume has a unique name (*_1) whereas the targets don't. And if I keep switching between sessions, the unique name increases by 2 everytime. (*_1, *_3, *_5) instead of increasing by one like the targets. This numbering should be consistent. 
Also, can the naming be changed to only generate a unique name if a node with the same name already exists in the scene? Otherwise the _{x} number can get very large for volumes and targets just by switching between scenes even though the base names are unique to the scene.

sadhana-r (commented):
A few clarifications/suggestions:
-  When a session is unloaded (when a target or volume is deleted), the session's objects do not get unloaded from the scene. If the session "owns" them, I would expect those objects to get removed as well. I understand that in manual mode its okay to have objects orphaned but it would be nice if the "clean_up_scene" flag option be exposed to the user when being informed that a session is being unloaded.
- Deleting points/targets does not invalidate a session. Is that the expected functionality?
- Do we plan to have protocols show up in the slicer data module as a parameter node? Right now, when session objects are orphaned, they can be cleared up if needed from the data module. But  there isn't an easy way to remove a protocol object if unwanted. 

Another thing noticed is that when a session is first loaded, the volume has a unique name (_1) whereas the targets don't. And if I keep switching between sessions, the unique name increases by 2 everytime. (_1, *_3, *_5) instead of increasing by one like the targets. This numbering should be consistent.
Also, can the naming be changed to only generate a unique name if a node with the same name already exists in the scene? Otherwise the _{x} number can get very large for volumes and targets just by switching between scenes even though the base names are unique to the scene.

ebrahimebrahim:
> When a session is unloaded (when a target or volume is deleted), the session's objects do not get unloaded from the scene. If the session "owns" them, I would expect those objects to get removed as well. I understand that in manual mode its okay to have objects orphaned but it would be nice if the "clean_up_scene" flag option be exposed to the user when being informed that a session is being unloaded.

This is a really good point, I agree that the best solution for clarity is to expose the "clean up scene" option when informing the user that the session is being unloaded.

> Deleting points/targets does not invalidate a session. Is that the expected functionality?

Yes, because it is okay for a session to have no targets.

> Do we plan to have protocols show up in the slicer data module as a parameter node? Right now, when session objects are orphaned, they can be cleared up if needed from the data module. But there isn't an easy way to remove a protocol object if unwanted.

There's no plan to make them show up in the data module. The loaded objects tree view where they currently show up is in a sense the SlicerOpenLIFU version of the data module, so if we add a way to delete protocols in the future that would be in here.

> Another thing noticed is that when a session is first loaded, the volume has a unique name (_1) whereas the targets don't. And if I keep switching between sessions, the unique name increases by 2 everytime. (_1, *_3, *_5) instead of increasing by one like the targets. This numbering should be consistent.
Also, can the naming be changed to only generate a unique name if a node with the same name already exists in the scene? Otherwise the _{x} number can get very large for volumes and targets just by switching between scenes even though the base names are unique to the scene.

Hmm so we're currently just using slicer's unique name function; not sure why it is adding 2 each time and why the number has to keep growing like that. Probably the easiest fix is to put in our own unique name creation function; I will look into that.

ebrahimebrahim:
I've exposed the option to clear scene data to clarify what goes on. The checkbox for whether to clear scene data has a tooltip that explains what it would do.

For the unique naming weirdness I opened #110 to track that task separately, so that I don't have to cover that in this PR.

ebrahimebrahim:
(I am re-thinking the way targets are treated -- I think you are right that there should be a clearer notion of which targets "belong" to a session. I might push more updates to adjust that and then I will re-request the review.)

ebrahimebrahim:
(Never mind! After thinking about it I think targets should not be owned by sessions; it complicates things down the line. Ready for review)

ebrahimebrahim:
(Or rather, the session does own the targets in a sense (because they are cleaned up when the session is cleaned up) but I don't think we need to be clear about which targets belong to the session at any given time. All fiducials in the scene are potential targets, and the ones that are listed in the `SlicerOpenLIFUSession.target_nodes` attribute are just the ones that are marked for cleanup when the session is cleaned up because they were loaded by loading the session)

sadhana-r (approved):
Great job with the session unloading dialog! EVerything looks good to me.

PR #111, opened 2024-10-01
Add session status view and enable saving/loading sessions

### LINKED ISSUE #80: Display session status
Make the "current session info" display in the OpenLIFU Data module. Show name, associated transducer and protocol and volume and target, or “no session” if no session is currently loaded.

[Relevant slide](https://docs.google.com/presentation/d/1yH7Nh0XqOb4q_BCZdTR4L2R6TGPShxaMDx2XHCmvxSg/edit#slide=id.g2f2d4b7456b_0_238)

### LINKED ISSUE #81: Enable saving and unloading sessions
Add button for saving the current session. (So we will have to have a function that updates the session’s transducer transform and the session’s targets list from the scene info.)

Add button for unloading the current session.

[Relevant slide](https://docs.google.com/presentation/d/1yH7Nh0XqOb4q_BCZdTR4L2R6TGPShxaMDx2XHCmvxSg/edit#slide=id.g2f2d4b7456b_0_238)

Depends on https://github.com/OpenwaterHealth/OpenLIFU-python/issues/94 and #77

### LINKED ISSUE #89: Enforce targets having only one point
Incorporate [this](https://apidocs.slicer.org/main/classvtkMRMLMarkupsFiducialNode.html#a8f8c1970e950c84815d770d39c547423) into the way SlicerOpenLIFU uses fiducial nodes to represent targets. Consider setting "type" in the loaded objects view to Points (or "Point list") or to Point depending on whether this is set to 1.

ebrahimebrahim:
- Adds a view of the session status
- Enables saving and loading of sessions
- Sets maximum number of points to 1 when creating a fiducial node from an openlifu `Point`.

Close #80 
Close #81
Close #89

- [x] Rebase to main after #107 is merged
- [x] Complete #81

### DIFF
--- a/OpenLIFUData/Resources/UI/OpenLIFUData.ui
+++ b/OpenLIFUData/Resources/UI/OpenLIFUData.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>337</width>
-    <height>687</height>
+    <height>718</height>
    </rect>
   </property>
   <layout class="QVBoxLayout" name="verticalLayout">
@@ -160,6 +160,13 @@
           </widget>
          </item>
         </layout>
+       </widget>
+      </item>
+      <item>
+       <widget class="QPushButton" name="saveSessionButton">
+        <property name="text">
+         <string>Save Session</string>
+        </property>
        </widget>
       </item>
       <item>
--- a/OpenLIFUData/Resources/UI/OpenLIFUData.ui
+++ b/OpenLIFUData/Resources/UI/OpenLIFUData.ui
@@ -7,14 +7,17 @@
     <x>0</x>
     <y>0</y>
     <width>337</width>
-    <height>749</height>
+    <height>648</height>
    </rect>
   </property>
   <layout class="QVBoxLayout" name="verticalLayout">
    <item>
-    <widget class="ctkCollapsibleButton" name="sessionsCollapsibleButton">
-     <property name="text">
+    <widget class="ctkCollapsibleButton" name="sessionsCollapsibleButton" native="true">
+     <property name="text" stdset="0">
       <string>Sessions</string>
+     </property>
+     <property name="collapsed" stdset="0">
+      <bool>false</bool>
      </property>
      <layout class="QVBoxLayout" name="verticalLayout_2">
       <item>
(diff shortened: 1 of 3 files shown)


ebrahimebrahim:
Things to check out:
- Session status view in the openlifu data module looks right
- You can unload a session
- You can save a session
- If you manipulate the position of a target and/or transducer and save the session, then when you load the session you should see that those states were in fact saved and can re-load back to the same state.
- If you create a new fiducial node and only add one point to it, then when you save a session that point will be included in the session (i.e. if you now unload the session it will be unloaded, if you re-load the session it will be re-loaded).

sadhana-r (commented):
Looks great - everything seems to work as expected! Only suggestion is that in the Active Session status, it may be helpful to also include the subject name  associated with the session incase multiple subjects are associated with the same session name/id. It would be easier to see that info there instead of having to reference the subject/session selector.

ebrahimebrahim:
> it may be helpful to also include the subject name associated with the session incase multiple subjects are associated with the same session name/id

great idea completely agree! I've added that and also included both subject and session ID in a  parenthetical notation

OpenLIFUData/Resources/UI/OpenLIFUData.ui
```diff
@@ -106,16 +106,16 @@
            <widget class="QWidget" name="activeSession">
             <layout class="QFormLayout" name="formLayout">
              <item row="0" column="0">
-              <widget class="QLabel" name="sessionStatusNameLabel">
+              <widget class="QLabel" name="sessionStatusSubjectNameIdLabel">
                <property name="text">
-                <string>Name:</string>
+                <string>Subject Name (ID)</string>
```
sadhana-r, line 111: Missing colon (:) at the end. Same with session name.

sadhana-r: And can the width of the label column be increased a bit to accomodate the longer label name.

ebrahimebrahim: On my screen the entire module widget seems to expand to whatever width it needs to in order for the session status view to fit into it. Maybe that's not the best behavior actually, because a really long subject name causes the entire  module widget to widen. Let me rethink this...

ebrahimebrahim:
@sadhana-r I added scrolling behavior to ensure that the status view can always be seen even when the names in it are very long. Can you check and let me know if it looks good before I merge?

sadhana-r:
Looks good! You can go ahead and merge

Issue #78, opened 2024-09-04
Deactivate planning manual controls during active session


ebrahimebrahim:
Some options in the planning module should be locked to certain values and unchangable whenever there is an active session. They would be determined by the session. If for example the ~~target~~ protocol needs to be changed then it should be done from the pre-planning module, because it would have to be changed _for the session_. 

[Relevant slide](https://docs.google.com/presentation/d/1yH7Nh0XqOb4q_BCZdTR4L2R6TGPShxaMDx2XHCmvxSg/edit#slide=id.g2f2d4b7456b_0_351)

Depends on #77

Issue #96, opened 2024-09-13
Adapt SlicerOpenLIFUSolution to represent openlifu.Solution


ebrahimebrahim:
Make SlicerOpenLIFUPlan (soon to be renamed SlicerOpenLIFUSolution) mirror or represent an openlifu Solution. It could have an openlifu.Solution underlying it, or not, but either way there should probably be methods that allow you to read an openlifu.Solution into a SlicerOpenLIFUSolution and vice versa.

We will also want to use the ported over version of `calc_solution` rather than our own call to `run_simulation` in SlicerOpenLIFU.

Depends on https://github.com/OpenwaterHealth/OpenLIFU-python/issues/126

ebrahimebrahim:
I think it makes sense to lump #95 into the same PR that does this.

PR #114, opened 2024-10-09
Fix certain planning controls based on active session (#78)

ebrahimebrahim:
When there is an active session, it should fix the planning controls for protocol, transducer, and volume

---

Remaining tasks after the first review:

- ~~[ ] When a session is unloaded after a user runs planning, `clear_session` needs to call `validate_plan` to remove the associated plan~~[on second thought no](https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/114#issuecomment-2403269952)
- ~~[ ] Plan/solution volumes should follow similar mechanics to other objects that occur in groups like session-related or transducer-related objects: when the group becomes invalid there should be a dialog that offers a checkbox to optionally clear out the entire group from the scene~~ [wait isn't this already done?](https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/114#issuecomment-2403229252)

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -670,6 +670,14 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
             self.clear_session(clean_up_scene=clean_up_scene)
             return False
 
+        # Check protocol is present
+        if not loaded_session.protocol_is_valid():
+            clean_up_scene = sessionInvalidatedDialogDisplay(
+                "The protocol that was in use by the active session is now missing. The session will be unloaded.",
+            )
+            self.clear_session(clean_up_scene=clean_up_scene)
+            return False
+
         return True
 
     def validate_plan(self) -> bool:
--- a/OpenLIFULib/OpenLIFULib/__init__.py
+++ b/OpenLIFULib/OpenLIFULib/__init__.py
@@ -383,12 +383,32 @@ class SlicerOpenLIFUSession:
         """Return whether this session's transducer is present in the list of loaded objects."""
         return self.get_transducer_id() in get_openlifu_data_parameter_node().loaded_transducers
 
+    def protocol_is_valid(self) -> bool:
+        """Return whether this session's protocol is present in the list of loaded objects."""
+        return self.get_protocol_id() in get_openlifu_data_parameter_node().loaded_protocols
+
     def volume_is_valid(self) -> bool:
         """Return whether this session's volume is present in the scene."""
         return (
             self.volume_node is not None
             and slicer.mrmlScene.GetNodeByID(self.volume_node.GetID()) is not None
         )
+
+    def get_transducer(self) -> SlicerOpenLIFUTransducer:
+        """Return the transducer associated with this session, from the  list of loaded transducers in the scene.
+
+        Does not check that the session is still valid and everything it needs is there in the scene; make sure to
+        check before using this.
+        """
+        return get_openlifu_data_parameter_node().loaded_transducers[self.get_transducer_id()]
+
+    def get_protocol(self) -> SlicerOpenLIFUProtocol:
+        """Return the protocol associated with this session, from the  list of loaded protocols in the scene.
+
+        Does not check that the session is still valid and everything it needs is there in the scene; make sure to
+        check before using this.
+        """
+        return get_openlifu_data_parameter_node().loaded_protocols[self.get_protocol_id()]
 
     def clear_volume_and_target_nodes(self) -> None:
         """Clear the session's affiliated volume and target nodes from the scene."""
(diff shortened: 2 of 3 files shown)


ebrahimebrahim:
This required changing `checkCanPlan` in `OpenLIFUSonicationPlanner`, since comboboxes can be disabled and still contain a valid selection now. Hopefully I didn't break it :slightly_smiling_face:

sadhana-r (commented):
This fix works as expected and doesn't seem to break anything! I had a few issues/questions that came up during my review that are not directly related to this issue. Let me know if you want to address any of these pointers in the current PR otherwise I can approve. 

- I think we can now "tag" the p_min and ita volumes so they do not show up in the sonication planning volume combobox. I don't see an issue tracking this currently. 
- When a session is unloaded after a user runs planning, I think `clear_session` needs to call `validate_plan` to remove the associated plan. Related to this, should removing the plan also remove the associated p_min/ita volumes? This doesn't seem to be the case in manual mode, i.e. when a volume associated with a plan is removed. If not, then maybe the p_min/ita volumes need to be named so that they can be associated with a specific volume ID in manual mode or session ID in treatment mode in the loaded objects view. 
- Sonication planning and pnp_rendering works for the example subject/session but something wonky happens with the OW2* subjects. The planning runs but takes a very long time, the python console shows: `[Python]   source.p has more time points than kgrid.Nt, remaining time points will not be used.`  and the rendering doesn't work correctly.

ebrahimebrahim:
> I think we can now "tag" the p_min and ita volumes so they do not show up in the sonication planning volume combobox. I don't see an issue tracking this currently.

Agreed, please feel free to open an issue about that. Last discussion I think we said it's not on the critical path because it only matters for the "manual" workflow, but I agree it's definitely a good "Misc" thing to do

> When a session is unloaded after a user runs planning, I think clear_session needs to call validate_plan to remove the associated plan.

Right... that is how it should work... I should do that in this PR

> Related to this, should removing the plan also remove the associated p_min/ita volumes? This doesn't seem to be the case in manual mode, i.e. when a volume associated with a plan is removed. If not, then maybe the p_min/ita volumes need to be named so that they can be associated with a specific volume ID in manual mode or session ID in treatment mode in the loaded objects view.

Good point, I think those volumes should follow similar mechanics to other objects that occur in groups like session-related or transducer-related objects: when the group becomes invalid there should be a dialog that offers a checkbox to optionally clear out the entire group from the scene. I can include that in this PR

> Sonication planning and pnp_rendering works for the example subject/session but something wonky happens with the OW2* subjects

Yes I see that wonkiness too (although I don't remember the error)  -- I have been putting off worrying about that because the [current planning code](https://github.com/OpenwaterHealth/SlicerOpenLIFU/blob/9b8ceeced2f4272571b60291f1501347e2d1beaf/OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py#L347-L368) will soon be replaced by a proper port of the analogous code from the matlab software once https://github.com/OpenwaterHealth/OpenLIFU-python/issues/126 is completed.

TY for the review!

ebrahimebrahim:
> Related to this, should removing the plan also remove the associated p_min/ita volumes? This doesn't seem to be the case in manual mode, i.e. when a volume associated with a plan is removed. If not, then maybe the p_min/ita volumes need to be named so that they can be associated with a specific volume ID in manual mode or session ID in treatment mode in the loaded objects view.

@sadhana-r Wait a sec, this should be [already done](https://github.com/OpenwaterHealth/SlicerOpenLIFU/blob/9b8ceeced2f4272571b60291f1501347e2d1beaf/OpenLIFUData/OpenLIFUData.py#L686-L691) in the main branch. When you delete one of the two volumes associated with a plan, it shows a dialog with a checkbox that you can use to clear the scene of the rest of the plan data

ebrahimebrahim:
> When a session is unloaded after a user runs planning, I think clear_session needs to call validate_plan to remove the associated plan.

Looking at this in more detail I don't think it's right to do this, because the [current definition of SlicerOpenLIFUPlan](https://github.com/OpenwaterHealth/SlicerOpenLIFU/blob/9b8ceeced2f4272571b60291f1501347e2d1beaf/OpenLIFULib/OpenLIFULib/__init__.py#L473) doesn't contain any reference or link to the active session. Once #96 is complete, `SlicerOpenLIFUPlan` will become `SlicerOpenLIFUSolution`, and it will reference more things, possibly including a session, then `validate_plan` will have to be extended to check those further things.

I think we actually do want plans/solutions to be able to exist independently of sessions, because we want them to be usable in the manual workflow.

ebrahimebrahim:
@sadhana-r I ended up creating tasks and then canceling them :slightly_smiling_face: 
Nothing has changed just awaiting further comments

sadhana-r (approved):
Ok sounds good! I was thinking that when a node or session used to create a plan are removed, the associated plan should be cleared and misinterepreted `validate_plan` as partly doing this for volume nodes. But since you're saying that sessions/plans are independent objects, my comments can be ignored!

PR #117, opened 2024-10-10
Build up pre-planning module

### LINKED ISSUE #82: Enable target selection
~~There should be a target selection button that raises a dialog and gives a choice between selecting an existing fiducial or placing a new target.~~ There should be a target management interface for placing, moving, and removing targets. ~~The final result should be to set the target of the active session.~~ (There is no such thing as "_the_ target of a session", though there [will be](https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/84) a notion of "the target for which a session has an approved virtual fit".) ~~Controls should be disabled if there is no active session.~~ If the user opts to place a new target then we should engage interactive target placement (the thing that slicer does when you want to add a markups module point).

~~[Relevant slide](https://docs.google.com/presentation/d/1yH7Nh0XqOb4q_BCZdTR4L2R6TGPShxaMDx2XHCmvxSg/edit#slide=id.g2f2d4b7456b_0_290)~~ [Revised slide!](https://docs.google.com/presentation/d/1yH7Nh0XqOb4q_BCZdTR4L2R6TGPShxaMDx2XHCmvxSg/edit#slide=id.g309a6484731_3_1)

Depends on #77

### LINKED ISSUE #83: Add target RAS controls
Add RAS position controls for the selected target, when a target is selected. If it's not hard to do in slicer, it would also be nice to add a button that allows interactive repositioning of a target.

[Relevant slide](https://docs.google.com/presentation/d/1yH7Nh0XqOb4q_BCZdTR4L2R6TGPShxaMDx2XHCmvxSg/edit#slide=id.g2f2d4b7456b_0_290)

~~The controls would be disabled if the active session doesn't have a target yet~~

Depends on #82

### LINKED ISSUE #84: Add (dummy) virtual fit button and approval
Add a virtual fit button that calls a placeholder function with the appropriate arguments.

Add approve button and have the approval info be recorded in the Session object.

~~[Relevant slide](https://docs.google.com/presentation/d/1yH7Nh0XqOb4q_BCZdTR4L2R6TGPShxaMDx2XHCmvxSg/edit#slide=id.g2f2d4b7456b_0_290)~~ [Revised slide](https://docs.google.com/presentation/d/1yH7Nh0XqOb4q_BCZdTR4L2R6TGPShxaMDx2XHCmvxSg/edit#slide=id.g309a6484731_3_1)

Depends on #82 
Depends on https://github.com/OpenwaterHealth/OpenLIFU-python/issues/95

## Virtual fit approval

**How approval happens:** When the virtual fit is approved by clicking the approval button, this writes to an `Optional[str]` attribute in the `openlifu.Session` called `virtual_fit_approval_for_target_id`. This attribute is `None` by default, which means no virtual fit is approved. And otherwise the value it takes on is the ID of the approved target. When you click "approve" in the pre-planning module, the in-memory session gets this attribute written to.

**How approval is revoked:**  If the target is moved or removed or the transducer is moved, then this approval is revoked in memory. (Transducer tracking would move the transducer which revokes the virtual fit approval just like any transducer movement would. That's fine because by the time you are doing transducer tracking you have already acted upon the result of the virtual fit and no longer care whether that's approved or not.)

**What approval affects:** Approval state can be displayed in the session status in the data module, and it can be displayed in the pre-planning module as just a status message "Virtual fit approved for target blah" or "no virtual fit approved". It can also be displayed in the planning module as "Virtual fit approved for target blah" or just no message if we're out of that state. So it's just useful information, it doesn't block anything ever.

### LINKED ISSUE #118: Include display name for protocol and transducer in status view

### LINKED ISSUE #120: Two OpenLIFUData widgets can get created if onNodeAdded is connected to a request to the widget representation
When the `widgetRepresentation` of a `qSlicerScriptedLoadableModule` is requested, either the existing widget is returned or if there isn't an existing one then it is created. If the `widgetRepresentation` is requested for the first time and starts the creation of a widget, and then requested again before the creation is completed, then this results in the creation of two widgets of a module. This has come up in SlicerOpenLIFU and led to some weird behaviors.

Here's how this came up in SlicerOpenLIFU:
- The pre-planning module has an observer listening to onNodeAdded to update some QComboBoxes. Updating those QComboBoxes requires getting some info from the parameter node of the data module, which is obtained via `slicer.util.getModuleLogic('OpenLIFUData')`, which internally calls `widgetRepresentation` on the data module. The important thing is this: onNodeAdded now triggers a request of the data module `widgetRepresentation`.
- Now as soon as any request for the data module parameter node is made, think about what will happen. The `widgetRepresentation` of the data module will be requested internally, this will result in a new node being added: OpenLIFUData. Before the creation of the widget is complete, an onNodeAdded signal is emitted and because of the connection described above this initiates a _second_ call to `widgetRepresentation`!

There is a simple workaround: take care to create the OpenLIFUData widget before making those problematic connections. But this bug was hard to figure out so I just wanted a record to look back to. Also I wonder if the workaround is really the best approach.

ebrahimebrahim:
Close #82
Close #83 
Close #84 
Close #120
Close #118 

---

before merge remember to:

- [x] Update the openlifu version once [this](https://github.com/OpenwaterHealth/OpenLIFU-python/issues/95) is done, replacing the WIP commit that currently does it.
- [x] probably better to wait for merge of #116 and rebase this, rather than making #116 rebase to here.

## Things done in this PR

- Pre-planning module is created with a target management UI and a virtual fitting UI.
- openlifu python is updated to a version that supports the needed approval states (here just virtual fit approval)
- The virtual fit algorithm inputs are identical to the sonication planning inputs. To avoid code duplication and make it easier to provide features to both sets of widgets, I made a standalone "algorithm inputs" widget that has the code in one place. That same widget class is instantiated in two places now: the planning module and the pre-planning module.
- One feature that was needed in the algorithm input widget is remembering the previous selection before clearing and repopulating the comboboxes. It is not nice to reset the user's selection every time the comboboxes might need to be updated.
- Session status view now shows display names of protocol and transducer

## Things to try out in order to observe and test the new functionality

- Creating fiducial nodes that have a single point in their point list should always populate those nodes into the algorithm input dropdown menus and the listview of targets in the preplanning module.
- The target management controls should work: placing and removing targets, moving them around, locking and unlocking them from movement by mouse, editing the RAS coords manually.
- The virtual fit button should be a place holder. It should only be enabled if the algorithm inputs are all selected.
- Algorithm inputs should be locked to session values when there is an active session (except for target). This applies to both the pre-planning and planning modules (it is now the same widget class)
- The the virtual fit approval button should become enabled while there is an active session.
- Clicking the virtual fit approval button should apply approval, and this positive approval state can be viewed in 3 places: session status view in the data module, and also messages in the pre-planning and planning modules.
- Any virtual fit approval should be revoked when the approved target is moved or removed or when the session's transducer is moved.

## remaining todo from review

- [x] enable/disable RAS control with the lock
- [x] format changes to status display


sadhana-r (changes requested):
Great work! A couple things: 
- When a target is locked, the target position inputs should be disabled. RIght now, the target can still be moved by changing the RAS coordinates even when locked. 
- When the status of virtual_fit_approval_for_target_id is changed, it may be helpful either have a shortcut button within pre-planning to save the session or a message box to prompt the user to save the session. Or another idea is that in the virtual fit approval status text shown in the data and planning modules, you can include a reminder to save the active session if the text doesn't match what is included in the database. Just a thought cause it seems like it would be easy to forget to go the data module and save the session afterwards. 
- Just confirming, a session can have multiple targets but virtual fit approval can only be given to one target?

ebrahimebrahim:
> Just confirming, a session can have multiple targets but virtual fit approval can only be given to one target?

Yes, for now -- Peter [has mentioned](https://github.com/OpenwaterHealth/OpenLIFU-python/pull/130#discussion_r1799981489) that eventually we do want to allow for multiple targets to be virtual-fitted-for all at once, e.g. a whole-head helmet-like transducer. I was thinking we can stick to just one target for now and make the change to multiple targets in the future, since we are still currently charting our way through how to handle a lot of things

ebrahimebrahim:
> When the status of virtual_fit_approval_for_target_id is changed, it may be helpful either have a shortcut button within pre-planning to save the session or a message box to prompt the user to save the session. Or another idea is that in the virtual fit approval status text shown in the data and planning modules, you can include a reminder to save the active session if the text doesn't match what is included in the database. Just a thought cause it seems like it would be easy to forget to go the data module and save the session afterwards.

Agreed totally. I think maybe this shouldn't be done in the current PR but be a standalone topic that we discuss separately-- the topic of how to inform the user the session was modified and not saved, and how and when to nudge them to save. It's kind of an overall application design thing we haven't talked through yet and I am not sure I see clearly how to approach it yet

Let me open a separate issue

sadhana-r (approved):

ebrahimebrahim:
Rebase note: Removed "Update openlifu to handle approvals (#84)" since #116 bumped openlifu to an even more recent version

PR #124, opened 2024-10-23
Solution revamp and approvals for solutions and tracking

ebrahimebrahim:
close #95, close #96, close #113

### Solutions

- "Treatment plan" is renamed to "sonication solution" throughout, and generally we have gone from "plan" to "solution". This includes the way it's displayed in the session status view.
- `SlicerOpenLIFUSolution` has been refactored so that it is now properly a representation of an underlying `openlifu.Solution`.
- Solution computation code is now relegated to openlifu (it is done in a test branch of openlifu-python for now, the real thing will come in https://github.com/OpenwaterHealth/OpenLIFU-python/issues/126
- When a solution is generated while there is an active session, it's immediately written to the database under that session. A solution can still be generated in manual workflow, in which case it just wouldn't be affiliated with a session and wouldn't get written to a database (there isn't even necessarily a database in manual workflow).
- When a solution is generated while there's an active session, that solution becomes "affiliated" to that session, in the sense that if the session is unloaded or reloaded then the solution is cleaned up along with it.

### Transducer tracking approval

- There is an approve button in the transducer tracking module, only enabled when there is active session.
- Pressing the button toggles approval of transducer tracking in the active session.
- The button acts as a toggle, pressing it turns it into an _unapprove_ button
- Transducer tracking approval is revoked when transducer is moved
- Transducer tracking approval status can be displayed
  - in the session status view
  - in the transducer tracking module
  - as an alarming red warning in the sonication planning module if there isn't approval, or a calm message if there is

### Solution approval

- A solution approval button is added to the sonication planning module. It should only be enabled if there's a solution. Clicking it updates the `approved` attribute of the underlying openlifu solution -- it does this in-memory and also potentially on disk (in the database). Whether it writes approval to disk depends on whether the active session is the one that was used to generate the solution being approved.
- When a solution is approved, the approve button becomes an _unapprove_ button, and follows the same rules regarding when it has the effect of writing to the database.
- Run sonication should only be possible when there is an _approved solution_.

### Sonication control updates

- There is now a placeholder "do nothing for 3 seconds" run function. This is a stand-in for the actual sonication control code that will come later. Having this placeholder allows for the next item:
- There is now a more detailed implementation of "Run" and "Abort". Clicking run should turn on a "running" state in the logic of the sonication control module. During this state, the run button is disabled and the abort button is enabled. When a run completes, the running state is turned off and the run button becomes enabled again, and the abort button becomes disabled again.

## Remaining todo

[...]

PR #125, opened 2024-11-01
Add SlicerOpenLIFURun class and implement sonication control module

### LINKED ISSUE #86: Create `SlicerOpenLIFURun`, writable to `Run`
Create SlicerOpenLIFURun class with basic stuff like success of the run, post-run comments, protocol/transducer/etc. that were used, etc. It should show up in the list of loaded objects in the scene.

Update: More specifics about the type of info that shows up in a `Run` [here](https://github.com/OpenwaterHealth/OpenLIFU-python/issues/96#issuecomment-2383202510). In order to represent the needed info, it may be sufficient to wrap and add a serializer for openlifu `Run`  and just lean on how the openlifu `Run` works, similar to how `SlicerOpenLIFUProtocol` just wraps `openlifu.Protocol` for example.

Include a function that writes out a `SlicerOpenLIFURun` to the openlifu database, including the session and protocol snapshots.

[Relevant slide](https://docs.google.com/presentation/d/1yH7Nh0XqOb4q_BCZdTR4L2R6TGPShxaMDx2XHCmvxSg/edit#slide=id.g2f2d4b7456b_0_398)

This issue is just about creating the function and making sure it works, not exposing it in the UI yet.

Depends on https://github.com/OpenwaterHealth/OpenLIFU-python/issues/96 ~~and #85~~

### LINKED ISSUE #87: Implement sonication control module
Add “run” and “abort” buttons, a progress bar, a placeholder function that gets called to “run the sonication”, and dialogs asking about success of the run and if there are any notes to write. Create a `SlicerOpenLIFURun` object at the end of the run, and write it to the database as an openlifu `Run`.

[Relevant slide](https://docs.google.com/presentation/d/1yH7Nh0XqOb4q_BCZdTR4L2R6TGPShxaMDx2XHCmvxSg/edit#slide=id.g2f2d4b7456b_0_398)

Depends on #86

sadhana-r:
Closes #86 
Closes #87 
- [x] Take user input from Save run dialog
- [x] Distinguish run ending due to completion vs abort signal
- [x] protocol and session snapshot
- [x] Add progress bar
- [x] Wait for #124 to be merged and then rebase this to the new main

There can only be one loaded_run at a time. Similar to solutions

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1407,7 +1407,6 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
             run_openlifu = run.run
             self.db.write_run(run_openlifu, session_openlifu)
             
-    @display_errors
     def add_subject_to_database(self, subject_name, subject_id):
         """ Adds new subject to loaded openlifu database.
 
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -79,6 +79,10 @@ class OpenLIFUDataParameterNode:
     loaded_solution : "Optional[SlicerOpenLIFUSolution]"
     loaded_session : "Optional[SlicerOpenLIFUSession]"
 
+#
+# OpenLIFUDataDialogs
+#
+
 class CreateNewSessionDialog(qt.QDialog):
     """ Create new session dialog """
 
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -24,6 +24,7 @@ from OpenLIFULib import (
     SlicerOpenLIFUProtocol,
     SlicerOpenLIFUTransducer,
     SlicerOpenLIFUSolution,
+    SlicerOpenLIFURun,
     SlicerOpenLIFUSession,
     get_target_candidates,
     assign_openlifu_metadata_to_volume_node,
@@ -78,6 +79,7 @@ class OpenLIFUDataParameterNode:
     loaded_transducers : "Dict[str,SlicerOpenLIFUTransducer]"
     loaded_solution : "Optional[SlicerOpenLIFUSolution]"
     loaded_session : "Optional[SlicerOpenLIFUSession]"
+    loaded_run: "Optional[SlicerOpenLIFURun]"
 
 #
 # OpenLIFUDataDialogs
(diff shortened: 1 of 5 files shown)


OpenLIFUData/OpenLIFUData.py
```diff
@@ -731,6 +737,15 @@ def updateLoadedObjectsView(self):
                 [solution_openlifu.name, "Solution", solution_openlifu.id]
             ))
             self.loadedObjectsItemModel.appendRow(row)
+        
+        if parameter_node.loaded_run is not None:
+            run_openlifu = parameter_node.loaded_run.run
+            row = list(map(
+                create_noneditable_QStandardItem,
+                [run_openlifu.id, "Run", run_openlifu.id]
```
ebrahimebrahim, line 745: Could include the run `name` in here

OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
```diff
@@ -53,6 +60,57 @@ class OpenLIFUSonicationControlParameterNode:
 
     """
 
+#
+# OpenLIFUSonicationControlDialogs
+#
+
+class onRunCompletedDialog(qt.QDialog):
```
ebrahimebrahim, line 67: Class names should us CamelCase starting with a captial so `OnRunCompletedDialog`

ebrahimebrahim (changes requested):
Clean and easy to read PR, and it works great!

I was initially confused trying to understand how the progress bar update was working, but after some thought I think it makes sense.

Besides the minor comments there are a couple of bigger things that come to mind that weren't in the original plan:

- If the run is aborted, I think we should still save a record that the run happened. When the run is aborted we can force the checkbox to "False" in the dialog because it can be considered not a success. 
- If the run was not successful, maybe we should force the description to be nonempty (this could be enforced at the level of `OnRunCompletedDialog`, so it doesn't let you submit the dialog with an empty description and a failed run).

What do you think?

sadhana-r:
> * If the run is aborted, I think we should still save a record that the run happened. When the run is aborted we can force the checkbox to "False" in the dialog because it can be considered not a success.
> * If the run was not successful, maybe we should force the description to be nonempty (this could be enforced at the level of `OnRunCompletedDialog`, so it doesn't let you submit the dialog with an empty description and a failed run).
> 
> What do you think?

I agree, those are good points. I am also wondering, since the run object is only created once a user selects 'Save' in the dialog, do you think we should add a second check inthe case that a user selects 'Cancel' when the dialog pops up. i..e. On cancel, a dialog pops up with "Are you sure you want to discard this run?". Since there is no other way to retrieve the run object if cancelled accidentally.

ebrahimebrahim:
> since the run object is only created once a user selects 'Save' in the dialog, do you think we should add a second check inthe case that a user selects 'Cancel' when the dialog pops up. i..e. On cancel, a dialog pops up with "Are you sure you want to discard this run?". Since there is no other way to retrieve the run object if cancelled accidentally.

Yes that makes sense. Or we could not even have a cancel button, I think both of those are acceptable ways to do it. I can't think of why someone would insist on not creating/saving  Run at the end.

ebrahimebrahim (approved):
Code changes look great and the abort and dialog-closing work as expected. TY!

ebrahimebrahim:
(Don't merge yet of course until the python library update is complete and we do the needed rebase! But we can consider this PR done and move on)

PR #131, opened 2024-11-08
Add guided mode status variable and toggle button 

### LINKED ISSUE #126: Add togglable guided mode variable to application logic
Add boolean variable `guided_mode` in the SlicerOpenLIFU home module parameter node. Add Button in SlicerOpenLIFU home module that toggles the state of it. There should be a “start guided mode” method in the home module that for now just puts you into the starting module of guided mode: the openlifu data module.

There should be a convenient functions or something in OpenLIFULib for determining whether we are in guided mode.

sadhana-r:
Closes #126

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -36,6 +36,8 @@ from OpenLIFULib.util import (
     add_slicer_log_handler,
 )
 
+from OpenLIFULib.guided_mode_util import get_guided_mode_status
+
 if TYPE_CHECKING:
     import openlifu # This import is deferred at runtime using openlifu_lz, but it is done here for IDE and static analysis purposes
     import openlifu.db
@@ -803,6 +805,9 @@ class OpenLIFUDataWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         """Called each time the user opens this module."""
         # Make sure parameter node exists and observed
         self.initializeParameterNode()
+
+        # Temp code to test that the data module is aware of guided mode when entering module
+        print("Setting up in Guided Mode?:", get_guided_mode_status())
 
     def exit(self) -> None:
         """Called each time the user opens a different module."""
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -36,7 +36,7 @@ from OpenLIFULib.util import (
     add_slicer_log_handler,
 )
 
-from OpenLIFULib.guided_mode_util import get_guided_mode_status
+from OpenLIFULib.guided_mode_util import get_guided_mode_state
 
 if TYPE_CHECKING:
     import openlifu # This import is deferred at runtime using openlifu_lz, but it is done here for IDE and static analysis purposes
@@ -805,9 +805,6 @@ class OpenLIFUDataWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         """Called each time the user opens this module."""
         # Make sure parameter node exists and observed
         self.initializeParameterNode()
-
-        # Temp code to test that the data module is aware of guided mode when entering module
-        print("Setting up in Guided Mode?:", get_guided_mode_status())
 
     def exit(self) -> None:
         """Called each time the user opens a different module."""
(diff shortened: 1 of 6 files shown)


OpenLIFUData/OpenLIFUData.py
```diff
@@ -804,6 +806,9 @@ def enter(self) -> None:
         # Make sure parameter node exists and observed
         self.initializeParameterNode()
 
+        # Temp code to test that the data module is aware of guided mode when entering module
+        print("Setting up in Guided Mode?:", get_guided_mode_status())
+
```
ebrahimebrahim, line 811: remember to get rid of this temporary code (but ty i benefited from having it while testing)

OpenLIFUHome/OpenLIFUHome.py
```diff
@@ -98,9 +96,15 @@ def setup(self) -> None:
         # Make sure parameter node is initialized (needed for module reload)
         self.initializeParameterNode()
 
+        # Set the initial guided mode status
+        self.logic.getParameterNode().guided_mode = False
+
```
ebrahimebrahim, line 101: Instead of initializing in setup, I think you should do
```py
@parameterNodeWrapper
class OpenLIFUHomeParameterNode:
    guided_mode : bool  = False
```
up above. (see [parameter node defaults docs](https://slicer.readthedocs.io/en/latest/developer_guide/parameter_nodes/defaults.html) -- if i understand correctly this will make it initialize to False)

One advantage of doing it that way is that whether or not we are in guided mode can persist over scene save and load. If guided mode is set to False in the module setup then it would be forced into false if you ever loaded a scene that was saved with guided mode enabled (I think)

OpenLIFUHome/OpenLIFUHome.py
```diff
@@ -185,7 +186,22 @@ def setParameterNode(self, inputParameterNode: Optional[OpenLIFUHomeParameterNod
             # ui element that needs connection.
             self._parameterNodeGuiTag = self._parameterNode.connectGui(self.ui)
 
+    def onGuidedModeClicked(self):
+        self.logic.toggle_guided_mode_status()
+        slicer.util.selectModule("OpenLIFUData")
```
ebrahimebrahim, line 191: A couple of things with putting `selectModule` here:
- this would switch us to the data module when we click to exit guided mode, which is a bit weird since the data module is really the _start_ of guided mode. clicking "exit" should keep us right where we are in the home module I think.
- the action of "starting guided mode" should be kind of a logic level thing. so for example if from the custom app we want to _start guided mode_ then there should be a way to do that, and we wouldn't do it by calling OpenLIFUHome widet's onGuidedModeClicked; we would probably do it by calling some method of the logic or some method in OpenLIFULib.guided_mode_util.

So I'm not sure what's the best way to design this but what I'm suggesting is that we have some reasonable way for the custom app to issue a python command that for sure turns on guided mode _and_ "starts" it, i.e. puts us into the OpenLIFUData module

OpenLIFUHome/OpenLIFUHome.py
```diff
@@ -212,6 +228,9 @@ def getParameterNode(self):
     def clear_session(self) -> None:
         self.current_session = None
     
+    def toggle_guided_mode_status(self):
```
ebrahimebrahim, line 231: perhaps `toggle_guided_mode_state` is better? since it's not a status display but rather internal application state

perhaps there could be an optional `bool` input for whether we want to set the state to on or off? in that case the name could be `set_guided_mode_state` and the docstring can explain that if not provided it acts as a toggle

the reason I suggest this is that the custom app wants to be 100% certain is turning _on_ guided mode, and calling this toggle feels a bit iffy

ebrahimebrahim (changes requested):
TY! A few changes requested

sadhana-r:
Thanks for the suggestions! I wasn't thinking about things from the custom app perspective so this was helpful! 
I added a util function to set the guided mode state. I kept the start function in the Home module's logic since the switch will mostly (always?) happen from the Home module. So now the custom app can just directly call the start_guided_mode function from the home widgets logic.

ebrahimebrahim (approved):
Works great and will work with https://github.com/OpenwaterHealth/OpenLIFU-app/issues/8 I think, TY!!

PR #135, opened 2024-11-13
Tag sonication planning outputs to not show up as algorithm inputs

### LINKED ISSUE #115: Solution volumes should not show up as volume inputs
Tag volumes output by sonication planning  so they don't show up in the volume combobox as potential planning inputs

### LINKED ISSUE #134: Adopt better parameter node conventions
> Looking at the template code for slicer modules recently, I realized that we should actually just use `self._parameterNode` in the widget class to access a module's parameterNode. This variable is set when initializeParameterNode is called. And technically, the observer's for the parameterNode should be added/removed in `setParameterNode`. You don't have to change this but just wanted to bring it up.

_Originally posted by @sadhana-r in https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/132#discussion_r1838567887_

### LINKED ISSUE #136: Fix outdated BusyCursor import in lazyimport
In lazyimport.py, the BusyCursor import neeeds to be updated after OpenLIFULib restructuring. This bug was introduced in #88

sadhana-r:
Closes #115
Closes #134 
Closes #136

- [x] Rebase to main after #132 merge
- [x] Update parameter node conventions in sonication planner

### DIFF
--- a/OpenLIFULib/OpenLIFULib/algorithm_input_widget.py
+++ b/OpenLIFULib/OpenLIFULib/algorithm_input_widget.py
@@ -104,7 +104,7 @@ class OpenLIFUAlgorithmInputWidget(qt.QWidget):
             self.volume_input.combo_box.setEnabled(True)
             for volume_node in slicer.util.getNodesByClass('vtkMRMLScalarVolumeNode'):
                 # Check that the volume is not an OpenLIFUSolution output volume
-                if not volume_node.GetAttribute('isOpenLIFUSolution'):
+                if volume_node.GetAttribute('isOpenLIFUSolution') is None:
                     self.add_volume_to_combobox(volume_node)
 
         self.set_session_related_combobox_tooltip("")
--- a/OpenLIFULib/OpenLIFULib/algorithm_input_widget.py
+++ b/OpenLIFULib/OpenLIFULib/algorithm_input_widget.py
@@ -103,7 +103,9 @@ class OpenLIFUAlgorithmInputWidget(qt.QWidget):
         else:
             self.volume_input.combo_box.setEnabled(True)
             for volume_node in slicer.util.getNodesByClass('vtkMRMLScalarVolumeNode'):
-                self.add_volume_to_combobox(volume_node)
+                # Check that the volume is not an OpenLIFUSolution output volume
+                if not volume_node.GetAttribute('isOpenLIFUSolution'):
+                    self.add_volume_to_combobox(volume_node)
 
         self.set_session_related_combobox_tooltip("")
 
--- a/OpenLIFULib/OpenLIFULib/lazyimport.py
+++ b/OpenLIFULib/OpenLIFULib/lazyimport.py
@@ -5,7 +5,7 @@ from pathlib import Path
 import slicer
 import importlib
 import sys
-from OpenLIFULib.busycursor import BusyCursor
+from OpenLIFULib.util import BusyCursor
 if TYPE_CHECKING:
     import openlifu # This import is deferred at runtime, but it is done here for IDE and static analysis purposes
     import xarray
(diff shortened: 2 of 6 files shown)


OpenLIFULib/OpenLIFULib/lazyimport.py
```diff
@@ -5,7 +5,7 @@
 import slicer
 import importlib
 import sys
-from OpenLIFULib.busycursor import BusyCursor
+from OpenLIFULib.util import BusyCursor
```
ebrahimebrahim, line 8: Huh, I wonder why this bug wasn't causing problems before

OpenLIFULib/OpenLIFULib/algorithm_input_widget.py
```diff
@@ -103,7 +103,9 @@ def _populate_from_loaded_objects(self) -> None:
         else:
             self.volume_input.combo_box.setEnabled(True)
             for volume_node in slicer.util.getNodesByClass('vtkMRMLScalarVolumeNode'):
-                self.add_volume_to_combobox(volume_node)
+                # Check that the volume is not an OpenLIFUSolution output volume
+                if not volume_node.GetAttribute('isOpenLIFUSolution'):
```
ebrahimebrahim, line 107:
```suggestion
                if volume_node.GetAttribute('isOpenLIFUSolution') is None:
```

Just for clarity about what's happening, since it's really `None` getting converted to `False` and `'True'` getting converted to `True`

OpenLIFULib/OpenLIFULib/solution.py
```diff
@@ -53,6 +53,10 @@ def initialize_from_openlifu_data(
         pnp_volume_node.GetDisplayNode().SetAndObserveColorNodeID("vtkMRMLColorTableNodeFilePlasma.txt")
         intensity_volume_node.GetDisplayNode().SetAndObserveColorNodeID("vtkMRMLColorTableNodeFilePlasma.txt")
 
+        # Set openlifu solution attribute
+        pnp_volume_node.SetAttribute('isOpenLIFUSolution', 'True')
+        intensity_volume_node.SetAttribute('isOpenLIFUSolution', 'True')
+
```
ebrahimebrahim, line 59: This is great! Can we also use these tags to omit the volume from the list of loaded openlifu objects displayed in the openlifu data module? This could also be considered part of #115.

ebrahimebrahim (changes requested):
This is great thank you!

Thanks for also addressing part of #134 here.
But I edited the description above so that #134 is not closed-- there is still the pre-planning module that needs to be fixed for #134. (feel free to do it and set #134 to close, but you don't have to close it)

OpenLIFULib/OpenLIFULib/lazyimport.py
sadhana-r: Yeah its weird. It was causing issues with the custom app build.

OpenLIFULib/OpenLIFULib/solution.py
ebrahimebrahim: My opinion:

I think they should be omitted because they are kind of built into the "Solution" which is displayed already as a loaded openlifu object

As slicer nodes they are their own thing independent of the Solution, sure

But as far as the openlifu world is concerned they are just there as a way to display something that is really in the Solution

Similar to the situation with trasnducer transform and model node -- they are two slicer nodes but in the openlifu world they are considered part of the one openlifu transducer

sadhana-r:
> Thanks for also addressing part of #134 here. But I edited the description above so that #134 is not closed-- there is still the pre-planning module that needs to be fixed for #134. (feel free to do it and set #134 to close, but you don't have to close it)

Did you mean the Home module? I don't see any reference to logic.GetParameterNode() in the preplanning module except for the setParameterNode function. I have updated the Home module now as well and checked the other modules. I think this issue can be closed now?

ebrahimebrahim:
> Did you mean the Home module? I don't see any reference to logic.GetParameterNode() in the preplanning module except for the setParameterNode function. I have updated the Home module now as well and checked the other modules. I think this issue can be closed now?

Oh, you are right, my mistake

ebrahimebrahim (approved):
Code looks great and works, thank you!

PR #141, opened 2024-11-19
Integrate virtual fitting

### LINKED ISSUE #139: Add virtual fitting UI
Virtual fitting API is already in place in this PR:
https://github.com/OpenwaterHealth/OpenLIFU-python/pull/162

The minimal implementation is getting filled in now, but the API is already there in the class `VirtualFit` so we just need to interface with that in SlicerOpenLIFU.

In this issue: 

- Figure out how to get the inputs into the constructor of `VirtualFit`. For example are they in `Protocol`? In `Transducer`? Or do they have to be brought in somehow?
- Make the virtual fit button use the placeholder class `VirtualFit` when the virtual fit button is pressed

### LINKED ISSUE #145: Update to the transfom-saving approach to virtual fitting.
Update SlicerOpenLIFU virtual fitting to work with the design introduced in https://github.com/OpenwaterHealth/OpenLIFU-python/issues/179 

This means
- After a virtual fit is complete, a new transform node is created representing the virtual fitted transducer transform. This happens regardless of whether there's an active session. It should be tagged with an attribute indicating which transducer ID it was meant for.
- Even though a new transform node was created, the transform node that is applied to the transducer model is still the "current transducer transform" that we already have. And this is still updated. So no change here.
- If there is an active session when running virtual fitting, then the virtual fit result (including the transform) is additionally written into the active session
- The session status view should be updated to show virtual fit related status
- The preplanning module should show virtual fit status as well

Yet to clarify: How should the new transform node be handled? Is it an "openlifu object" to be listed in the list of openlifu objects? Or is it an affiliated object of other things (like how the "current transducer transform node" is affiliated to the SlicerOpenLIFUTransducer)? What happens when you load a session that has virtual fitting info -- does it load the transform?

Depends on https://github.com/OpenwaterHealth/OpenLIFU-python/issues/179
Depends on #139

ebrahimebrahim:
Close #139 

This also close #145 finally, though most of the work to actually close that was in #179.

## Remaining TODO

- [x] Clean out the old WIP commit here which I think contains a function we no longer need
- [x] Clean up commits for final draft
- [x] Make the virtual fit button both update the transducer transform and populate in the virtual fit results
- [x] Add the virtual fit logger that shows which step we are on
- [x] Come up with good defaults to stick in for steering ranges and such until https://github.com/OpenwaterHealth/OpenLIFU-python/issues/165 and https://github.com/OpenwaterHealth/OpenLIFU-python/issues/166 are done
- [x] Put VF nodes under the correct parent folder.

## Outdated comments (ignore this section)

(The below comments are from before virtual fitting was as developed as it is now, I just don't want to delete the history.)

This is intended to be a minimal integration of the virtual fitting algorithm https://github.com/OpenwaterHealth/OpenLIFU-python/issues/147 into SlicerOpenLIFU. After this is done we can work on refining the virtual fitting "workflow" and all that, but here we just want to click "virtual fit", have the algorithm run, and then see the transducer position update.

Some development on openlifu-python is needed before this can be ready.

- [x] a thing I am wondering is where the skin surface is extracted. maybe the skin surface is assumed to be one of the inputs into the virtual fitting algorithm? Since we aren't doing https://github.com/OpenwaterHealth/OpenLIFU-python/issues/148 yet and we haven't decided how we'd approach that
  - answer: a simple skin surface extraction will be included in the virtual fit code to be merged in, but ultimately we do want to have a robust standalone skin surface extraction in openlifu-python, since it is needed for both VF and TT.
- [ ] we are currently resampling the volume into the simulation coordinate grid, ~~but we probably want to have a separate "virtual fitting" coordinate grid, so we might need another version of `make_xarray_in_transducer_coords_from_volume`. see https://github.com/OpenwaterHealth/OpenLIFU-python/issues/165#issuecomment-2486522010~~ Actually the volume should be provided as a raw array without resampling, just needs to be in LPS coords.
- [ ] make sure target is the coordinates in which the target is expected. It should be LPS coordinates, same as the volume. See https://github.com/OpenwaterHealth/OpenLIFU-python/pull/162#discussion_r1848791128
- [ ] make sure transform units are specified in the right coordinates -- see https://github.com/OpenwaterHealth/OpenLIFU-python/pull/162#discussion_r1848947468 -- notice we are switching the output type to ArrayTransform, which comes with units.
- [ ] fix the way the transform is interpreted: it's not an update but rather mapping the transducer into the LPS mri volume space. that means it's always mapping from the transducer's native coordinate system, where it's centered at the origin and "facing up" (forgetting about units -- the transform has its own units which are most likely the volume's units)
- [ ] use protocol to set pitch and yaw ranges. see https://github.com/OpenwaterHealth/OpenLIFU-python/issues/165
- [ ] await merging of https://github.com/OpenwaterHealth/OpenLIFU-python/pull/162, and then update the `python-requirements.txt`

---

Note to self: I started work on the branch `update_vf_placeholder_to_work_with_P#179` at 934e86d5ad3e0e8c881476f897cbcbfc52f43fe1 which is in this PR, and that branch is likely to be merged to main before this PR. So when rebasing this PR to main I may find myself having to remove the first commit.

### DIFF
--- a/OpenLIFULib/OpenLIFULib/transducer.py
+++ b/OpenLIFULib/OpenLIFULib/transducer.py
@@ -159,3 +159,11 @@ class SlicerOpenLIFUTransducer:
 
     def stop_observing_transform_modified(self, tag:int) -> None:
         self.transform_node.RemoveObserver(tag)
+
+    def set_current_transform_to_match_transform_node(self, transform_node : vtkMRMLTransformNode) -> None:
+        """Set the matrix on the current transform node of this transducer to match the matrix of a given transform node.
+        (This is done by a copy not reference, so it's a one-time update -- the tranforms do not become linked in any way.)"""
+
+        transform_matrix = vtk.vtkMatrix4x4()
+        transform_node.GetMatrixTransformToParent(transform_matrix)
+        self.transform_node.SetMatrixTransformToParent(transform_matrix)
--- a/OpenLIFULib/OpenLIFULib/transducer.py
+++ b/OpenLIFULib/OpenLIFULib/transducer.py
@@ -6,6 +6,7 @@ import slicer
 from slicer import (
     vtkMRMLModelNode,
     vtkMRMLTransformNode,
+    vtkMRMLNode,
 )
 from slicer.parameterNodeWrapper import parameterPack
 from OpenLIFULib.parameter_node_utils import SlicerOpenLIFUTransducerWrapper
@@ -167,3 +168,11 @@ class SlicerOpenLIFUTransducer:
         transform_matrix = vtk.vtkMatrix4x4()
         transform_node.GetMatrixTransformToParent(transform_matrix)
         self.transform_node.SetMatrixTransformToParent(transform_matrix)
+
+    def move_node_into_transducer_sh_folder(self, node : vtkMRMLNode) -> None:
+        """In the subject hiearchy, move the given `node` into this transducer's transform node folder."""
+        shNode = slicer.vtkMRMLSubjectHierarchyNode.GetSubjectHierarchyNode(slicer.mrmlScene)
+        shNode.SetItemParent(
+            shNode.GetItemByDataNode(node),
+            shNode.GetItemParent(shNode.GetItemByDataNode(self.transform_node)),
+        )
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1709,15 +1709,12 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
             replace=True, # If there happen to already be some virtual fit result nodes that clash, loading a session will silently overwrite them.
         )
 
-        shNode = slicer.mrmlScene.GetSubjectHierarchyNode()
-        transducer_parent_folder_id = shNode.GetItemParent(shNode.GetItemByDataNode(newly_loaded_transducer.transform_node))
-        
         for vf_node in newly_added_vf_result_nodes:
             preplanning_widget : OpenLIFUPrePlanningWidget = slicer.modules.OpenLIFUPrePlanningWidget
             preplanning_widget.watchVirtualFit(vf_node)
 
             # Place virtual fit results under the transducer folder
-            shNode.SetItemParent(shNode.GetItemByDataNode(vf_node), transducer_parent_folder_id)
+            newly_loaded_transducer.move_node_into_transducer_sh_folder(vf_node)
 
         # === Load transducer tracking results ===
 
@@ -1729,8 +1726,8 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
         )
 
         for (transducer_to_photoscan_node, photoscan_to_volume_node) in newly_added_tt_result_nodes:
-            shNode.SetItemParent(shNode.GetItemByDataNode(transducer_to_photoscan_node), transducer_parent_folder_id)
-            shNode.SetItemParent(shNode.GetItemByDataNode(photoscan_to_volume_node), transducer_parent_folder_id)
+            newly_loaded_transducer.move_node_into_transducer_sh_folder(transducer_to_photoscan_node)
+            newly_loaded_transducer.move_node_into_transducer_sh_folder(photoscan_to_volume_node)
 
         # === Toggle slice visibility and center slices on first target ===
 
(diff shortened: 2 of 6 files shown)


OpenLIFULib/OpenLIFULib/transducer.py
```diff
@@ -70,6 +77,37 @@ def initialize_from_openlifu_transducer(
...
 
+    def update_transform(self, transform_matrix:np.ndarray, transform_matrix_units:Optional[str]=None):
+        """ Update the transducer transform by postcomposing an additional matrix on top of the current transform.
+
+        The transform_matrix is assumed to be in "openlifu" style transducer coordinates, which is currently hardcoded to being LPS,
+        so this function does the needed conversions.
+
+        This function is useful for applying transform updates that come from algorithms in openlifu-python,
+        where the transform would be in openlifu conventions.
+        """
```
ebrahimebrahim, line 88: @sadhana-r I just want to make you aware that this `SlicerOpenLIFUTransducer.update_transform` function is going to be part of my virtual fitting PR. I think your transducer tracking work would also involve an update to the transducer transform, so this could save you the effort of having to write that function.

ltetrel:
>    another thing I am wondering is where the skin surface is extracted. maybe the skin surface is assumed to be one of the inputs into the virtual fitting algorithm? Since we aren't doing [Implement the virtual fitting optimization metric OpenLIFU-python#148](https://github.com/OpenwaterHealth/OpenLIFU-python/issues/148) yet and we haven't decided how we'd approach that

IMO the surface extraction should be done [by the virtual fitting](https://github.com/OpenwaterHealth/OpenLIFU-python/blob/5594c67057ff754860b698b76149bdcb0bbfa087/src/openlifu/vf/virtual_fit.py#L59-L66) because it is really specific to the search of transducer position. But I am not sure if we will handle the timing, short-term we can have it done by Slicer API easily (on the UI: I made a segmentation using auto/otsu thresholding, then right-clik on seg and `Export visible segments to models`)

ltetrel:
> we are currently resampling the volume into the simulation coordinate grid, but we probably want to have a separate "virtual fitting" coordinate grid, so we might need another version of make_xarray_in_transducer_coords_from_volume. see https://github.com/OpenwaterHealth/OpenLIFU-python/issues/165#issuecomment-2486522010

What is the rational behind having a separate coordinate grid ?
You mean we could have all in LPS coordinates from Slicer, work in this for virtual fit and handle the resampling to sim grid (lat, ele, ax) only in `calc_solution` ? Actually this could help user-interaction Slicer-side (if we stay in same, i.e. LPS coords, we can show output from vf to the user, like surface, transducer pos etc...).

ebrahimebrahim:
@sadhana-r @arhowe00 Added you both as reviewers for visibility but I am not expecting anything from this review, you can just approve :)

What this adds is:
- virtual fitting placeholder removed and replaced by actual virtual fitting: when you click "virtual fit" the algorithm runs  and at the end the transducer transform is updated
- virtual fitting status messages are shown in the bottom status bar while the algorithm runs

(@sadhana-r If this gets merged before yours then there will be an annoyance of conflicting with the skinseg tools and IJK transform modifications that are also in one of your working branches, sorry if so! I can always help out if a rebase gets gnarly don't hesitate to let me know!)

ebrahimebrahim:
(just rebased to main, still awaiting approval)

sadhana-r (approved):
Very cool to have this integrated! I just tested it out on a couple sessions and it works well. I noticed that the virtual fit transforms don't get added under the transducer parent folder (unlike the transforms that get loaded in) but otherwise seems to work as expected.

ebrahimebrahim:
Yay, thanks for testing!

Good point I need to put the nodes under the correct folder... I think I will do that before merging.

OpenLIFULib/OpenLIFULib/transducer.py
```diff
@@ -159,3 +160,19 @@ def observe_transform_modified(self, callback : "Callable[[SlicerOpenLIFUTransdu
...
+        transform_node.GetMatrixTransformToParent(transform_matrix)
+        self.transform_node.SetMatrixTransformToParent(transform_matrix)
+
+    def move_node_into_transducer_sh_folder(self, node : vtkMRMLNode) -> None:
+        """In the subject hiearchy, move the given `node` into this transducer's transform node folder."""
+        shNode = slicer.vtkMRMLSubjectHierarchyNode.GetSubjectHierarchyNode(slicer.mrmlScene)
+        shNode.SetItemParent(
+            shNode.GetItemByDataNode(node),
+            shNode.GetItemParent(shNode.GetItemByDataNode(self.transform_node)),
+        )
```
ebrahimebrahim, line 178: @sadhana-r
added this function under SlicerOpenLIFUTransducer to make it simpler to move things into the transducer folder

PR #142, opened 2024-11-19
Add transducer tracking UI

### LINKED ISSUE #137: Add transducer tracking UI
# Background

We don’t have an API yet, but we have an idea of what the inputs will be.
Kedar should be merging his work shortly to [this repo](https://github.com/OpenwaterHealth/OpenLIFU-3DRegistration) and this should contain an API.

We have a plan for the transducer tracking module here:[transducer tracking plan](https://docs.google.com/document/d/193eyWhFVwuidXAGcjt8vmyy_ga6Q1j8w5VVpFaFt5q0/edit?usp=drive_link)
Currently still needs to be adjusted a bit, but we can already carry out the “temporary plan” at the bottom for now.
The adjustments involve some sweeping changes to the way transducer transforms are handled – want to put this off to SOW 2 as it will block us from starting to pull in Kedar’s code which we want to do sooner rather than later.

We expect the inputs to be:
- an MRI volume
- a transducer registration surface mesh
- a photoscan

And the outputs should be:
- a transform that aligns the photoscan to the MRI volume space
- a transform that aligns the transducer registration surface mesh to the photoscan

Ultimately the outputs will be in pieces, and there should be previews and approvals for individual pieces.
But for now let’s focus on the end-to-end workflow.

For the MRI volume we’re all set to make that an input in SlicerOpenLIFU.
The other two will need work.

For now, the transducer registration surface mesh can just be a model that we load using a button in the transducer tracking module

The photoscan on the other hand has to be a model _with texture_, and whatever representation we have of a textured model in Slicer we will want the same thing to be input into Kedar’s algorithm. However, Kedar’s initial implementation will use a model without texture, which makes slicer integration much simpler for us.

# This issue

Following the “temporary plan” at the end of the [transducer tracking plan](https://docs.google.com/document/d/193eyWhFVwuidXAGcjt8vmyy_ga6Q1j8w5VVpFaFt5q0/edit?usp=drive_link) doc, add the input selectors and buttons int he transducer tracking module. The input selectors and buttons can be fully functional, with the exception that the "run transducer tracking button" will still be a placeholder.

The “load transducer registration surface mesh” and “load photoscan” buttons just open dialogs for loading models (later we will want “load photoscan” to allow loading some representation of a textured mesh, but for now it can just be a model)

# Follow-up work

- (#138) Integrate with Kedar's algorithms once those are available, installing the needed dependencies and  linking the "run transducer tracking" button to the right things.
- Determine what’s the best way to represent textured meshes in Slicer, and use those for photoscans instead of plain untextured models.
  - Update: The best way is simply a pair consisting of a model and a texture. So a vtkPolyData for the model and an image array, like a numpy array or a vtkTexture, for the texture image. The texture coordinate mapping would just be on the vtkPolyData. In the photoscans, the texture coordinates are there in the obj file, and when you load an obj file using slicer they are automatically loaded in. vtk [natively supports texturing an actor](https://examples.vtk.org/site/Python/Texture/TexturePlane/) so a (`vtkPolyData` with built in texture coordinates, `vtkTexture`) pair is a perfectly reasonable way to represent a textured mesh in openlifu-python.
- After a TBD overhaul of transducer transforms is done on openlifu python, we can break down the transducer tracking workflow into appropriate fine grained steps.

sadhana-r:
Closes #137 

The way that transducers surfaces and photoscans are being handled in this PR will be updated in future work. I updated the algorithm inputs to include Photoscans in the current PR since it seemed like this approach would also be helpful in the final implementation.

### DIFF
--- a/OpenLIFULib/OpenLIFULib/algorithm_input_widget.py
+++ b/OpenLIFULib/OpenLIFULib/algorithm_input_widget.py
@@ -2,7 +2,7 @@ from typing import Tuple, Any, List
 from dataclasses import dataclass
 import qt
 import slicer
-from slicer import vtkMRMLScalarVolumeNode, vtkMRMLMarkupsFiducialNode
+from slicer import vtkMRMLScalarVolumeNode, vtkMRMLMarkupsFiducialNode, vtkMRMLModelNode
 from OpenLIFULib.parameter_node_utils import SlicerOpenLIFUProtocol
 from OpenLIFULib.util import get_openlifu_data_parameter_node
 from OpenLIFULib.transducer import SlicerOpenLIFUTransducer
@@ -45,6 +45,9 @@ class OpenLIFUAlgorithmInputWidget(qt.QWidget):
             elif input_name == "Target":
                 self.target_input = AlgorithmInput("Target", qt.QComboBox(self))
                 self.inputs.append(self.target_input)
+            elif input_name == "Photoscan":
+                self.photoscan_input = AlgorithmInput("Photoscan", qt.QComboBox(self))
+                self.inputs.append(self.photoscan_input)
             else:
                 raise RuntimeError("Invalid algorithm input specified.")
                 
@@ -60,6 +63,9 @@ class OpenLIFUAlgorithmInputWidget(qt.QWidget):
 
     def add_volume_to_combobox(self, volume_node : vtkMRMLScalarVolumeNode) -> None:
         self.volume_input.combo_box.addItem("{} (ID: {})".format(volume_node.GetName(),volume_node.GetID()), volume_node)
+
+    def add_photoscan_to_combobox(self, model_node : vtkMRMLModelNode) -> None:
+        self.photoscan_input.combo_box.addItem("{} (ID: {})".format(model_node.GetName(),model_node.GetID()), model_node)
 
     def set_session_related_combobox_tooltip(self, text:str):
         """Set tooltip on the transducer and volume comboboxes."""
@@ -88,7 +94,6 @@ class OpenLIFUAlgorithmInputWidget(qt.QWidget):
     def _populate_from_loaded_objects(self) -> None:
         """" Update protocol, transducer, and volume comboboxes based on the OpenLIFU objects loaded into the scene.
         Adds the items only; does not clear the ComboBoxes."""
-
         dataParameterNode = get_openlifu_data_parameter_node()
 
         # Update protocol combo box
@@ -168,6 +173,20 @@ class OpenLIFUAlgorithmInputWidget(qt.QWidget):
                 self.target_input.combo_box.setEnabled(True)
                 for target_node in target_nodes:
                     self.target_input.combo_box.addItem(target_node.GetName(), target_node)
+
+        # Update photoscans combobox 
+        # NOTE: This code can be moved to populate_from_loaded_objects once photoscans are associated with sessions
+        # This is temporarily here to populate the combobox with the photoscan loaded to the scene in the
+        # transducer tracking module. This may change based on how openlifu-python handles photoscans. 
+        if "Photoscan" in self.algorithm_input_names:
+            num_photoscans = 0
+            for model_node in slicer.util.getNodesByClass('vtkMRMLModelNode'):
+                # Check that the model is a loaded photoscan model
+                if model_node.GetAttribute('isOpenLIFUPhotoscan'):
+                    self.add_photoscan_to_combobox(model_node)
+                    num_photoscans += 1
+            if num_photoscans == 0:
+                self.photoscan_input.indicate_no_options()
 
         # Set selections to the previous ones when they exist
         self._set_most_recent_selections()
--- a/OpenLIFULib/OpenLIFULib/algorithm_input_widget.py
+++ b/OpenLIFULib/OpenLIFULib/algorithm_input_widget.py
@@ -58,7 +58,7 @@ class OpenLIFUAlgorithmInputWidget(qt.QWidget):
         """Set tooltip on the transducer, protocol and volume comboboxes."""
 
         for input in ["Protocol", "Transducer", "Volume"]:
-            if input in self.inputs_dict.keys():
+            if input in self.inputs_dict:
                 self.inputs_dict[input].combo_box.setToolTip(text)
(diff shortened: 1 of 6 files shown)


sadhana-r:
The way that transducers surfaces and photoscans are being handled in this PR will be updated in future work. I updated the algorithm inputs to include Photoscans in the current PR since it seemed like this approach would also be helpful in the final implementation.

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -224,3 +302,7 @@ def toggleTransducerTrackingApproval(self) -> None:
             raise RuntimeError("Cannot toggle tracking approval because there is no active session.")
         session.toggle_transducer_tracking_approval() # apply the approval or lack thereof
         data_parameter_node.loaded_session = session # remember to write the updated session object into the parameter node
+
+    def runTransducerTracking(self) -> None:
```
ebrahimebrahim, line 306: (I am just looking quickly here, this doesn't look like it has the right function signature given how it's called above)

sadhana-r: Oh you're right, I missed updating that

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -224,3 +306,14 @@ def toggleTransducerTrackingApproval(self) -> None:
...
+
+    def runTransducerTracking(self,
+                              inputProtocol: SlicerOpenLIFUProtocol,
+                              inputTransducer : SlicerOpenLIFUTransducer,
+                              inputVolume: vtkMRMLScalarVolumeNode,
+                              inputPhotoscan: vtkMRMLModelNode,
+                              inputTRS: vtkMRMLModelNode
+                              ) -> Tuple[vtkMRMLTransformNode, vtkMRMLTransformNode]:
+        ## Need to integrate with transducer tracking library here
+        print("Transducer tracking integration under development")
```
ebrahimebrahim, line 318: Better to raise an infobox here, since we plan to merge this PR and it could end up in a release.

ebrahimebrahim (changes requested):
I like the way you approached this, very nice work!

one additional item:

- add `#` to the commit message "Update definition for function to run transducer tracking (137)"


BTW for an obj file example to test with I found [this trumpet](https://github.com/pyvista/vtk-data/blob/master/Data/trumpet.obj) useful -- the mannequin photoscan takes too long to load.

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -161,6 +182,67 @@ def setParameterNode(self, inputParameterNode: Optional[OpenLIFUTransducerTracke
...
+        # Using a custom file dialog instead of slicer.util.openAddModelDialog() incase database specific 
+        # customizations are required later 
+        qsettings = qt.QSettings()
+
+        filepath: str = qt.QFileDialog.getOpenFileName(
+            slicer.util.mainWindow(), # parent
+            'Load photoscan', # title of dialog
+            qsettings.value('OpenLIFU/databaseDirectory','.'), # starting dir, with default of '.'
+            "Model (*.obj; *.vtk);;All Files (*)", # file type filter
+        )
```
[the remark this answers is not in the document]
sadhana-r: I am not able to replicate this issue since it filters correctly on Windows. But I removed the semicolon separating the file types since that it how it is implemented in the Data module. I'm assuming that's related to why the obj files aren't showing up for you.

sadhana-r:
Thank you for the review and helpful suggestions! I agree, using dictionaries is alot cleaner and avoids having to keep track of the order of inputs in the algorithm widget.

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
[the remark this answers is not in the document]
ebrahimebrahim: (that fix worked, thanks!)

OpenLIFULib/OpenLIFULib/algorithm_input_widget.py
```diff
@@ -20,97 +20,99 @@ def indicate_no_options(self):
...
-        for input in [
-            self.protocol_input,
-            self.transducer_input,
-            self.volume_input,
-        ]:
-            input.combo_box.setToolTip(text)
+        """Set tooltip on the transducer, protocol and volume comboboxes."""
+
+        for input in ["Protocol", "Transducer", "Volume"]:
+            if input in self.inputs_dict.keys():
```
ebrahimebrahim, line 61:
```suggestion
            if input in self.inputs_dict:
```

Minor style comment for anywhere you have `x in d.keys()` : in python this is equivalent to `x in d` which looks a bit nicer. No need to change it just nice to know

OpenLIFULib/OpenLIFULib/algorithm_input_widget.py
```diff
@@ -149,23 +154,47 @@ def update(self):
...
+    def get_current_data(self) -> Dict[str, Any]:
+        """Get the current selections as a Dictionary. Potential output data types are:
+            Protocol: SlicerOpenLIFUProtocol
+            Transducer: SlicerOpenLIFUTransducer
+            Volume: vtkMRMLScalarVolumeNode
+            Target: vtkMRMLMarkupsFiducialNode
+            Photoscan: vtkMRMLModelNode """
+        current_data_dict = {}
+        for input in self.inputs_dict.values():
+            current_data_dict[input.name] = input.combo_box.currentData
```
ebrahimebrahim, line 199:
```suggestion
        current_data_dict = {
            input.name : input.combo_box.currentData
            for input in self.inputs_dict.values()
        }
```

(just style preference, no need to change)

ebrahimebrahim (approved):
All looks good, thank you! Feel free to merge with or without style changes, those were just fyi

OpenLIFULib/OpenLIFULib/algorithm_input_widget.py
sadhana-r: Thanks, didn't realize that!

PR #154, opened 2024-12-06
Use manual virtual fit as stand-in for automated

ebrahimebrahim:
This is a temporary measure until we get the automated virtual fitting in place.

When you click "Virtual fit" the interaction handles on the transducer transform are enabled, and you now get this dialog:

image

Clicking "Virtual fit" again removes the handles

### DIFF
--- a/OpenLIFUPrePlanning/OpenLIFUPrePlanning.py
+++ b/OpenLIFUPrePlanning/OpenLIFUPrePlanning.py
@@ -455,10 +455,20 @@ class OpenLIFUPrePlanningLogic(ScriptedLoadableModuleLogic):
             volume: vtkMRMLScalarVolumeNode,
             target: vtkMRMLMarkupsFiducialNode,
         ):
-        slicer.util.infoDisplay(
-            text="The virtual fit button is a placeholder. Virtual fit algorithm not yet implemented.",
-            windowTitle="Not implemented"
-        )
+        # Temporary measure of "manual" virtual fitting. See https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/153
+        transducer.transform_node.CreateDefaultDisplayNodes()
+        if not transducer.transform_node.GetDisplayNode().GetEditorVisibility():
+            slicer.util.infoDisplay(
+                text=(
+                    "The automatic virtual fitting algorithm is not yet implemented."
+                    " Use the interaction handles on the transducer to manually fit it."
+                    " You can click the Virtual fit button again to remove the interaction handles."
+                ),
+                windowTitle="Not implemented"
+            )
+            transducer.transform_node.GetDisplayNode().SetEditorVisibility(True)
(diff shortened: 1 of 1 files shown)


sadhana-r (commented):
As I understand it, for this temporary fix, as currently implemented, for the transducer tracking module to access the manual fit transform for initialization, it would need to include this code snippet from` "update_underlying_openlifu_session"`  to get the updated transform:
```
# Update transducer transform in the underlying Session
transducer = get_openlifu_data_parameter_node().loaded_transducers[self.get_transducer_id()]
transducer_openlifu = transducer.transducer.transducer
transducer_transform_node : vtkMRMLTransformNode = transducer.transform_node
transducer_transform_array = slicer.util.arrayFromTransformMatrix(transducer_transform_node, toWorld=True)
        openlifu2slicer_matrix = linear_to_affine(
            get_xxx2ras_matrix('LPS') * get_xx2mm_scale_factor(transducer_openlifu.units)
        )
self.session.session.array_transform = openlifu_lz().db.session.ArrayTransform(
            matrix = np.linalg.inv(openlifu2slicer_matrix) @ transducer_transform_array,
            units = transducer_openlifu.units,
 )
```
It may be helpful if instead the interactions handles are turned off when the "Approve virtual fit" button is clicked. And this can also include updating the array transform in the loaded session (without saving the session to file), so its easier for the TT module to access? This can later be changed once we update the way virtual fit transforms are saved. What do you think?

ebrahimebrahim:
> As I understand it, for this temporary fix, as currently implemented, for the transducer tracking module to access the manual fit transform for initialization, it would need to include this code snippet from` "update_underlying_openlifu_session"` to get the updated transform:

But then what happens if there isn't an active session?

I was thinking that in the temporary solution here, the transducer tracking module will use the matrix on the `transducer.transform_node` where `transducer` is the input SlicerOpenLIFUTransducer.

Then later for the actual implementation there could be different behaviors depending on whether there's a session:
- If there is no active session then the matrix is obtained from the virtual fit transform node, and we just have the one matrix.
- If there is an active session then the list of matrices is obtained from the virtual fit result in the session

---



> It may be helpful if instead the interactions handles are turned off when the "Approve virtual fit" button is clicked. And this can also include updating the array transform in the loaded session (without saving the session to file), so its easier for the TT module to access? This can later be changed once we update the way virtual fit transforms are saved.

I agree that workflow-wise it makes more sense to use the approve virtual fit button, but it brings up the same question:  how would you proceed when there isn't an active session?

sadhana-r:
Hmm ok that makes sense. I wasn't considering that the approval step is specific to sessions. This works for now then! Once we add the virtual fit transform node to the scene, we can figure out the manual workflow for transducer tracking.

sadhana-r (approved):

ebrahimebrahim:
Thanks for your comments and review!

PR #155, opened 2024-12-11
Add photoscan to database  and add segmentation selector to TT workflow

### LINKED ISSUE #146: Enable adding photoscans to sessions
Add a button to the OpenLIFUData module “add photoscan to session”, which works similarly to “add subject to volume”. This opens a dialog with file browsers and input boxes, and calls the underlying functionality in `openlifu.Database`

What we really want eventually is to link this to the phone app / companion desktop app that does the photogrammetric reconstruction-- that will be done in #150. Here we are just adding a preliminary functionality that assumes the user somehow got the photoscan in hand and just wants to point the application to the right files.

Depends on https://github.com/OpenwaterHealth/OpenLIFU-python/issues/177

### LINKED ISSUE #152: Add skin segmentation selector
One of the inputs to the transducer tracking algorithm that we haven't accounted for yet is the skin surface from the MRI. The raw MRI shouldn't actually be an input to the registration algorithm.

And until [P#149](https://github.com/OpenwaterHealth/OpenLIFU-python/issues/149) is done, as a temporary measure to get a minimal transducer tracking workflow working we can add an input combobox that just includes all model nodes in the scene, requiring the user to somehow get a skin surface model node into the scene however they want.

(Let's not remove the volume node as an input since it will be used in #149 after we remove the present temporary measure.)

Depends on #137

sadhana-r:
Closes #146 
Closes #152 

Depends on photoscan writing fix in [https://github.com/OpenwaterHealth/OpenLIFU-python/tree/180_load_photoscans](https://github.com/OpenwaterHealth/OpenLIFU-python/tree/180_load_photoscans).

### DIFF
--- a/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
+++ b/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
@@ -198,6 +198,13 @@ class OpenLIFUTransducerTrackerWidget(ScriptedLoadableModuleWidget, VTKObservati
         """Update the algorithm input options"""
         self.algorithm_input_widget.update()
 
+        # Temporary code to include skin segmentation model as input
+        loaded_models = slicer.util.getNodesByClass('vtkMRMLModelNode')
+        if len(loaded_models) == 3:
+            self.ui.skinSegmentationModelqMRMLNodeComboBox.enabled = False
+        else:
+            self.ui.skinSegmentationModelqMRMLNodeComboBox.enabled = True
+
         # Determine whether transducer tracking can be run based on the status of combo boxes
         self.checkCanRunTracking()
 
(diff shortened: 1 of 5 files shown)


OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -198,6 +198,13 @@ def updateInputOptions(self):
         """Update the algorithm input options"""
         self.algorithm_input_widget.update()
 
+        # Temporary code to include skin segmentation model as input
+        loaded_models = slicer.util.getNodesByClass('vtkMRMLModelNode')
+        if len(loaded_models) == 3:
```
ebrahimebrahim, line 203: I couldn't understand what's the idea behind this check, why disable at exactly 3?

ebrahimebrahim (commented):
This worked beautifully and the organization is very good! Love the right click context menu. No need for changes -- I just had one question where I didn't follow something. But I will hold off on giving approval until the python library version can be updated, to help block myself from eagerly merging before that is done.

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
sadhana-r: By default, when slicer is started up, there are 3 hidden model nodes. So the dropdown is disabled when no new models have been added to the scene.

ebrahimebrahim: Oh that makes sense. 

a brittle solution but this is temporary code so that's the way to do it :smile:

sadhana-r:
I rebased, updated the python library version and tested out the photoscan features. Once approved, I can go ahead and merge this PR!

ebrahimebrahim (approved):
excellent!!

PR #159, opened 2025-01-14
Update SlicerOpenLIFUTransducer 

### LINKED ISSUE #151: Affiliate new transducer models with SlicerOpenLIFUTransducer
Just as a SlicerOpenLIFUTransducer owns two "affiliated" mrml nodes, we should add in the model nodes introduced by https://github.com/OpenwaterHealth/OpenLIFU-python/issues/181 as more affiliated model nodes. So they should be loaded in if available, and they should be part of the SlicerOpenLIFUTransducer similarly to the mrml nodes that are currently part of SlicerOpenLIFUTransducer

Depends on https://github.com/OpenwaterHealth/OpenLIFU-python/issues/181

sadhana-r:
Closes #151 

This PR depends on https://github.com/OpenwaterHealth/OpenLIFU-python/pull/188

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -548,8 +548,7 @@ class OpenLIFUDataWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         self.addObserver(slicer.mrmlScene, slicer.mrmlScene.EndCloseEvent, self.onSceneEndClose)
 
         # This ensures that we properly handle SlicerOpenLIFU objects that become invalid when their nodes are deleted
-        shNode = slicer.mrmlScene.GetSubjectHierarchyNode()
-        shNode.AddObserver(shNode.SubjectHierarchyItemAboutToBeRemovedEvent, self.onSHItemAboutToBeRemoved)
+        self.setupSHNodeObserver()
         self.addObserver(slicer.mrmlScene, slicer.vtkMRMLScene.NodeAddedEvent, self.onNodeAdded)
         self.addObserver(slicer.mrmlScene, slicer.vtkMRMLScene.NodeRemovedEvent, self.onNodeRemoved)
 
@@ -1074,6 +1073,7 @@ class OpenLIFUDataWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         # If this module is shown while the scene is closed then recreate a new parameter node immediately
         if self.parent.isEntered:
             self.initializeParameterNode()
+        self.setupSHNodeObserver()
 
     @vtk.calldata_type(vtk.VTK_INT)
     def onSHItemAboutToBeRemoved(self, caller, event, removedItemID):
@@ -1122,7 +1122,13 @@ class OpenLIFUDataWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
 
         self.setParameterNode(self.logic.getParameterNode())
         self.updateParametersFromSettings()
-
+    
+    def setupSHNodeObserver(self) -> None:
+        """Set and observe the subject hierarchy node.
+        Observation is needed so that certain actions can take place when an openlifu affiliated node or folder is removed from the scene.
+        """
+        shNode = slicer.mrmlScene.GetSubjectHierarchyNode()
+        shNode.AddObserver(shNode.SubjectHierarchyItemAboutToBeRemovedEvent, self.onSHItemAboutToBeRemoved)
 
     def updateParametersFromSettings(self):
         parameterNode : vtkMRMLScriptedModuleNode = self._parameterNode.parameterNode


ebrahimebrahim:
@sadhana-r if this is pretty much ready to be reviewed, you can rebase to main and then tag me as reviewer

sadhana-r:
I was working on some changes today and think I've almost figured out the deletion problem. I'm just doing some final testing/cleanup and should have this ready to review tomorrow!

sadhana-r:
Ok this is now ready for review - changes for this issue are the last 4 commits. Things to review

- Loading/reloading sessions and manually loading a transducer
- Deleting a transducer by manually removing a node. In the session workflow and manual workflow. 
- Deleting a transducer by manually removing the whole folder. In the session workflow and manual workflow. 
- In the session workflow, when a folder is deleted and the nodes of the unloaded session are cleared (i.e. click the checkbox), then my slicer hangs. I don't have this problem when I don't clear the session data. I also don't get any errors when I delete a transducer (by deleting a node, not folder) and then clear the session data. 
- My slicer hangs when I remove the timer and directly call `delayed_node_deletion`. Could you see what behavior you get when you do that? I thought it would just throw VTK warnings. I tested the reducing the timer length to 100 instead of 500 and it still works well so I don't think we should run into issues due to the timer failing..

ebrahimebrahim:
On ubuntu 22.04:

> In the session workflow, when a folder is deleted and the nodes of the unloaded session are cleared (i.e. click the checkbox), then my slicer hangs. I don't have this problem when I don't clear the session data. I also don't get any errors when I delete a transducer (by deleting a node, not folder) and then clear the session data.

My Slicer doesn't hang from this

> My slicer hangs when I remove the timer and directly call delayed_node_deletion. Could you see what behavior you get when you do that? I thought it would just throw VTK warnings. I tested the reducing the timer length to 100 instead of 500 and it still works well so I don't think we should run into issues due to the timer failing..

My Slicer _crashes_ instead of hanging when I do this

ebrahimebrahim:
(the commit "Fix bug in adding photoscans to combobox" should reference #160 not #151)

OpenLIFULib/OpenLIFULib/transducer.py
```diff
@@ -64,17 +71,50 @@ def initialize_from_openlifu_transducer(
...
+        else:
+            body_model_node = None
+
+        if transducer.registration_surface_filename:
+            surface_model_node = slicer.util.loadModel(Path(parent_dir)/transducer.registration_surface_filename)
+            shNode.SetItemParent(shNode.GetItemByDataNode(surface_model_node), parentFolderItem)
+            surface_model_node.SetAndObserveTransformNodeID(transform_node.GetID())
+            surface_model_node.SetName(f"{slicer_transducer_name}-surface")
+        else:
+            surface_model_node = None
```
ebrahimebrahim, line 98: The `Path(parent_dir)/transducer.registration_surface_filename` is making assumptions about how the openlifu database works. Having `initialize_from_openlifu_transducer` accept a `parent_dir` makes sense if we're using it to load from a json file directly, but when we load a transducer via the database I think we should be relying on what `openlifu` provides us in order to load the transducer body and surface models. Can we solve this using a similar method to what you came up with for photoscan data in #158? It may require exposing something new in `openlifu`. It may require exposing something new in `openlifu`

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -206,24 +202,11 @@ def updateInputOptions(self):
...
-            slicer.util.mainWindow(), # parent
-            'Load transducer', # title of dialog
-            qsettings.value('OpenLIFU/databaseDirectory','.'), # starting dir, with default of '.'
-            "Model (*.obj *.vtk);;All Files (*)", # file type filter
-        )
-        if filepath:
-            self.activeTRS = slicer.util.loadModel(filepath) # Temporary approach
-            self.checkCanRunTracking()
 
     def checkCanRunTracking(self,caller = None, event = None) -> None:
```
ebrahimebrahim, line 206: In this function do we now also need to check that the loaded transducer has a registration surface?
Since we allow transducers to be loaded that do not have an affiliated surface model.

If the surface model is not present and that's the only reason the tracking button is disabled, then the runTrackingButton tooltip could be "This transducer does not have an affiliated registration surface, which is needed to run tracking."

OpenLIFUData/OpenLIFUData.py
```diff
@@ -1445,8 +1462,10 @@ def load_session(self, subject_id, session_id) -> None:
 
         # === Load transducer ===
 
+        transducer_metadata_filepath = self.db.get_transducer_filename(session_openlifu.transducer_id)
         newly_loaded_transducer = self.load_transducer_from_openlifu(
             transducer = self.db.load_transducer(session_openlifu.transducer_id),
+            parent_dir= Path(transducer_metadata_filepath).parent,
```
ebrahimebrahim, line 1468: This is the point where we probably want to ask `openlifu` for the info we need rather than messing around with database internals.

(I feel like in `openlifu.Database` we should put a `_` at the beginning of functions like `get_transducer_filename` to avoid this kind of thing in first place)

OpenLIFUData/OpenLIFUData.py
```diff
@@ -1536,11 +1555,13 @@ def remove_protocol(self, protocol_id:str) -> None:
 
     def load_transducer_from_file(self, filepath:str) -> None:
         transducer = openlifu_lz().Transducer.from_file(filepath)
-        self.load_transducer_from_openlifu(transducer)
+        transducer_parent_dir = Path(filepath).parent
+        self.load_transducer_from_openlifu(transducer, parent_dir = transducer_parent_dir)
```
ebrahimebrahim, line 1559: Here the use of `transducer_parent_dir` does make sense I think and can be kept

ebrahimebrahim (changes requested):
I reviewed everything except the folder deletion and timer stuff, in case you were still planning on changing something there. On the next review I'll include the folder deletion and timer stuff.

OpenLIFUData/OpenLIFUData.py
sadhana-r: You're right, I'm thinking that I'll instead call `get_transducer_absolute_filepaths`  to get all the model filepaths and use that to load the transducer from the database since that was the purpose of having that function. What do you think of that?  
Following the approach used for photoscans would involve loading the actual polydata in openlifu i.e. generalizing the code for loading models out of `photoscan.py` in openlifu and returning vtkPolyData for each model included in the transducer file. While loading the model/texture data made sense for photoscans due to the special EXR file type, it probably isn't needed for general model files?

sadhana-r: I'll open an openlifu issue for this so we have it tracked. In an ideal scenario we would want this approach for all data files i.e. volumes as well?

ebrahimebrahim: Thanks!

I think yes it would be nice to do that even for volume files.

It's not that openlifu-python has to be responsible for _file reading_, but it's more about getting the data in the first place. Maybe I am mixing up those two things. I am just thinking for example what if the data were stored online and the way to get it is not an absolute filepath but rather a URL that needs to be sent to a download utility. Or what if the data is shoved inside the sql database somehow (can sql databases do that? idk). Maybe openlifu should be responsible for providing a kind of "file object" that hasn't necessarily been read yet.

Digging around a bit, I found [fsspec](https://filesystem-spec.readthedocs.io/) which seems to have exactly such a generic concept of "file object" that we could use. (It seems it's one of the tools DVC uses under the hood!) So maybe we should use that

Okay but sorry none of that is related to this issue, it's dreams for the future. so ty for opening an issue

ebrahimebrahim (changes requested):
Trying timer stuff again with `slicer.app.processEvents` :D

OpenLIFUData/OpenLIFUData.py
```diff
@@ -547,7 +547,8 @@ def setup(self) -> None:
         self.addObserver(slicer.mrmlScene, slicer.mrmlScene.EndCloseEvent, self.onSceneEndClose)
 
         # This ensures that we properly handle SlicerOpenLIFU objects that become invalid when their nodes are deleted
-        self.addObserver(slicer.mrmlScene, slicer.vtkMRMLScene.NodeAboutToBeRemovedEvent, self.onNodeAboutToBeRemoved)
+        shNode = slicer.mrmlScene.GetSubjectHierarchyNode()
+        shNode.AddObserver(shNode.SubjectHierarchyItemAboutToBeRemovedEvent, self.onSHItemAboutToBeRemoved)
```
ebrahimebrahim, line 551: I found an interesting bug here...

When you clear the slicer scene by File > Close scene, the shNode actually _changes_.
So the obserever that was added here no longer applies, because it's an observer on the _old_ shNode (which might have been destroyed, idk, but anyway it's the wrong one at this point)

So when you clear the slicer scene, you can proceed as normal (after bug fix of #183) but you will have lost the observer and `onSHItemAboutToBeRemoved` will no longer get called.

I am not sure what's the most elegant fix... maybe it needs to be treated similarly to the parameter node  (which also gets destroyed and rebuilt upon scene clearing). I see that with parameter nodes we have `initializeParameterNode` called in the widget `setup` method and also called in the widget `onSceneEndClose` method and in the widget `enter` method. Do we perhaps need some kind of `setup_shNode_observers` function that gets called in various places? @sjh26 what do you think also?

OpenLIFUData/OpenLIFUData.py
```diff
@@ -1166,7 +1182,10 @@ def __init__(self) -> None:
         self.db : Optional[openlifu.Database] = None
 
         self._subjects : Dict[str, openlifu.db.subject.Subject] = {} # Mapping from subject id to Subject
-
+        self._folder_deletion_in_progress = False
+        self._timer = None 
+        # Secondary flag incase the timer fails during folder deletion. Prevents two transducer deletio
+        self._transducer_node_deletion_in_progress = False
```
ebrahimebrahim, line 1188: These flags are evil in my opinion, because they scatter application state all over the place and make it confusing to maintain the code and fix bugs later. Sometimes it's a necessary evil like in the case of `_folder_deletion_in_progress`, which I can't think of a way to avoid. But  I propose to get rid of `_transducer_node_deletion_in_progress`. I tried removing it and didn't notice any issues in my testing. If it's just an "in case the timer fails" thing then we can do away with it.

ebrahimebrahim (changes requested):
A few more changes requested, plus the previous review.

The folder deletion and node deletion stuff worked perfectly in my testing!

By the way since there were six cases to test, I found it useful to automate testing those cases a little bit while playing with different approaches to the affiliated node deletion business. Here is my testing snippet in case it comes in handy later:

```py
from OpenLIFUData import ObjectBeingUnloadedMessageBox

slicer.util.selectModule("OpenLIFUData")
dw = slicer.modules.OpenLIFUDataWidget
dw.onLoadDatabaseClicked(True) # or dw.ui.databaseLoadButton.click()
dl = dw.logic

def load_session():
    dl.load_session("mannequin", "demo_4_simulated")

def clear():
    dl.clear_session(clean_up_scene=True)
    slicer.mrmlScene.Clear()

    # This shNode part makes up for a bug in which the observer disappears after scene clear. Remove when bug is fixed
    shNode = slicer.mrmlScene.GetSubjectHierarchyNode()
    shNode.AddObserver(shNode.SubjectHierarchyItemAboutToBeRemovedEvent, dw.onSHItemAboutToBeRemoved)

def delete_transducer_node(clear_affiliated_transducer_nodes = False, clear_affiliated_session_nodes = False):
    
    def custom_exec_monkeypatch(self,) -> bool:
        if self.windowTitle == "Transducer removed":
            print(f"automatically answering '{clear_affiliated_transducer_nodes}' to whether to clear affiliated trasducer data")
            return clear_affiliated_transducer_nodes
        if self.windowTitle == "Session invalidated":
            print(f"automatically answering '{clear_affiliated_session_nodes}' to whether to clear affiliated session data")
            return clear_affiliated_session_nodes
        return None
    ObjectBeingUnloadedMessageBox.customexec_ = custom_exec_monkeypatch
    
    dp = dl.getParameterNode()
    transducer = list(dp.loaded_transducers.values())[0]
    slicer.mrmlScene.RemoveNode(transducer.model_node)

def delete_transducer_folder(clear_affiliated_session_nodes = False):
    
    def custom_exec_monkeypatch(self,) -> bool:
        if self.windowTitle == "Session invalidated":
            print(f"automatically answering '{clear_affiliated_session_nodes}' to whether to clear affiliated session data")
            return clear_affiliated_session_nodes
        return None
    ObjectBeingUnloadedMessageBox.customexec_ = custom_exec_monkeypatch

    dp = dl.getParameterNode()
    transducer = list(dp.loaded_transducers.values())[0]
    shNode = slicer.vtkMRMLSubjectHierarchyNode.GetSubjectHierarchyNode(slicer.mrmlScene)
    shNode.RemoveItem(shNode.GetItemByName(transducer.name))


def run_node_deletion_test(clear_affiliated_transducer_nodes, clear_affiliated_session_nodes):
    print(
        f"Commencing NODE deletion test in which we {'do' if clear_affiliated_transducer_nodes else 'dont'} clear affiliated transducer nodes"
        f" and {'do' if clear_affiliated_session_nodes else 'dont'} clear affiliated session nodes"
    )
    clear()
    load_session()
    delete_transducer_node(clear_affiliated_transducer_nodes,clear_affiliated_session_nodes)
    print("Test complete")

def run_folder_deletion_test(clear_affiliated_session_nodes):
    print(
        f"Commencing FOLDER deletion test in which we"
        f" {'do' if clear_affiliated_session_nodes else 'dont'} clear affiliated session nodes"
    )
    clear()
    load_session()
    delete_transducer_folder(clear_affiliated_session_nodes)
    print("Test complete")
```

So I'd paste that into the python console in slicer and then run these tests individually and check that everything worked:

```py
run_node_deletion_test(True, True)
run_node_deletion_test(True, False)
run_node_deletion_test(False, True)
run_node_deletion_test(False, False)
run_folder_deletion_test(True)
run_folder_deletion_test(False)
```

I couldn't come up with a better way to do what's currently done with a timer... I am okay with the approach I think. The folder organization is really nice to have

OpenLIFUData/OpenLIFUData.py
sadhana-r: I agree. I added it as a failesafe but I don't think it is really needed and makes it confusing. I'll remove it

OpenLIFULib/OpenLIFULib/transducer.py
```diff
@@ -40,11 +40,10 @@ def initialize_from_openlifu_transducer(
         Args:
             transducer: The openlifu Transducer object
             transducer_matrix: The transform matrix of the transducer. Assumed to be the identity if None.
+            transducer_abspaths_info: Dictionary containing absolute filepath info to any data affiliated with the transducer object.
```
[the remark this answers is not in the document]
sadhana-r: I decided to keep the default value as {} but I removed the `Optional` from the type annotation as you suggested. I felt that this was cleaner than adding a bunch of conditional statements checking to see if this variable is specified/relevant or not.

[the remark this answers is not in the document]
ebrahimebrahim: I see. Now that I look at it, it looks like the empty case is handled as "empty", i.e. you can provide the paths or not and the right thing will happen in all cases ... so my comment about how it doesn't seem like an "empty" case is wrong. Never mind!

ebrahimebrahim:
> Same as before, this test `run_folder_deletion_test(False)` still causes my Slicer to hang/crash. It would be helpful to test this on another Windows machine. It could be due to some issue with my build (?).

Let's... allow the potential bug so we can merge this PR and keep things moving... and then if we confirm that it's a true windows bug and not just your current build then we can open a bug report and dig into it. Keep your slicer debug build hanging around in case we get a chance to debug the crash while on a call.

OpenLIFUData/OpenLIFUData.py
sjh26: If its in `setup` is shouldn't need to be in `enter` as well.  Your approach is fine, this quirk is one of the things I hate about the subject hierarchy node.

sadhana-r:
Tested the shNode fix and rebased to main

PR #166, opened 2025-01-29
Modify protocol config UI and load in setup()

### LINKED ISSUE #163: Add UI to protocol config module
Enhance the `OpenLIFUProtocolConfig` module by adding a simple UI file with the following features:

- Buttons for editing protocol fields such as `name`, `id`, `description`, or other fields introduced in PR#54 (possibly start by copying the widget from PR#54).
- A dropdown menu (`QComboBox`) to select which protocol to edit and an option to create new protocols with default parameters (`Create new protocol...`).
- The `Save Protocol` button will save the protocol to the database, given the `protocol_id`, with a warning if overwriting an existing `protocol_id` (the warning part should be deferred to the widget and logic, not the UI).
- Clone the `Load Protocol` button from the `OpenLIFUData` module for loading protocols into the scene.
- A button to delete protocols.

### References
- ~~Depends on https://github.com/OpenwaterHealth/OpenLIFU-python/issues/196~~
- Reuse and modify elements from PR #54.
- Clone functionality from the `OpenLIFUData` module as needed.

arhowe00:
Closes #163

- This commit features a tentative UI file for managing protocol configurations (i.e. loaded Protocol classes), along with the functionality proposed in #163. `setup()` within `OpenLIFUProtocolConfigWidget` has been configured to load the UI and dynamically display a widget for modifying the focal pattern, depending on which is loaded.

- The widget is not observing any of the buttons, and no logic is yet implemented for actually modifying buttons.


ebrahimebrahim (changes requested):
Very nice!!

Couple of things that would be good to settle while we are doing this:

- Can you add a button to the home module that takes you to the protocol config module? Similar to the other buttons that are there, just at the bottom of the stack of buttons
- The `protocolSelector` will need to show something appropriate when there are no protocols to select. The "create new protocol" text can't be that thing, since that isn't supposed to be truly a selectable option.
  - Side note: If it gets too annoying to do the "create new protocol" thing as an option in the `protocolSelector` dropdown menu then it's okay to make it a separate button instead. There is a pattern in slicer of having "create new..." be an option in dropdown menus, but it's a little weird and not necessarily trivial to implement yourself without dealing with annoying cases. Up to you!

Also, since the UI is being added here but doesn't do anything yet, we should indicate that:

- Add some indication, such a big red QLabel or something like that, to the protocol configuration module that makes it clear that this is just a dummy UI and the module is not implemented yet. Just in case one of our releases has the content of this PR but not the upcoming logic implementation PR

arhowe00:
> Side note: If it gets too annoying to do the "create new protocol" thing as an option in the protocolSelector dropdown menu then it's okay to make it a separate button instead. [...] Up to you!

I notice what you mean as I imagined how I would _actually implement_ this selection. I was thinking of what is done for "Select a Text" in the Texts module. However, since you made this (simpler) suggestion, I'll go with it!

---

Now, since I did some digging and it's fresh in my mind, here's for the public good:

The Combo Box widget in the Texts module, which was the pattern I was interested in ("Select a ..."), is not just a QComboBox. The Combo Box in the Texts module is instead a qMRMLNodeComboBox, which is a more "seasoned" QComboBox that generalizes the functionality of the widget pattern (Dropdown with "Select a ___"; "Create new ___") and how it interacts with the scene. Buttons such as "Create new <node>" are automatically created for modules when using widgets of that type, and Modules can conditionally activate other selections such as "Delete current <node>".  Those buttons, along with the text for them, are defined in the PIMPL for qMRMLNodeComboBox, generalizable to any node. The qMRMLNodeComboBox also directly fetches all the associated MRML nodes (e.g. Text nodes) from the scene, so none of that is managed in the Texts module widget itself, other than specifying which node the qMRMLNodeComboBox is used for. In our specific case, doing this would either mean reinventing the logic in qMRMLNodeComboBox or making a MRML node and attaching it to the scene. 

- The header is in `Libs/MRML/Widgets/qMRMLNodeComboBox.h`
- PIMPL is in `Libs/MRML/Widgets/qMRMLNodeComboBox.cxx`.

arhowe00:
> Very nice!!
> 
> Couple of things that would be good to settle while we are doing this:
> 
> * Can you add a button to the home module that takes you to the protocol config module? Similar to the other buttons that are there, just at the bottom of the stack of buttons
> * The `protocolSelector` will need to show something appropriate when there are no protocols to select. The "create new protocol" text can't be that thing, since that isn't supposed to be truly a selectable option.
>   
>   * Side note: If it gets too annoying to do the "create new protocol" thing as an option in the `protocolSelector` dropdown menu then it's okay to make it a separate button instead. There is a pattern in slicer of having "create new..." be an option in dropdown menus, but it's a little weird and not necessarily trivial to implement yourself without dealing with annoying cases. Up to you!
> 
> Also, since the UI is being added here but doesn't do anything yet, we should indicate that:
> 
> * Add some indication, such a big red QLabel or something like that, to the protocol configuration module that makes it clear that this is just a dummy UI and the module is not implemented yet. Just in case one of our releases has the content of this PR but not the upcoming logic implementation PR

I've added these items. Here's what things look like.
image

ebrahimebrahim:
Nice! I like the big red text :laughing:. Re-request review when ready for review

ebrahimebrahim (changes requested):
Very nice, but one more small thing: the protocol selector just isn't feeling right to me.

Look at the UI and pretend you don't already know what the controls will do. The protocol selector sort of just feels like another parameter to control right? It doesn't feel like it's at a "higher level" than everything, controlling _which_ protocol is being view/edited. Really the protocol selector is the same "level" as the load/save/delete/create buttons, but it looks like it's living among the individual controls within a protocol.

Can you reorganize a bit to address this?

arhowe00:
Sounds good.  I decided to use the ctkComboBox which does handle a lot of the headaches involved with managing default text in the QComboBox directly, and it also provides some extra functions that can be used but may not be.

Here's the header for ctkComboBox: https://github.com/commontk/CTK/blob/master/Libs/Widgets/ctkComboBox.h

ebrahimebrahim (approved):
Awesome!!

PR #167, opened 2025-01-30
Populate photoscan input combobox during active session

### LINKED ISSUE #160: Populate photoscan dropdown in transducer tracking inputs when there is an active session
In the manual workflow, the Photoscan combo box in the transducer tracking module is popuated with any SlicerOpenLIFUPhotoscan objects that have been manually loaded into the Slicer scene via the data module, as tracked by 'loaded_photoscans' in the data parameter node. This was addressed in 

When there is an active session, the the dropdown should list all of the photoscans associated with the session (even if they have not been loaded into Slicer yet). The list of photoscans associated with a session is found using the `get_photoscan_ids` function in the database. Since only the data module has information regarding the database path, this information needs to be stored as part of SlicerOpenLIFUSession when a session is loaded, so that it can be accessed by the transducer tracking module. 

Depends on #143

sadhana-r:
Closes #160 

Depends on https://github.com/OpenwaterHealth/OpenLIFU-python/pull/199


ebrahimebrahim:
for the record the reason for closing this is that it will be part of #158 instead

PR #170, opened 2025-02-05
Implement caching for WIPs in the protocol config

### LINKED ISSUE #169: Cache unsaved protocols
Depends on #164 

> I've mostly addressed the problem of choosing when to write something in memory and in the database, which you can see in the ongoing commits.  I'm pretty much through implementing all the buttons, but I'm now readdressing the issue of creating new protocols. I think it makes intuitive sense for newly created protocols to be selectable but unsaved to both memory and the database, and for the user to have as many unsaved protocols in progress that they want as long as the application is open.  I have an idea of how I would address this, but I want to know whether implementing this at all is the way I should go (I could just make it so that "Create New Protocol" doesn't do anything to the selector until it is saved in the database and loaded).  If I go the route that I think makes sense, it would involve storing a list of "ephemeral" protocols in the logic memory, but not in the application memory or database. They can be selected from the drop down and maybe are appended with (UNSAVED).
> 
> To think about this, it might lead to making every protocol with desired edits ephemeral and unsaved until the save button is pressed.  That way somebody can edit a protocol, switch to another with unsaved changes, and switch back. The "Saved changes" state machine will be easy to implement with this, but it will require a sort-of caching mechanism for unsaved protocols.
> 
> Thoughts? 

 _Originally posted by @arhowe00 in [#164](https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/164#issuecomment-2631762856)_

arhowe00:
Closes #169

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1533,7 +1533,7 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
 
         # check if user wants to overwrite WIPs
         protocolConfigLogic = slicer.util.getModuleLogic('OpenLIFUProtocolConfig')
-        if not protocolConfigLogic.confirm_and_overwrite_protocol_cache(protocol.id):
+        if not protocolConfigLogic.confirm_and_overwrite_protocol_cache(protocol):
             return
 
         self.getParameterNode().loaded_protocols[protocol.id] = SlicerOpenLIFUProtocol(protocol)
@@ -1546,7 +1546,7 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
         if not protocol_id in loaded_protocols:
             raise IndexError(f"No protocol with ID {protocol_id} appears to be loaded; cannot remove it.")
 
-        # check if user wants to save changes if cached
+        # check if user wants to save changes
         protocolConfigLogic = slicer.util.getModuleLogic('OpenLIFUProtocolConfig')
         if protocolConfigLogic.protocol_id_is_in_cache(protocol_id):
             if slicer.util.confirmYesNoDisplay(
@@ -1555,10 +1555,10 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
             ):
                 protocolConfigLogic.save_protocol_to_database(protocolConfigLogic.cached_protocols[protocol_id])
 
-            # Regardless of save, we must delete from cache
-            self.delete_protocol_from_cache(protocol.id)
 
         loaded_protocols.pop(protocol_id)
+        # We must delete from cache after because parameter node update might add it back to cache
+        protocolConfigLogic.delete_protocol_from_cache(protocol_id)
 
     def load_transducer_from_file(self, filepath:str) -> None:
         transducer = openlifu_lz().Transducer.from_file(filepath)
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1532,7 +1532,8 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
                 return
 
         # check if user wants to overwrite WIPs
-        if not self.confirm_and_overwrite_protocol_cache(protocol.id):
+        protocolConfigLogic = slicer.util.getModuleLogic('OpenLIFUProtocolConfig')
+        if not protocolConfigLogic.confirm_and_overwrite_protocol_cache(protocol.id):
             return
 
         self.getParameterNode().loaded_protocols[protocol.id] = SlicerOpenLIFUProtocol(protocol)
@@ -1546,29 +1547,11 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
             raise IndexError(f"No protocol with ID {protocol_id} appears to be loaded; cannot remove it.")
 
         # check if user wants to overwrite WIPs
-        if not self.confirm_and_overwrite_protocol_cache(protocol_id):
+        protocolConfigLogic = slicer.util.getModuleLogic('OpenLIFUProtocolConfig')
+        if not protocolConfigLogic.confirm_and_overwrite_protocol_cache(protocol_id):
             return
 
         loaded_protocols.pop(protocol_id)
-
-    def confirm_and_overwrite_protocol_cache(self, protocol_id: str) -> bool:
-        """
-        Checks if the protocol ID exists in the protocol config cache. If so,
-        prompts the user to confirm overwriting it. Returns False if the user
-        cancels, otherwise updates the cache and returns True.
-        """
-        # Check protocol config for a WIP
-        protocolConfigLogic = slicer.util.getModuleLogic('OpenLIFUProtocolConfig')
-
-        if protocolConfigLogic.protocol_id_is_in_cache(protocol_id):
-            if not slicer.util.confirmYesNoDisplay(
-                text=f"You have unsaved changes in a protocol with the same ID. Are you sure you want to discard changes?",
-                windowTitle="Discard Changes Confirmation",
-            ):
-                return False # User canceled discarding unsaved protocol changes
-
-            self.logic.delete_protocol_from_cache(protocol_id)
-        return True
 
     def load_transducer_from_file(self, filepath:str) -> None:
         transducer = openlifu_lz().Transducer.from_file(filepath)
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1546,10 +1546,17 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
         if not protocol_id in loaded_protocols:
             raise IndexError(f"No protocol with ID {protocol_id} appears to be loaded; cannot remove it.")
 
-        # check if user wants to overwrite WIPs
+        # check if user wants to save changes if cached
         protocolConfigLogic = slicer.util.getModuleLogic('OpenLIFUProtocolConfig')
-        if not protocolConfigLogic.confirm_and_overwrite_protocol_cache(protocol_id):
-            return
+        if protocolConfigLogic.protocol_id_is_in_cache(protocol_id):
+            if slicer.util.confirmYesNoDisplay(
+                text=f"You have unsaved changes in the protocol you are about to remove. Do you want to save changes?",
+                windowTitle="Save Changes Confirmation",
+            ):
+                protocolConfigLogic.save_protocol_to_database(protocolConfigLogic.cached_protocols[protocol_id])
+
+            # Regardless of save, we must delete from cache
+            self.delete_protocol_from_cache(protocol.id)
 
         loaded_protocols.pop(protocol_id)
 
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1514,7 +1514,9 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
 
     def load_protocol_from_openlifu(self, protocol:"openlifu.Protocol", replace_confirmed: bool = False) -> None:
         """Load an openlifu protocol object into the scene as a SlicerOpenLIFUProtocol,
-        adding it to the list of loaded openlifu objects.
+        adding it to the list of loaded openlifu objects. If there are
+        changes in the protocol config, also confirms user wants to discard
+        changes.
 
         Args:
             protocol: The openlifu Protocol object
@@ -1528,14 +1530,45 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
                 "Protocol already loaded",
             ):
                 return
+
+        # check if user wants to overwrite WIPs
+        if not self.confirm_and_overwrite_protocol_cache(protocol.id):
+            return
+
         self.getParameterNode().loaded_protocols[protocol.id] = SlicerOpenLIFUProtocol(protocol)
 
     def remove_protocol(self, protocol_id:str) -> None:
-        """Remove a protocol from the list of loaded protocols."""
+        """Remove a protocol from the list of loaded protocols. If there are
+        changes in the protocol config, also confirms user wants to discard
+        changes."""
         loaded_protocols = self.getParameterNode().loaded_protocols
         if not protocol_id in loaded_protocols:
             raise IndexError(f"No protocol with ID {protocol_id} appears to be loaded; cannot remove it.")
+
+        # check if user wants to overwrite WIPs
+        if not self.confirm_and_overwrite_protocol_cache(protocol_id):
+            return
+
         loaded_protocols.pop(protocol_id)
+
+    def confirm_and_overwrite_protocol_cache(self, protocol_id: str) -> bool:
+        """
+        Checks if the protocol ID exists in the protocol config cache. If so,
+        prompts the user to confirm overwriting it. Returns False if the user
+        cancels, otherwise updates the cache and returns True.
+        """
+        # Check protocol config for a WIP
+        protocolConfigLogic = slicer.util.getModuleLogic('OpenLIFUProtocolConfig')
+
+        if protocolConfigLogic.protocol_id_is_in_cache(protocol_id):
+            if not slicer.util.confirmYesNoDisplay(
+                text=f"You have unsaved changes in a protocol with the same ID. Are you sure you want to discard changes?",
+                windowTitle="Discard Changes Confirmation",
+            ):
+                return False # User canceled discarding unsaved protocol changes
+
+            self.logic.delete_protocol_from_cache(protocol_id)
+        return True
 
     def load_transducer_from_file(self, filepath:str) -> None:
         transducer = openlifu_lz().Transducer.from_file(filepath)
--- a/OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
+++ b/OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
@@ -533,29 +533,16 @@ class OpenLIFUProtocolConfigWidget(ScriptedLoadableModuleWidget, VTKObservationM
             qsettings.value('OpenLIFU/databaseDirectory','.'), # starting dir, with default of '.'
             "Protocols (*.json);;All Files (*)", # file type filter
         )
-        if filepath:
-            protocol = openlifu_lz().Protocol.from_file(filepath)
+        if not filepath:
+            return
 
-            if protocol.id in self.logic.cached_protocols:
-                if not slicer.util.confirmYesNoDisplay(
-                    text = f"You have a unsaved changes in a protocol with the same ID. Discard and load the new one?",
-                    windowTitle = "Discard Changes Confirmation",
-                ):
-                    return
+        protocol = openlifu_lz().Protocol.from_file(filepath)
 
-                #  This does same as revert changes, but for an arbitrary id
-                self.logic.cached_protocols.pop(protocol.id, None)  # remove from cache
-                if protocol.id in self.logic.new_protocol_ids:
-                    self.logic.new_protocol_ids.discard(protocol.id)
-                self.updateWidgetSaveState(SaveState.NO_CHANGES)
-                self.reloadProtocols()
+        if not self.load_protocol_from_openlifu(protocol):
+            return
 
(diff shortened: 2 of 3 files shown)


ebrahimebrahim:
> There is an edge case which requires manually caching the protocol.
   If the user loads a protocol, makes edits, and then tries to load a
   protocol with the same exact ID, they will be trying to overwrite a
   protocol that is not in the cache, but is marked as UNSAVED_CHANGES.
   Protocols are cached when changing selections, not when
   UNSAVED_CHANGES is marked, because caching a protocol on every
   keypress is expensive (or, is it? @ebrahimebrahim, should I really
   cache the WIP every single time a key is pressed?). Since we need
   confirmation that the user wants to overwrite unsaved changes in the
   currently loaded WIP of the same ID they are trying to load, we cache
   it and then ask.

@arhowe00 good catch with this edge case! I agree that it makes sense to ask the user whether they really meant to overwrite what is in the WIP protocol in this case.

I agree that caching should occur when changing selections (or something like that), not on every keypress

ebrahimebrahim:
@sadhana-r setting you as reviewer on this one, ty!

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -649,7 +739,7 @@ def setAllWidgetsEnabled(self, enabled: bool) -> None:
 
         self.ui.protocolDatabaseSaveButton.setEnabled(enabled)
         self.ui.protocolDatabaseDeleteButton.setEnabled(enabled)
-        self.ui.protocolEditButton.setEnabled(enabled)
+        self.ui.protocolEditRevertDiscardButton.setEnabled(enabled)
 
     def cache_protocol(self, protocol_changes: "openlifu.plan.Protocol") -> None:
         self.logic.cached_protocols[self._cur_protocol_id] = protocol_changes
```
sadhana-r, line 745: When caching a protocol, we need to account for the case where the user changes the protocol id. When that happens, `self._cur_prototocol_id` and the id keys in `self.logic.cached_protocols` use the 'original' protocol id, but `protocol.id` will be different. This becomes an issue for new protocols, for example the check at line 376. In the case I tested, if I create a new protocol, change the ID , load a session in the data module (which causes the new protocol to get cached and reloads the list), and then select the new protocol in the drop down, it isn't recognized correctly as a new protocol i.e. the `Delete Protocol From Database` option is enabled. 

The fix might be to just replace the original ID in the list of `new_protocol_ids` and include the new one, if the id changes. I think the key in `cached_protocols` would need to be changed as well? But then the user could potentially change to the ID to something that already exists.  Let me know your thoughts. It would be good to think about any other situations where a change to the ID may break things.

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -187,6 +187,7 @@ def __init__(self, parent=None) -> None:
         VTKObservationMixin.__init__(self)  # needed for parameter node observation
         self.logic: Optional[OpenLIFUProtocolConfigLogic] = None
         self._is_saving_changes: bool = False
+        self._is_updating_display: bool = False
```
sadhana-r, line 190: I think it would be helpful to add some comments/documentation explaining why these flags are needed. Is it becase the updating/saving steps takes some time and certain actions shouldn't take place while they are happening?

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -493,11 +522,36 @@ def onLoadProtocolFromFileClicked(self, checked:bool) -> None:
...
+                self.logic.cached_protocols.pop(protocol.id, None)  # remove from cache
+                if protocol.id in self.logic.new_protocol_ids:
+                    self.logic.new_protocol_ids.discard(protocol.id)
+                self.updateWidgetSaveState(SaveState.NO_CHANGES)
+                self.reloadProtocols()
+
+                self.logic.dataLogic.load_protocol_from_openlifu(protocol, replace_confirmed=True)  # load to memory
+            else:
+                self.logic.dataLogic.load_protocol_from_openlifu(protocol, replace_confirmed=False)  # load to memory
+
```
sadhana-r, line 543: There is some repetition of code between `onLoadProtocolFromDataClicked` and `onLoadProtocolFromFileClicked` suggeting that the code can be modularized. You could create a function here called `load_protocol_from_openlifu` that calls  `self.logic.dataLogic.load_protocol_from_openlifu(protocol)` and includes a check on the cached_protocols. 

Also, there is also the case where a user is editing a protocol, and then loads a session with that same protocol id. In that case, when the the protocol is loaded in, we should probably also be asking the user if they want to discard and load the new one?

sadhana-r (changes requested):
This module is looking great! good job! I have added some minor comments and noted a couple more edge cases to handle.

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -510,22 +564,56 @@ def onLoadProtocolFromDatabaseClicked(self, checked:bool) -> None:
             selected_protocol_id = dialog.get_selected_protocol_id()
```
sadhana-r, line 564: I also noticed that this dialog does not block the main window but I think its meant to since its Modal? Maybe becase the parent is not being set?

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
arhowe00: Yes, currently thinking on this. The overarching idea was that the entire concept of a wip is stored in the protocol object state as saved from the gui, everything else that is recorded and indexed (in `cached_protocols`, `new_protocol_ids`, `_cur_protocol_id`, is the "official" id as captured when the unique id was made for a new one, or the id was loaded from something in memory/a file.  Thinking about how I can keep this convention and deal with these edge cases at the moment.

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
arhowe00: Sounds good -- I will update documentation around this.  For more information as to why I chose the flag approach as opposed to disabling signals, check 2394d24d78c0fe9cbf7997a3db4807cccf92fef5

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
arhowe00: > Also, there is also the case where a user is editing a protocol, and then loads a session with that same protocol id. In that case, when the the protocol is loaded in, we should probably also be asking the user if they want to discard and load the new one?

This is probably a case we should handle, but I don't know if there's a convenient solution. We have our new WIP system with caching, but external actors can load protocols into the DB that we then reload from in here.  To address this we can either 1) add manual confirmations in the location(s) which people are loading protocols from (which might be wieldy, as it would then require other people checking in with our module before doing stuff with the db), or 2) assume implicit overwrite in this odd case that a session is loaded with the same id as a wip protocol.  I don't know whether there is a super convenient solution, because caches and wips are a thing for this module, but other people who are interacting with the db will clash there. @ebrahimebrahim  any insights on how to deal with this?

arhowe00: Might be a bit of a disruption, but what if we completely decoupled protocol configuration from loading and selecting? It seems that what we have is an editor, nothing to do with actually using protocols, in this module.

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
arhowe00: > When caching a protocol, we need to account for the case where the user changes the protocol id. When that happens, `self._cur_prototocol_id` and the id keys in `self.logic.cached_protocols` use the 'original' protocol id, but `protocol.id` will be different. This becomes an issue for new protocols, for example the check at line 376. In the case I tested, if I create a new protocol, change the ID , load a session in the data module (which causes the new protocol to get cached and reloads the list), and then select the new protocol in the drop down, it isn't recognized correctly as a new protocol i.e. the `Delete Protocol From Database` option is enabled.
> 
> The fix might be to just replace the original ID in the list of `new_protocol_ids` and include the new one, if the id changes. I think the key in `cached_protocols` would need to be changed as well? But then the user could potentially change to the ID to something that already exists. Let me know your thoughts. It would be good to think about any other situations where a change to the ID may break things.

I actually think this is related to the problem below -- that the session protocol loading and load protocol buttons from external modules aren't bound to the same checks that the loading functions here are.  Loading stuff in this module will check with the WIP protocols and make sure to cache new protocols before loading new ones, and these checks are evaded because other parts of the app are loading directly into the database. Something we could do, since caching is now an extra arm of the protocol config module, is have other buttons of the app load their protocols through this module. That way they don't have to know about our cache, but they do have to know about our module, if they want to load protocols.

ebrahimebrahim: > have other buttons of the app load their protocols through this module

Not a fan of this, as it breaks some of the symmetry of the application. Only protocols need their own editor, not transducers, etc. The loading of protocols in other parts of the app ought to keep its parallelism with the loading of other things

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
ebrahimebrahim: > Might be a bit of a disruption, but what if we completely decoupled protocol configuration from loading into memory and selecting protocols? It seems that what we have is an editor, nothing to do with actually using protocols, in this module. (but we shouldn't let perfect be the enemy of good )

It does sound a bit too disruptive...

I do like the way it interacts with other things right now. For example if you load a session, then the protocol of that session is loaded in. Then in the protocol config module you see that protocol there, and it's probably the exact one you'd want to mess with.

ebrahimebrahim: As for handling the case Sadhana mentioned above, it's a good catch. Asking the user if they want to discard and load the new protocol seems to make sense

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
arhowe00: > > have other buttons of the app load their protocols through this module
> 
> Not a fan of this, as it breaks some of the symmetry of the application. Only protocols need their own editor, not transducers, etc. The loading of protocols in other parts of the app ought to keep its parallelism with the loading of other things

Right, but in defense of this idea, the way that protocols are loaded into memory from files and the database are currently handled in 2 locations, the data module and protocol config module. At one point, when building the protocol config module, I actually called `load_protocol_from_openlifu` _and_ `load_protocol_from_file` from within the data logic, while making the protocol config module.  However, now there are some checks that only the protocol config module does, related to the cache and WIP protocols, before loading protocols into memory. This is because the selectable protocols are tied to what is loaded in memory, so what is loaded in memory must be vetted to make sure it isn't destroying any progress on new protocols. 

If we want to keep the act of loading a protocol into memory within the data module, then might it make sense to have it check the cache as well?

ebrahimebrahim: Hmm, that is convincing. So the protocol config module takes over the protocol loading code so it can do all its checks, and the data module uses that. Yes I think it's reasonable, Sadhana has been looking at all this in more detail and may have some comments but I take back being against that type of reorganization :smile:

sadhana-r: Hmm either way there will have to be cross over between the data module and protocol config module - either the data module calls checks against the cache in protocol config or protcol config is calling the load functions from the data module. My thought is that in terms of overall organization, its nice to have all data loading aspects contained within the data module. Maybe within `load_protocol_from_openlifu` in the data module, you could add a function call to the protocol config module's logic that does the required checks against the cached and new protocol ids and returns a True/False? The data module doesn't need to include all logic for checking. And the protocol config can then just call the data module for loading. What do you think of that idea?

ebrahimebrahim: > Maybe within `load_protocol_from_openlifu` in the data module, you could add a function call to the protocol config module's logic that does the required checks against the cached and new protocol ids and returns a True/False? The data module doesn't need to include all logic for checking. And the protocol config can then just call the data module for loading. What do you think of that idea?

:+1: :+1: :+1: !

arhowe00:
Okay, I was able to resolve the issue of interacting with protocols that have wips from other modules, but there is one behavior that may or may not be preferred.  When unloading a session, the confirmation of unloading a protocol if it has a wip works, but the session does unload. In the case that a user does this, the wip and loaded protocol remain loaded, but the session itself is unloaded. This can change by modifying the session unload to *first* check for wips. To see what I mean:

1. Load example session from Data module
2. Start a wip by editing and keeping unsaved changes in the editor
3. Switch back to Data module and unload session
4. If you select that you don't want to delete your wip, the session will unload, but the protocol will remain loaded. This may be fine or not preferred, let me know.

OpenLIFUData/OpenLIFUData.py
```diff
@@ -1528,15 +1530,46 @@ def load_protocol_from_openlifu(self, protocol:"openlifu.Protocol", replace_conf
...
+        if protocolConfigLogic.protocol_id_is_in_cache(protocol_id):
+            if not slicer.util.confirmYesNoDisplay(
+                text=f"You have unsaved changes in a protocol with the same ID. Are you sure you want to discard changes?",
+                windowTitle="Discard Changes Confirmation",
+            ):
+                return False # User canceled discarding unsaved protocol changes
+
+            self.logic.delete_protocol_from_cache(protocol_id)
+        return True
+
```
sadhana-r, line 1572: I don't think you need to repeat this function here when it is in the protocol config module. Why not just call the protocolConfigLogic directly in `load_protocol_from_openlifu` and `remove_protocol`.

OpenLIFUData/OpenLIFUData.py
```diff
@@ -1528,15 +1530,46 @@ def load_protocol_from_openlifu(self, protocol:"openlifu.Protocol", replace_conf
...
-        """Remove a protocol from the list of loaded protocols."""
+        """Remove a protocol from the list of loaded protocols. If there are
+        changes in the protocol config, also confirms user wants to discard
+        changes."""
         loaded_protocols = self.getParameterNode().loaded_protocols
         if not protocol_id in loaded_protocols:
             raise IndexError(f"No protocol with ID {protocol_id} appears to be loaded; cannot remove it.")
+
+        # check if user wants to overwrite WIPs
+        if not self.confirm_and_overwrite_protocol_cache(protocol_id):
```
sadhana-r, line 1549: Directly call the protocol config logic here

OpenLIFUData/OpenLIFUData.py
```diff
@@ -1528,15 +1530,46 @@ def load_protocol_from_openlifu(self, protocol:"openlifu.Protocol", replace_conf
...
+        protocolConfigLogic = slicer.util.getModuleLogic('OpenLIFUProtocolConfig')
+
+        if protocolConfigLogic.protocol_id_is_in_cache(protocol_id):
+            if not slicer.util.confirmYesNoDisplay(
+                text=f"You have unsaved changes in a protocol with the same ID. Are you sure you want to discard changes?",
+                windowTitle="Discard Changes Confirmation",
+            ):
+                return False # User canceled discarding unsaved protocol changes
+
+            self.logic.delete_protocol_from_cache(protocol_id)
```
sadhana-r, line 1570: Also this isn't correct - this should be protocol config logic? Either way, don't think this function is needed here.

sadhana-r (changes requested):

sadhana-r:
> Okay, I was able to resolve the issue of interacting with protocols that have wips from other modules, but there is one behavior that may or may not be preferred. When unloading a session, the confirmation of unloading a protocol if it has a wip works, but the session does unload. In the case that a user does this, the wip and loaded protocol remain loaded, but the session itself is unloaded. This can change by modifying the session unload to _first_ check for wips. To see what I mean:
> 
> 1. Load example session from Data module
> 2. Start a wip by editing and keeping unsaved changes in the editor
> 3. Switch back to Data module and unload session
> 4. If you select that you don't want to delete your wip, the session will unload, but the protocol will remain loaded. This may be fine or not preferred, let me know.

Can we make it so that on unload, it asks "Do you want to save changes"? So it behaves like how other editors would - it either saves and unloads, or reverts changes and then unloads.

OpenLIFUData/OpenLIFUData.py
arhowe00: Sounds good -- I repeated it because it uses GUI elements and I didn't feel right about asking for the protocol config **widget**, only the logic. But I'll just call the logic in `load_protocol_from_openlifu` and `remove_protocol` as suggested here.

arhowe00: Sorry, I meant I'll put it into the protocol config logic and then call it from the data module.

OpenLIFUData/OpenLIFUData.py
arhowe00: Yes -- your point is correct

arhowe00:
> Can we make it so that on unload, it asks "Do you want to save changes"? So it behaves like how other editors would - it either saves and unloads, or reverts changes and then unloads.

👍 good idea.   I am implementing this as part of the `remove_protocol` function, as that is what is called when a session is unloaded. It is also, seemingly, the only way to _unload_ a protocol in the entire app.

arhowe00:
Thanks for the comments and suggestions Sadhana. Let me know what you think!

sadhana-r:
> I just realized that we don't know whether it was opened as a file or in the database. I'll assume database, since this pertains mainly to the session-based workflow, and because it's in the data module that this occurs which really only has features enabled with database loaded.

Yep in a session based workflow its safe to assume database. If the same protocol_id was loaded from file initially, then loading the session would have reloaded the protocol from the database anyways.

sadhana-r:
I am wondeirng the cache check needs to be combined with the `replace_confirmed` check for protocols. i.e. instead of having two levels of checks/message boxes, we have a single message box that determines the appropriate action. Its a little weird that if I am editing a protocol, and then load a session with the same protocol id, then currently

1. I get prompted with "Loading this session will replace the already loaded protocol with ID XX. Proceed?"
2. If I say yes, then I get a second message that says "You have unsaved changes in a protocol with the same ID as the protocol you are trying to load. Discard and load the new one?"
3. And if I say no, then the session is loaded but the WIP protocol remains. This seems unintuitive since I confirmed replacement in the first step.

Can these checks be combining such that the message in the box varies based on whether the protocol is exists in the scene/cache.  So 
1. If its in the scene but not cache, the original message is fine.
2. If its in the scene and also in the cache, will prompt with the replacement messge but also mention the unsaved changes and ask if the user wants to save changes?

Sorry, maybe I am complicating things. We can also keep it as is for now and fine tune these workflows later if we want. @ebrahimebrahim do you have thoughts?

ebrahimebrahim:
> I am wondeirng the cache check needs to be combined with the `replace_confirmed` check for protocols. i.e. instead of having two levels of checks/message boxes, we have a single message box that determines the appropriate action. Its a little weird that if I am editing a protocol, and then load a session with the same protocol id, then currently
> 
> 1. I get prompted with "Loading this session will replace the already loaded protocol with ID XX. Proceed?"
> 2. If I say yes, then I get a second message that says "You have unsaved changes in a protocol with the same ID as the protocol you are trying to load. Discard and load the new one?"
> 3. And if I say no, then the session is loaded but the WIP protocol remains. This seems unintuitive since I confirmed replacement in the first step.
> 
> Can these checks be combining such that the message in the box varies based on whether the protocol is exists in the scene/cache. So
> 
> 1. If its in the scene but not cache, the original message is fine.
> 2. If its in the scene and also in the cache, will prompt with the replacement messge but also mention the unsaved changes and ask if the user wants to save changes?
> 
> Sorry, maybe I am complicating things. We can also keep it as is for now and fine tune these workflows later if we want. @ebrahimebrahim do you have thoughts?

I agree that fixing this case is needed. I think with a current behavior, it is a bug. It's an edge case, but if a user goes through it they will end up confused about what happened.

But yes I also think we should accept it for this PR and open an issue (the description of which could be exactly your comment) to handle it later

sadhana-r (approved):

Issue #164, opened 2025-01-28
Implement protocol config module logic


arhowe00:
## Description

Enhance the `OpenLIFUProtocolConfig` module by implementing the following:

- Map widget buttons to new logic functions for protocol management.
- Add logic functions to get/set protocol values.
- Connect UI widgets to logic functions, including:
  - Populate a dropdown menu in the protocol configuration module with currently loaded protocols (from the `OpenLIFUData` module’s parameter node), including a "create new" option.
  - Implement save protocol functionality with overwrite dialog support.
  - Add delete protocol functionality.
  - Dynamically adjust widget availability based on values of certain widgets (e.g., `FocalPattern` determines available focal pattern parameters).

### References

- Depends on #163 and https://github.com/OpenwaterHealth/OpenLIFU-python/issues/196

arhowe00:
When a user is editing a selected protocol in the config, there are two possible behaviors:
1. The selected protocol is edited and, in real time, the protocol in memory changes to reflect what is configured. This would occur even if the user does not hit "Save Protocol."  Then, if the user hits "Save Protocol," the selected protocol will be written to the database.
2. The selected protocol is edited, but nothing changes in memory, nor in the database, until "Save Protocol" is hit.  This means that the configuration would be adjusted until desired, and the protocol in memory only changes when "Save Protocol" is hit, which will invoke two functionalities: a) modifying the protocol loaded in memory and b) writing that protocol to the database.
Which of these behaviors do you think I should implement @ebrahimebrahim ?

arhowe00:
Similar to above, what someone hits "Create new Protocol," should anything be written to the DB?  What about protocols loaded in memory?

ebrahimebrahim:
> When a user is editing a selected protocol in the config, there are two possible behaviors:
> 
> 1. The selected protocol is edited and, in real time, the protocol in memory changes to reflect what is configured. This would occur even if the user does not hit "Save Protocol."  Then, if the user hits "Save Protocol," the selected protocol will be written to the database.
> 2. The selected protocol is edited, but nothing changes in memory, nor in the database, until "Save Protocol" is hit.  This means that the configuration would be adjusted until desired, and the protocol in memory only changes when "Save Protocol" is hit, which will invoke two functionalities: a) modifying the protocol loaded in memory and b) writing that protocol to the database.

There's
- (A) protocol on disk (in openlifu database)
- (B) protocol in memory (serialized into the data module's parameter node)
[...]

Issue #79, opened 2024-09-04
Enable session creation


ebrahimebrahim:
Button to create session, raises dialog, mutates database. Also loads the session so that it becomes now the active session.

[Relevant slide](https://docs.google.com/presentation/d/1yH7Nh0XqOb4q_BCZdTR4L2R6TGPShxaMDx2XHCmvxSg/edit#slide=id.g2f2d4b7456b_0_187)

Depends on https://github.com/OpenwaterHealth/OpenLIFU-python/issues/92 and https://github.com/OpenwaterHealth/OpenLIFU-python/issues/93

PR #178, opened 2025-02-12
Make protocol configuration module work without a database

### LINKED ISSUE #171: Make protocol configuration module work without a database
Make the protocol configuration module work without a database

Also make it consistent in whether it's based on database actions or file actions. It's weird that we do file-based loading only and db-based saving only. What's missing then is db-based loading and file-based saving.

- allow the protocol editor to be used without a database connection
- rename the current "save" and "delete" to "save to database" "delete to database" to make it clear they are database actions
- rename the current "load protocol" to "load from file" to make it clear it's a file action
- add a "load from database" button and associated dialog that allows you to select protocols from the list of IDs or name/ID pairs (this one is the most work because you'd be creating a custom dialog)
- add a "save to file" button with a `ctkLineEdit` to specify where to save

Depends on #164

arhowe00:
Closes #171 

Note that this did not use the `ctkLineEdit` when letting the user choose a filepath to save, but everything seemed to work with a similar dialog used by the "Load Protocol From File" button.


OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -484,19 +599,36 @@ def setNewProtocolWidgets(self) -> None:
 
     def setProtocolEditorEnabled(self, enabled: bool) -> None:
         self.ui.protocolEditorSectionGroupBox.setEnabled(enabled)
-        if enabled and get_openlifu_data_parameter_node().database_is_loaded:
-            self.setDatabaseSaveAndDeleteEnabled(True)
-        elif not enabled:
-            self.setDatabaseSaveAndDeleteEnabled(False)
+        self.setAllSaveAndDeleteButtonsEnabled(enabled)
+        if not get_openlifu_data_parameter_node().database_is_loaded:
```
ebrahimebrahim, line 603: I noticed that you added `database_is_loaded` to the `OpenLIFUDataLogic`'s parameter node. I don't think it's needed, because you can instead just check whether `OpenLIFUDataLogic`'s `db` attribute is `None` to determine whether the database is loaded. If you agree, can you remove `database_is_loaded` and do that instead? Better not to maintain two separate representations of the same state and invite more bugs.

For example here is a probable bug: if you save and then load the slicer scene, the `database_is_loaded` would probably retain its previous state which may have been true, but the database "connection" would have been lost actually and `OpenLIFUDataLogic`'s `db` attribute would be `None`

ebrahimebrahim: (btw such a change could be lumped into this PR, but would just reference a different issue like probably #164)

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -108,6 +109,67 @@ class OpenLIFUProtocolConfigParameterNode:
 
     """
 
+#
+# OpenLIFUProtocolConfigDialogs
+#
+
+class ProtocolSelectionFromDatabaseDialog(qt.QDialog):
```
ebrahimebrahim, line 116: can you make this dialog start out a bit wider? wide enough to fit a typical protocol name and id entry

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -108,6 +109,67 @@ class OpenLIFUProtocolConfigParameterNode:
...
+        self.selected_protocol = None
+
+        self.setup()
+
+    def setup(self):
+
+        self.boxLayout = qt.QVBoxLayout()
+        self.setLayout(self.boxLayout)
+
+        self.listWidget = qt.QListWidget(self)
```
ebrahimebrahim, line 138: It's pretty intuitive to double click an item in this listWidget to try and make and accept that selection in one go. If that's not just me and you agree it's an intuitive thing to do, then I suggest making double click work that way.

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -396,6 +495,22 @@ def onLoadProtocolFromFileClicked(self, checked:bool) -> None:
...
+
+        if not filepath:
+            return  # TODO: should emit error?
+
+        # Open the protocol selection dialog
+        dialog = ProtocolSelectionFromDatabaseDialog(self.logic.dataLogic)
+        if dialog.exec_() == qt.QDialog.Accepted:
+            selected_protocol = dialog.get_selected_protocol()
+            if selected_protocol:
+                self.logic.dataLogic.load_protocol_from_openlifu(selected_protocol)
```
ebrahimebrahim, line 512: I think that after loading a protocol, it makes sense to then make that become the currently selected protocol. Like I just loaded it, so I probably want to see its contents in my face rather than having to go and find what I loaded in the dropdown.

Same goes for loading a protocol from file -- it should become the currently selected protocol

OpenLIFUProtocolConfig/Resources/UI/OpenLIFUProtocolConfig.ui
```diff
@@ -212,7 +223,14 @@
         </widget>
       </item>
       <item>
-        <layout class="QHBoxLayout" name="protocolButtonsLayout">
+        <layout class="QHBoxLayout" name="protocolSaveDeleteButtonsLayout">
+          <item>
+            <widget class="QPushButton" name="protocolFileSaveButton">
```
ebrahimebrahim, line 228: This button is lacking a tooltip that explains what it does. The tooltip is a great opportunity to alleviate any confusion that may arise from having two different types of save buttons. And as with all our buttons: when the button is disabled, the tooltip should explain why it's disabled and/or what can be done to enable it.

ebrahimebrahim (changes requested):
A couple of additional thoughts not necessarily related to this PR:

- I noticed that when I click "edit protocol" it's kind of difficult to get out of "editing mode". I guess the upcoming "revert" button would do that? One thought I had is that the "edit protocol" button could _become_ a revert button when you click it, so it becomes like a toggle button. It makes sense because once you click "edit protocol" the button remains there but is kind of useless.
- I noticed the tooltip of the protocol selector says "Select a protocol ...". I think it could instead say "Select from among the currently loaded protocols", to help make it clear it is only showing you the loaded protocols.

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -108,6 +109,67 @@ class OpenLIFUProtocolConfigParameterNode:
...
+    """ Create new protocol selection from database dialog """
+
+    def __init__(self, dataLogic, parent=None):
+        super().__init__(slicer.util.mainWindow() if parent == "mainWindow" else parent)
+        """ Args:
+                dataLogic: instance of the data logic, needed to access the loaded database
+        """
+
+        self.setWindowTitle("Select a Protocol")
+        self.setWindowModality(1)
```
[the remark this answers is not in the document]
arhowe00: Sounds good, I was following an example from the Data module FYI

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -396,6 +495,22 @@ def onLoadProtocolFromFileClicked(self, checked:bool) -> None:
...
             self.logic.dataLogic.load_protocol_from_file(filepath)
 
+    @display_errors
+    def onLoadProtocolFromDatabaseClicked(self, checked:bool) -> None:
+        qsettings = qt.QSettings()
+
+        filepath: str = qsettings.value('OpenLIFU/databaseDirectory')  # No default value
+
+        if not filepath:
+            return  # TODO: should emit error?
```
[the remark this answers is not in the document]
arhowe00: I left a TODO in there because I was unsure if I should emit an error in the case where the button is enabled but there isn't a database location. I think if we assume the application won't erroneously call this when that would return None, we don't need these lines, but this case would just ignore the behavior and return from the call. (I guess you can also think of this of a "question inside a commit" for the reviewer, but it is also valid code)

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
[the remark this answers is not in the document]
ebrahimebrahim: Feel free to fix that too 😅

(you can use git blame to help identify an issue number to associate your fix with)

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
sadhana-r: I think I had suggested this since the protocol config module needs to watch for changes to the `db` variable and update accordingly. Since the `DataParameterNode` is being observed for any changes, having a `database_is_loaded` flag in the parameter node would ensure that the buttons get updated automatically whenever a database is loaded. . I assumed it isn't possible to add an observer to a specific attribute, but maybe I'm wrong? Or alternatively, the `DataModule` could reference `slicer.util.getModule('OpenLIFUProtocolConfig').widgetRepresentation()` and update relevant widgets? I agree that it would be better to check the `db` attribute directly, since we may want to catch changes to the `db` path instead of just `not None`.

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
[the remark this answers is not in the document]
ebrahimebrahim: You can probably find a relevant existing issue to reference That's why I suggested git blame

By using git blame I found that the line you referenced in the data module is associated to #79. it makse sense that a fix to the readability of that code could also link to #79

It can be in this PR. It's okay for a PR to hit multiple issues, and it doesn't have to try and close every issue it hits. (Just don't be too mean to the reviewer :smile: )

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
sadhana-r: Is there a specific reason why the database is stored as a `datamodulelogic` attribute and not as part of the `dataparameternode`? If not, it could be moved to the paramater node instead, kind of like `loaded_session`.

ebrahimebrahim: It's a very good question. The reason was that there's more to parameter nodes than just having convenient ways to observe for changes. Parameter nodes are really about sticking the data/state into the mrml scene. One additional consequence for example is persistence across saves.

For the `db` database connection object I was thinking of it more as _application_ state rather than _scene_ state -- by that I mean I didn't want it to persist across save and load of the slicer scene. It's more "is my application connected right now?" rather than "is a database connection currently loaded into the scene" which felt a bit weird.

(@sjh26 can chime in in case I am misrepresenting the parameter node concept here)


 >If not, it could be moved to the paramater node instead, kind of like `loaded_session`.

This is one way to do it, but we'd have to make sure we take extra measures to make it work the way we want across scene save and load.

But we don't actually care about making `db` be part of the scene... we just want to observe it for changes. So another approach is we could just create our own observer pattern. For example we did this for the "running" variable in the sonication control module -- see [here](https://github.com/OpenwaterHealth/SlicerOpenLIFU/blob/84f9972365fa225a35b4a3a109e29f0dc01bb872/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py#L358-L362).

thoughts @sadhana-r @arhowe00?

ebrahimebrahim: Oh good point about sessions. It's not right to have a session hanging around with no database. It would be unpleasant to have to go and add _another_ case to worry about in addition to manual workflow and session-based workflow -- session-based workflow with a missing database.

(A) One way to handle this is what you suggested originally: move `db` into the parameter node. If session gets to be scene-level data, then maybe `db` ought to be as well.

(B) Another way is to handle the errors. That is, allow for `db` to be `None` while there is a `loaded_session` -- the application won't work until you do connect to database, but it also won't break or crash on you. So allow for the dangling-session-with-missing-database-connect situation that arises when you save and load a slicer scene in the middle of a session based workflow. And whenever something in the session based workflow needs to use `db` but finds that it is `None`, the user is informed in an error/warning that they need to reconnect to the database to make things work properly. So for example when the photoscans dropdown is trying to populate, it sees that there is a session but no database, so it raises this error dialog (not a python exception just a friendly error dialog) and remains an empty dropdown.

(A) and (B) have their pros and cons. One advantage of (B) is that if the database ever became a remote connection, then with (B) we'd already be ready to handle connectivity issues. (A) seems a bit easier though since we'd just repeat boilerplate that we already have

I lean towards (B) and having this handled in a separate issue... but let's resolve the conversation here ... more thoughts?

sadhana-r: I think option (B) sounds good - so basically making sure we add a check for a valid database whenever `db` is referenced in the context of a session and prompting the user if its `None`. I think this would also allow for a user to potentially share a saved slicer scene with another user, who would have the same database but at a different path. Right? Whereas option (A) is hardcode to the directory structure of the initial user (this may not be relevant after the database format changes).

arhowe00:
Rebased to latest main.

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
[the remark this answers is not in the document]
ebrahimebrahim: (No objection to you adding a git-blame-ignore-revs, but it's not a big deal either way)

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
ebrahimebrahim: for the record here is the issue that tracks the change to (B) mentioned above: #181
to be done "eventually"

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -108,6 +110,67 @@ class OpenLIFUProtocolConfigParameterNode:
...
+#
+
+class ProtocolSelectionFromDatabaseDialog(qt.QDialog):
+    """ Create new protocol selection from database dialog """
+
+    def __init__(self, protocol_names_and_IDs : "List[Tuple[str,str]]", parent=None):
+        super().__init__(slicer.util.mainWindow() if parent == "mainWindow" else parent)
+        """ Args:
+                dataLogic: instance of the OpenLIFUData module logic, needed to access the loaded database
+        """
```
ebrahimebrahim, line 124: update this docstring

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -108,6 +110,67 @@ class OpenLIFUProtocolConfigParameterNode:
...
+
+        self.setWindowTitle("Select a Protocol")
+        self.setWindowModality(qt.Qt.WindowModal)
+        self.resize(600, 400)
+
+        self.protocol_names_and_IDs : "List[Tuple[str,str]]" = protocol_names_and_IDs
```
ebrahimebrahim, line 130:
```suggestion
        self.protocol_names_and_IDs : List[Tuple[str,str]] = protocol_names_and_IDs
```

(we generally don't need quotes on type annotation, we only need those for the lazy importing stuff)

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -14,10 +15,11 @@
 from slicer.util import VTKObservationMixin
 from slicer.parameterNodeWrapper import parameterNodeWrapper
 
-# TODO: somehow you need to observe the data logic so you can connect the button here to it.
+# TODO: somehow you need to observe the OpenLIFUData module logic so you can connect the button here to it.
```
ebrahimebrahim, line 18: remove this TODO comment?

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
ebrahimebrahim: > Do you think I should add conditional logic to change the displayed protocol only if it is not being edited at the time the protocol is changed?

Yes, I think so. After testing it out again it does feel a bit rough to have to select the protocol after loading it. As long as the `cur_save_state` isn't `UNSAVED_CHANGES`, that's how I think it ought to work.

Maybe after you add the caching it can even work this way when `cur_save_state` is `UNSAVED_CHANGES`, allowing you to remove that branching logic -- we can decide that when you are doing the caching PR

(The earlier discussion that you linked doesn't seem to come to any particular resolution on this question of whether to select a protocol automatically after loading it, unless I missed something. Feel free to argue about it)

ebrahimebrahim (changes requested):
Few minor things and one major thing: automatically select a protocol after it's loaded using the OpenLIFUProtocolConfig widgets

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
arhowe00: > Feel free to argue about it
No arguments needed. The proof of the pudding, as it were, will be in the eating, or should I say the next review request.

ebrahimebrahim (approved):
It works really well! I like all the changes you made since the last one.

Also I hadn't fully noticed this, but it looks like caching (#169) is already in place? A few notes about the commit messages:

- If there's a particular commit that introduced the caching then you can add a reference to #169 in it. It's okay for a commit to reference multiple issues if it touched upon multiple issues, I think 
- If #169 will actually be closed by this PR then can you indicate that? I don't see what's left of #169 -- the caching seems to work.
- Can you make sure the commit messages are written in imperative mood? And clarify the commit message "Updated protocol config module" which is a bit too generic.

No code changes requested, but just the above commit message changes if they apply. Then I left approval so you can go ahead and hit the rebase-and-merge button when you have made those changes

arhowe00:
Sounds good, rewording and going to request review again.  In regards to your point about the commit that introduced caching, that was a commit that was already merged (913b9e54a2033af483276a0e4469e5da49bac065).  I did not reference the caching issue because I added a form of caching in response to an inconsistent behavior in the application that was only fixed by this addition, but that was in response to a comment about the issue I was addressing. As you said, I think there still is room for #169 , such as a "revert changes" and an official notification that there are unsaved WIP protocols. However I wouldn't be able to reword anything in this PR to address #169

ebrahimebrahim:
All sounds good, TY for fixing and merging!

Issue #149, opened 2024-12-04
Implement end-to-end transducer tracking


ebrahimebrahim:
## Goal

Make the "Run transducer tracking" button run transducer tracking.

There are actually several steps involved in transducer tracking and we will later want to break down the workflow into a refined step-by-step process, but for now in this issue we just want to get the blind end-to-end workflow working smoothly.

We will also remove the temporary controls that were created for the initial "mimimal" transducer tracking workflow.

## Details

After running transducer tracking, the algorithm gives us two transforms as output: one mapping the photoscan into MRI space, and one mapping the transducer registration surface into photoscan space.

Whether in treatment/session mode or in manual mode:
- these transforms should become slicer transform nodes
- the photoscan textured model is loaded into the slicer scene
- the transforms should get applied so that
  - the photoscan is transformed to the MRI space
  - the transducer is also transformed to the MRI space. So the current transducer transform node gets updated with the appropriate composition of the two matrices output by the algorithm, but we also save a separate transform node to serve as a snapshot of this matrix (Similar principle to what we do in #145 for virtual fitting).

In treatment/session mode:
- the result of the transducer tracking is saved into the session as well: the session stores a list of “transducer tracking results”, each of which is a photoscan ID and two transforms, and we just use the underlying openlifu functionality introduced in https://github.com/OpenwaterHealth/OpenLIFU-python/issues/182 to append a transducer tracking result

### Skin mesh from mri

Part of the transducer tracking is to take the raw volume input and generate a skin mesh. We will use openlifu to do this once  https://github.com/OpenwaterHealth/OpenLIFU-python/issues/149 is done. This will replace the temporary measure #152 

### Initialization of transducer transform

Update: We just found out that the transducer tracking algorithm needs a transducer transform to use as initialization to help it do the registration between transducer-registration-surface and photoscan.

In manual (no session) mode we can just use the transducer transform, so it is up to the user to move around the transducer to a good initialization. 
In treatment (session) mode we can use as candidates all the virtual fit candidates that are in the session, from https://github.com/OpenwaterHealth/OpenLIFU-python/issues/179.

## Things yet to clarify

A few things are unclear to me, and maybe we can think about them as we get into this item.

How should the new transform nodes be handled? Are they "openlifu objects" to be listed in the list of openlifu objects? Or are they affiliated objects of other things (like how the "current transducer transform node" is affiliated to the SlicerOpenLIFUTransducer)?

[...]

PR #180, opened 2025-02-14
Add functionality to preview photoscan

### LINKED ISSUE #148: Enable 3D view of photoscan before it's registered
In the transducer tracking module, there is an input dropdown menu for selecting an input photoscan. Add a button near there called "Preview photoscan". This should show 3D view of the photoscan separate from the ordinary slicer 3D view. It can't be in the same slicer 3D view because before registration there is no correspondence between the photoscan coordinates and the RAS+mm space. So it might be a pop-out vtk render window?

The button would allow for preview of the photoscan that is currently selected in the photoscan input dropdown menu.

**Instead of the popout vtk-render window, the photoscan will be displayed in the Slicer scene usinng the approach below**

There should also be a checkbox somewhere that controls approval of the photoscan itself (not approval of transducer tracking outcome, but just approval of the photoscan as being valid to use in transducer tracking)

Depends on #143 

If #146 is done then we can also automatically trigger this preview of a photoscan as soon as it is added to a session.

sadhana-r:
Closes #148 
Partly addresses the manual step of putting facial landmarks on the photoscans by introducing the markups widget within the preview dialog (#149)

In this PR, when the user clicks 'Place Registration Landmarks' a markups node is added to the scene and associated with the markups widget in the preview dialog. Further work is needed to add the created markups node to the photoscan object, ensure that any previously created markups node shows up in the dialog etc. This will be handled as part of a separate PR addressing https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/149. The current PR focuses on introducing the preview dialog and toggling the approval status of a photoscan during the session-based workflow. 

- [x] Rebase to main after merging https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/159
- [x] Disable the photoscan preview button when there are no valid inputs selected.

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1939,24 +1939,33 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
 
         # If the user selects a json file,use the photoscan_metadata included in the json file to load the photoscan. 
         elif Path(model_or_json_filepath).suffix == '.json':
-            return self.load_photoscan_from_json(model_or_json_filepath)
-
+            photoscan_openlifu = openlifu_lz().photoscan.Photoscan.from_file(model_or_json_filepath)
+            return self.load_photoscan_from_openlifu(photoscan_openlifu, parent_dir = Path(model_or_json_filepath).parent)
         else:
             slicer.util.errorDisplay("Invalid photoscan filetype specified")
 
-    def load_photoscan_from_json(self, json_filepath: str, replace_confirmed : bool = False) -> SlicerOpenLIFUPhotoscan:
+    def load_photoscan_from_openlifu(self, photoscan_openlifu, load_from_active_session: bool = False, parent_dir: str = None, replace_confirmed : bool = False) -> SlicerOpenLIFUPhotoscan:
         """Load an openlifu photoscan object into the scene as a SlicerOpenLIFUPhotoscan,
         adding it to the list of loaded openlifu objects.
 
         Args:
-            json_filepath: json file containing openlifu photoscan metadata
+            photoscan_openlifu: openlifu photoscan object
+            load_from_active_session (bool): If True, the photoscan is loaded based on the active session, from the openlifu database.
+            parent_dir (str): If load_from_active_session is False,then the absolute path to the parent directory containing the 
+                photoscan object and affiliated model and texture files needs to be specified. 
             replace_confirmed: Whether we can bypass the prompt to re-load an already loaded Photoscan.
                 This could be used for example if we already know the user is okay with re-loading the photoscan.
 
         Returns: The newly loaded SlicerOpenLIFUPhotoscan.
         """
 
-        photoscan_openlifu = openlifu_lz().photoscan.Photoscan.from_file(json_filepath)
+        # Check for valid inputs
+        if not load_from_active_session:
+            if parent_dir is None:
+                raise RuntimeError("Cannot load photoscan because the parent_dir was not specified for the photoscan object.")
+        else:
+            if self.db is None: # This shouldn't happen
+                raise RuntimeError("Cannot load photoscan because there is a session but no database connection to load the data from.")
 
         if photoscan_openlifu.id in self.getParameterNode().loaded_photoscans:
             if not replace_confirmed:
@@ -1968,9 +1977,11 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
             self.remove_photoscan(photoscan_openlifu.id) 
 
         with BusyCursor():
-            model_data, texture_data = openlifu_lz().photoscan.load_data_from_photoscan(
-                photoscan_openlifu,
-                parent_dir = Path(json_filepath).parent)
+            if load_from_active_session:
+                loaded_session = self.getParameterNode().loaded_session
+                _, (model_data, texture_data) = self.db.load_photoscan(loaded_session.get_subject_id(),loaded_session.get_session_id(),photoscan_openlifu.id, load_data = True)
+            else:
+                model_data, texture_data = openlifu_lz().photoscan.load_data_from_photoscan(photoscan_openlifu,parent_dir = parent_dir)
 
         newly_loaded_photoscan = SlicerOpenLIFUPhotoscan.initialize_from_openlifu_photoscan(
             photoscan_openlifu,
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1229,6 +1229,9 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
             ):
                 self.clear_solution(clean_up_scene=True)
             clear_virtual_fit_results(session_id = loaded_session.get_session_id(), target_id=None)
+            for photoscan_id in loaded_session.get_affiliated_photoscan_ids():
+                if photoscan_id in self.getParameterNode().loaded_photoscans:
+                    self.remove_photoscan(photoscan_id)
 
     def save_session(self) -> None:
         """Save the current session to the openlifu database.
@@ -1955,11 +1958,11 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
         # If the user selects a json file,use the photoscan_metadata included in the json file to load the photoscan. 
         elif Path(model_or_json_filepath).suffix == '.json':
             photoscan_openlifu = openlifu_lz().photoscan.Photoscan.from_file(model_or_json_filepath)
-            return self.load_photoscan_from_openlifu(photoscan_openlifu, parent_dir = Path(model_or_json_filepath).parent)
+            return self.load_photoscan_from_openlifu(photoscan_openlifu, parent_dir = str(Path(model_or_json_filepath).parent))
         else:
             slicer.util.errorDisplay("Invalid photoscan filetype specified")
 
-    def load_photoscan_from_openlifu(self, photoscan_openlifu, load_from_active_session: bool = False, parent_dir: str = None, replace_confirmed : bool = False) -> SlicerOpenLIFUPhotoscan:
+    def load_photoscan_from_openlifu(self, photoscan_openlifu, load_from_active_session: bool = False, parent_dir: Optional[str] = None, replace_confirmed : bool = False) -> SlicerOpenLIFUPhotoscan:
         """Load an openlifu photoscan object into the scene as a SlicerOpenLIFUPhotoscan,
         adding it to the list of loaded openlifu objects.
 
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1476,6 +1476,17 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
                 # Remove the volume already in the scene
                 idx = loaded_volume_ids.index(session_openlifu.volume_id)
                 slicer.mrmlScene.RemoveNode(loaded_volumes[idx])
+            
+        session_affiliated_photoscans = self.db.get_photoscan_ids(subject_id, session_id)
+        conflicting_loaded_photoscans =  [photoscan for photoscan in session_affiliated_photoscans if photoscan in self.getParameterNode().loaded_photoscans]
+        if (
+            conflicting_loaded_photoscans
+        ):
+            if not slicer.util.confirmYesNoDisplay(
+                f"Loading this session will replace the already loaded photoscan(s) with ID(s) {conflicting_loaded_photoscans}. Proceed?",
+                "Confirm replace photoscan"
+            ):
+                return
 
         # === Proceed with loading session ===
 
@@ -1543,6 +1554,10 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
         # === Set the newly created session as the currently active session ===
 
         self.getParameterNode().loaded_session = new_session
+
+        # === Keep track of affiliated photoscans and unload any conflicting photoscans that have been previously loaded ===
+        for photoscan_id in conflicting_loaded_photoscans:
+            self.remove_photoscan(photoscan_id) 
         self.update_photoscans_affiliated_with_loaded_session()
 
     def _on_transducer_transform_modified(self, transducer: SlicerOpenLIFUTransducer) -> None:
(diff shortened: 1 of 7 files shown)


ebrahimebrahim:
@arhowe00 I will also test and review, but leaving to you the deeper code-level review

sadhana-r:
> * if I try to preview a photoscan again after previewing it once, I get prompted about replacement -- maybe instead it should only try to load photoscan to preview it in the first place if it's not already loaded, otherwise use the already loaded one.

Thinking about this more, and there is the sitation where a user loads a photoscan manually ( so it exists in `loaded_protocols`) and then loads a session and previews the same photoscan, but now within the session workflow. If I don't prompt for replacement at the time of preview, the session will just use the photoscan that was loaded as part of the manual workflow, along with any marksups etc that were done then. 
For transducers and volumes, we prompt at the time of loading a sesion if a  manually loaded transducer/volume will get replaced. I could do that for photoscans as well? Even though the photoscansare not technically loaded at the time of loading a session, I can unload any existing `loaded_photoscans` that conflict with photoscan ids affiliated with the session. Then I don't need to re-load everyhing a photoscan is previewed. Thoughts?

ebrahimebrahim:
> Even though the photoscansare not technically loaded at the time of loading a session, I can unload any existing `loaded_photoscans` that conflict with photoscan ids affiliated with the session. Then I don't need to re-load everyhing a photoscan is previewed.

This makes sense to me, so what you're suggesting is:

- When clicking "preview photoscan" while in session-mode, it first checks if the photoscan is already loaded into the scene and only loads it if not, like I had suggested.
- (When clicking "preview photoscan" while not in session-mode, the photoscan is always already loaded.)
- When loading a session, if there is a photoscan ID already loaded in that conflicts one of the photoscan IDs affiliated with the session, then the user is asked (in [this section of code](https://github.com/OpenwaterHealth/SlicerOpenLIFU/blob/55731517fc382071ea1dc7a7406e36ad6e72473a/OpenLIFUData/OpenLIFUData.py#L1408)) if they are cool with some of their photoscans getting unloaded, and only proceed if so.

OpenLIFUData/OpenLIFUData.py
arhowe00, line 1: It appears that loading the photoscanPreviewDialog breaks unloading sessions. Here was the behavior I experienced:

1. Load a Database
2. Load demo_4_simulated session
3. Open TT
4. Select a photoscan
5. Hit 'Preview Photoscan'
6. Exit the dialog
7. Unload the session from the Data module

The photoscan stays selected in OpenLIFU TT. When I try to reload the session demo_4_simulated, I get the following window:

image

This behavior may not be a problem, but was just pointing it out.

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -204,6 +212,182 @@ def updateInputOptions(self):
...
+            viewNode.SetAxisLabelsVisible(False) # Turn off axis labels visibility
+
+        # Create widget
+        viewWidget = slicer.qMRMLThreeDWidget()
+        viewWidget.setMRMLScene(slicer.mrmlScene)
+        viewWidget.setMRMLViewNode(viewNode)
+    
+        return viewWidget
+
+    def _display_photoscan_in_widget(self, photoscan: SlicerOpenLIFUPhotoscan, viewWidget: qMRMLThreeDWidget) -> None:
```
arhowe00, line 317: For this function, it would be helpful to know more about why some bits of  code are there, as opposed to what the code does. For instance, from how `views_mainwindow` is set, I can mostly parse what it does, but I'm not sure what the purpose of it is. Like all things, inference can take me pretty far, but because this function makes heavy use of MRML that could be accelerated with a short description. I'll give a few small examples below.

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -204,6 +212,182 @@ def updateInputOptions(self):
...
+
+        photoscan_view_node = viewWidget.mrmlViewNode()
+
+        # IDs of all the view nodes in the main Window. This excludes the photoscan widget's view node
+        views_mainwindow = [node.GetID() for node in slicer.util.getNodesByClass('vtkMRMLViewNode') if node.GetName() != photoscan_view_node.GetName()]
```
arhowe00, line 322:
```suggestion
        views_mainwindow = [node.GetID() for node in slicer.util.getNodesByClass('vtkMRMLViewNode') if node.GetID() != photoscan_view_node.GetID()]
```

May be a rare occurrence, but aren't node IDs guaranteed to be unique, whereas names _can_ (usually by negligence) be duplicated?

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -204,6 +212,182 @@ def updateInputOptions(self):
...
+        
+        # Set the view nodes for all displayable nodes.
```
arhowe00, line 324: Ok, so the display node of the volume rendering module wants to be visible, so we will set the views to all the view nodes in the main window, except the photoscan view node.

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -204,6 +212,182 @@ def updateInputOptions(self):
...
+        for displayable_node in list(slicer.util.getNodesByClass('vtkMRMLDisplayableNode')):
+            if displayable_node.IsA('vtkMRMLScalarVolumeNode'):
+                # Check for any volume renderings
+                vrDisplayNode = slicer.modules.volumerendering.logic().GetFirstVolumeRenderingDisplayNode(displayable_node)
+                if vrDisplayNode and vrDisplayNode.GetVisibility():
+                        vrDisplayNode.SetViewNodeIDs(views_mainwindow)
+            elif displayable_node.GetDisplayVisibility():
+                displayable_node.GetDisplayNode().SetViewNodeIDs(views_mainwindow)
+        
+        # Set the view nodes for the Red, Green and Yellow slice nodes
```
arhowe00, line 334: But why do we need to set all the view nodes from the main window for each slice node to display the photoscan in the widget?

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -204,6 +212,182 @@ def updateInputOptions(self):
...
+                slice_node.AddThreeDViewID(view_nodeID)
+
+        # Display the photoscan (TODO: and fiducials if previously placed)
+        photoscan.toggle_model_display(True, photoscan_view_node) # Specify a view node for display
+
+        # Center and fit displayed photoscan in 3D view
+        layoutManager = slicer.app.layoutManager()
+        for threeDViewIndex in range(layoutManager.threeDViewCount):
+            view = layoutManager.threeDWidget(threeDViewIndex).threeDView()
+            if view.mrmlViewNode().GetName() == photoscan_view_node.GetName():
```
arhowe00, line 346:
```diff
-            if view.mrmlViewNode().GetName() == photoscan_view_node.GetName():
+            if view.mrmlViewNode().GetID() == photoscan_view_node.GetID():
```

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -204,6 +212,182 @@ def updateInputOptions(self):
...
+        threeDView = threeDWidget.threeDView() 
+        threeDView.rotateToViewAxis(3)  # look from anterior direction
+        threeDView.resetFocalPoint()  # reset the 3D view cube size and center it
+        threeDView.resetCamera()  # reset camera zoom
+
+    def updatePhotoscanApproveButton(self, photoscan_is_approved: bool):
+
+        loaded_session = get_openlifu_data_parameter_node().loaded_session
+        if photoscan_is_approved:
+            self.photoscanPreviewDialogUI.photoscanApprovalButton.setText("Unapprove photoscan")
```
arhowe00, line 359: Very pedantic, but I think "unapprove**d**" is a word, not "unapprove". Maybe we could use "disapprove" here, but it wouldn't follow the precedent of using "unapprove", already set elsewhere in the file.

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -291,6 +480,21 @@ def toggleTransducerTrackingApproval(self) -> None:
...
         session.toggle_transducer_tracking_approval() # apply the approval or lack thereof
         data_parameter_node.loaded_session = session # remember to write the updated session object into the parameter node
+    
+    def togglePhotoscanApproval(self, photoscan: SlicerOpenLIFUPhotoscan) -> None:
+        """Approve the specified photoscan if it was not approved. Revoke approval if it was approved. Write changes
+        to the underlying openlifu photoscan object to the database. """
+        data_parameter_node = get_openlifu_data_parameter_node()
+        session = data_parameter_node.loaded_session
+        if session is None: # We should never be calling togglePhotoscanApproval if there's no active session.
+            raise RuntimeError("Cannot toggle photoscan approval because there is no active session.")
```
arhowe00, line 490:
```diff
             raise RuntimeError("Cannot toggle photoscan approval because there is no active session.")
         photoscan.toggle_approval()
         # Write changes to the database
-        self._loaded_db = slicer.util.getModuleLogic('OpenLIFUData').db
-        if self._loaded_db is None: # This shouldn't happen
+        loaded_db = slicer.util.getModuleLogic('OpenLIFUData').db
+        if loaded_db is None: # This shouldn't happen
             raise RuntimeError("Cannot toggle photoscan approval because there is a session but no database connection to write the approval.")
-        OnConflictOpts : "openlifu.db.database.OnConflictOpts" = openlifu_lz().db.database.OnConflictOpts
-        self._loaded_db.write_photoscan(session.get_subject_id(), session.get_session_id(), photoscan.photoscan.photoscan, on_conflict=OnConflictOpts.OVERWRITE)
+        loaded_db.write_photoscan(session.get_subject_id(), session.get_session_id(), photoscan.photoscan.photoscan, on_conflict=openlifu_lz().db.database.OnConflictOpts.OVERWRITE)
```
If you grab the loaded_db every time, does it matter to associate it with the instance?  Maybe you have this for later use...

arhowe00 (changes requested):
The functionality works pretty well, I mostly only have styling suggestions.  There was one edge case I tested which didn't seem to behave as expected. Have a look and let me know what you think

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -291,6 +480,21 @@ def toggleTransducerTrackingApproval(self) -> None:
...
+        photoscan.toggle_approval()
+        # Write changes to the database
+        self._loaded_db = slicer.util.getModuleLogic('OpenLIFUData').db
+        if self._loaded_db is None: # This shouldn't happen
+            raise RuntimeError("Cannot toggle photoscan approval because there is a session but no database connection to write the approval.")
+        OnConflictOpts : "openlifu.db.database.OnConflictOpts" = openlifu_lz().db.database.OnConflictOpts
+        self._loaded_db.write_photoscan(session.get_subject_id(), session.get_session_id(), photoscan.photoscan.photoscan, on_conflict=OnConflictOpts.OVERWRITE)
```
arhowe00, line 497:
```diff
             raise RuntimeError("Cannot toggle photoscan approval because there is no active session.")
         photoscan.toggle_approval()
         # Write changes to the database
-        self._loaded_db = slicer.util.getModuleLogic('OpenLIFUData').db
-        if self._loaded_db is None: # This shouldn't happen
+        loaded_db = slicer.util.getModuleLogic('OpenLIFUData').db
+        if loaded_db is None: # This shouldn't happen
             raise RuntimeError("Cannot toggle photoscan approval because there is a session but no database connection to write the approval.")
-        OnConflictOpts : "openlifu.db.database.OnConflictOpts" = openlifu_lz().db.database.OnConflictOpts
-        self._loaded_db.write_photoscan(session.get_subject_id(), session.get_session_id(), photoscan.photoscan.photoscan, on_conflict=OnConflictOpts.OVERWRITE)
+        loaded_db.write_photoscan(session.get_subject_id(), session.get_session_id(), photoscan.photoscan.photoscan, on_conflict=openlifu_lz().db.database.OnConflictOpts.OVERWRITE)
```
If you grab the loaded_db every time, does it matter to associate it with the instance?  Maybe you have this for later use...

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
ebrahimebrahim: or "Revoke photoscan approval" to match with [the language](https://github.com/OpenwaterHealth/SlicerOpenLIFU/blob/9778e923b4c328bcc3142c485aaac19705dc2320/OpenLIFUPrePlanning/OpenLIFUPrePlanning.py#L421) coming in my virtual fit PR #179

sadhana-r: Sounds good. I was following the language used for the transducer tracking approval but I can update that to "Revoke transducer tracking approval" as well when I work on the transducer tracking PR next.

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
sadhana-r: good point!

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
sadhana-r: You're right. I have added docstrings to explain why this is needed. I also now deal with the edge cases where the user has already created a scene with specific view nodes specified.

arhowe00:
Hey @sadhana-r , reviewing this at the moment. Do you have anything in particular you think I should check out?

arhowe00 (approved):
Looks good! I didn't see the same errors that I saw before, and the photos can preview works well. I also like the comment you added, which helps maintianability for MRML-specific stuff.

Go ahead and rebase and merge.  Note that if you have any rebase conflicts in the UI, it likely has to do with making sure any new buttons go inside of a permissions widget.  In case this might help, that would mean making sure your button goes right here inside the permissionsWidget. 
image
If *all you added* was that button to the UI, when rebasing this conflict you could simply re-add the new button to what's currently the permissions widget on main (in QDesigner).  If you prefer fixing the merge conflict in the file, this case might be easy enough so that you can move your XML directly to the location, but I've had annoying moments when fixing merge conflicts on UI elements through incoming diffs.

PR #182, opened 2025-02-14
Create simple login module

### LINKED ISSUE #172: Create login module
# Description

Add a Login module.  The login module has a `Username/Password` field that can be filled out; if a matching `User` is found in `get_all_users()`, the application transfers to the Home module with the `user_id` and `roles` set as parameters in the Home module.  Note: If there is no `User` with the `root` role in `get_all_users()`, the user cannot type in a `Username/Password`, and the role is set to `root`.    
 
## How other modules would ask about the login state

There should also be a convenience function in OpenLIFULib somewhere that all other modules can use to ask the login module what are the roles of the currently logged in user. Here is how it can work:

- If user account mode is off, then this would return “root” as the role. If user acct mode is on then we follow the following logic:  
  - If no one is logged in, then the roles returned by this function would be either
    - root, if there is no User with the root role in `get_all_users()`, or
    - some very low-permissions role that we might call  ‘anonymous’, if there exists a root user already.

Then other modules won't have to mess too much with the internals of the login module -- they can get all the info they need out of this convenience function.

UPDATE: Check the description of https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/175 to get a sense for the names of the actual user roles we want to have.

## How other modules would be notified when the login state is changed

Not sure what's the best way yet, but if the login state is in the login module's parameter node then other modules can observe that node for changes. Or we can create our own little observer thing. Or maybe ask Sam about this. In any case this problem should be addressed here in this issue.

## Persistence

Ensure reasonable behavior when a slicer scene is saved and loaded. Whether user account mode is on should be preserved, but whether a user is logged in should not – you should have to re-log-in after loading.

# Dependencies

Depends on https://github.com/OpenwaterHealth/OpenLIFU-python/issues/203
Depends on #174

arhowe00:
Closes #172 

Some decisions are still to be made, such as whether "Login" should be a button, or whether the fields should be directly in the widget.

If you want to test logging in, the credentials are given thus (`username/password`):
```
complex_user/complex_user_hash
example_admin/example_admin_hash
example_operator/example_operator_hash
example_restricted/example_restricted_hash
```

Note that you will have to update the database to the version in OpenLIFU-Python [PR#215](https://github.com/OpenwaterHealth/OpenLIFU-python/pull/215), as passwords are checked against the salted hashes, not the password text.

### DIFF
--- a/OpenLIFULib/OpenLIFULib/Resources/python-requirements.txt
+++ b/OpenLIFULib/OpenLIFULib/Resources/python-requirements.txt
@@ -1,1 +1,2 @@
 git+https://github.com/OpenwaterHealth/OpenLIFU-python.git@0db543a787fc61b59958f8689987a9308c6fe183
+bcrypt
--- a/OpenLIFULib/OpenLIFULib/user_account_mode_util.py
+++ b/OpenLIFULib/OpenLIFULib/user_account_mode_util.py
@@ -2,6 +2,7 @@ from typing import Optional, TYPE_CHECKING
 
 from OpenLIFULib.util import (
         get_openlifu_login_parameter_node,
+        get_openlifu_login_logic,
         )
 
 if TYPE_CHECKING:
@@ -11,7 +12,7 @@ def get_current_user() -> "Optional[openlifu.db.User]":
     """Get the active openlifu user. If no user is logged in or user account
     mode is off, a default user is returned, with the intention of being the most
     restricted"""
-    return get_openlifu_login_parameter_node().active_user.user
+    return get_openlifu_login_logic().active_user.user
 
 def get_user_account_mode_state() -> bool:
     """Get user account mode state from the OpenLIFU Login module's parameter node"""
--- a/OpenLIFULib/OpenLIFULib/user_account_mode_util.py
+++ b/OpenLIFULib/OpenLIFULib/user_account_mode_util.py
@@ -1,6 +1,15 @@
+from typing import Optional, TYPE_CHECKING
+
 from OpenLIFULib.util import (
         get_openlifu_login_parameter_node,
         )
+
+if TYPE_CHECKING:
+    import openlifu.db
+
+def get_current_user() -> "Optional[openlifu.db.User]":
+    """Get the active openlifu user"""
+    return get_openlifu_login_parameter_node().active_user.user
 
 def get_user_account_mode_state() -> bool:
     """Get user account mode state from the OpenLIFU Login module's parameter node"""
--- a/OpenLIFULib/OpenLIFULib/user_account_mode_util.py
+++ b/OpenLIFULib/OpenLIFULib/user_account_mode_util.py
@@ -8,7 +8,9 @@ if TYPE_CHECKING:
     import openlifu.db
 
 def get_current_user() -> "Optional[openlifu.db.User]":
-    """Get the active openlifu user"""
+    """Get the active openlifu user. If no user is logged in or user account
+    mode is off, a default user is returned, with the intention of being the most
+    restricted"""
     return get_openlifu_login_parameter_node().active_user.user
 
 def get_user_account_mode_state() -> bool:
--- a/OpenLIFULogin/OpenLIFULogin.py
+++ b/OpenLIFULogin/OpenLIFULogin.py
@@ -149,6 +149,13 @@ class OpenLIFULoginWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         self._cur_login_state = LoginState.NOT_LOGGED_IN
         self._parameterNode = None
         self._parameterNodeGuiTag = None
+        self._default_anonymous_user = SlicerOpenLIFUUser(openlifu_lz().db.User(
+                id = "anonymous", 
+                password_hash = "",
+                roles = [],
+                name = "Anonymous",
+                description = "This is the default role set when the app opens, without anyone logged in, and when user account mode is deactivated. It has no roles, and therefore is the most restricted."
+        ))
 
     def setup(self) -> None:
         """Called when the user opens the module the first time and the widget is initialized."""
@@ -189,14 +196,7 @@ class OpenLIFULoginWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         self.updateUserAccountModeButton()
         self.updateLoginButton()
         self.updateWidgetLoginState(LoginState.NOT_LOGGED_IN)
-        default_anonymous_user = SlicerOpenLIFUUser(openlifu_lz().db.User(
-                id = "anonymous", 
-                password_hash = "",
-                roles = [],
-                name = "Anonymous",
-                description = "This is the default anonymous role automatically assigned when the app opens, before anyone is logged in. It has no roles, and therefore is the most restricted."
-                ))
-        self._parameterNode.active_user = default_anonymous_user
+        self._parameterNode.active_user = self._default_anonymous_user
 
     def cleanup(self) -> None:
         """Called when the application closes and the module widget is destroyed."""
@@ -258,7 +258,7 @@ class OpenLIFULoginWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         if new_user_account_mode_state:
             self.logic.start_user_account_mode()
         else:
-            self._parameterNode.active_user = None
+            self._parameterNode.active_user = self._default_anonymous_user
             self.updateWidgetLoginState(LoginState.NOT_LOGGED_IN)
             set_user_account_mode_state(new_user_account_mode_state)
 
@@ -287,7 +287,7 @@ class OpenLIFULoginWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
 
     @display_errors
     def onLogoutClicked(self, checked:bool):
-        self._parameterNode.active_user = None
+        self._parameterNode.active_user = self._default_anonymous_user
         self.updateWidgetLoginState(LoginState.NOT_LOGGED_IN)
 
     def updateLoginButton(self):
--- a/OpenLIFULogin/OpenLIFULogin.py
+++ b/OpenLIFULogin/OpenLIFULogin.py
@@ -22,6 +22,7 @@ from OpenLIFULib import (
     openlifu_lz,
     bcrypt_lz,
     SlicerOpenLIFUUser,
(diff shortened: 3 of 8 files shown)


OpenLIFULogin/OpenLIFULogin.py
```diff
@@ -0,0 +1,378 @@
...
+            self.ui.loginButton.setToolTip("The login feature is only available with user account mode turned on.")
+            return
+
+        if not get_openlifu_data_parameter_node().database_is_loaded:
+            self.ui.loginButton.setEnabled(False)
+            self.ui.loginButton.setToolTip("The login feature requires a database connection.")
+            return
+
+        # Now we see if there is an admin in the database
+        users = self.logic.dataLogic.db.load_all_users()
```
arhowe00, line 300: Things work fine when there are no users in the database, but if removing the `db_dvc/users/` directory entirely, it throws instead of returning an empty list:
```
  File "/Users/andrew.howe/projects/ow/SlicerOpenLIFU-Debug/lib/Slicer-5.7/qt-scripted-modules/OpenLIFULogin.py", line 261, in onUserAccountModeClicked
    self.updateLoginButton()
  File "/Users/andrew.howe/projects/ow/SlicerOpenLIFU-Debug/lib/Slicer-5.7/qt-scripted-modules/OpenLIFULogin.py", line 300, in updateLoginButton
    users = self.logic.dataLogic.db.load_all_users()
  File "/Users/andrew.howe/projects/ow/Slicer-Debug/python-install/lib/python3.9/site-packages/openlifu/db/database.py", line 819, in load_all_users
    raise FileNotFoundError("Users file not found.")
FileNotFoundError: Users file not found.
```
Should we edit OpenLIFU-Python to return an empty list in this case instead of throwing, or should we assume that a connected database is at least empty for all data types?  This question would apply to anything reading from the database.

ebrahimebrahim: A missing users directory could be though of as an improperly initialized database
But whether that's the way to think about it is a matter of opinion...
if you agree then I think it's fine that this throws

arhowe00:
Rebased to latest main and requesting review.

ebrahimebrahim:
sorry this is late, just letting you know that reviewing this is the next thing on my todo list :smile:

OpenLIFULib/OpenLIFULib/user_account_mode_util.py
```diff
@@ -1,7 +1,16 @@
...
 from OpenLIFULib.util import (
         get_openlifu_login_parameter_node,
         )
 
+if TYPE_CHECKING:
+    import openlifu.db
+
+def get_current_user() -> "Optional[openlifu.db.User]":
+    """Get the active openlifu user"""
+    return get_openlifu_login_parameter_node().active_user.user
```
ebrahimebrahim, line 12: This currently errors out if there is no `active_user` (i.e. if user account mode is off). It would be more useful for it to return `None` or something informative like that, and the docstring can explain under exactly which circumstances `None` would be returned and under what circumstances a default anonymous User gets returned.

(this function is sort of like a "front end" of the login module internals for other parts of the application to refer to in the future, that's why the docstring is more important)

OpenLIFULogin/OpenLIFULogin.py
```diff
@@ -54,13 +64,73 @@ def __init__(self, parent):
...
+    def customexec_(self):
+
+        returncode = self.exec_()
+        if returncode == qt.QDialog.Accepted:
+            id = self.username.text
+            password_text = self.password.text
+            # TODO: We should hash passwords with bcrypt.hashpw() and the gensalt, but then use checkpw against the stored hash
+            # TODO: This will be implemented with the ability to create new passwords, issue #173
+            # salt = bcrypt_lz().gensalt()
+            # password_hash = bcrypt_lz().hashpw(password_text, salt).decode('utf-8')  # convert bytestring back to string
```
ebrahimebrahim, line 131: looking forward to see how this works later :grin:

OpenLIFULogin/OpenLIFULogin.py
```diff
@@ -54,13 +64,73 @@ def __init__(self, parent):
...
 #
 # OpenLIFULoginParameterNode
 #
 
 @parameterNodeWrapper
 class OpenLIFULoginParameterNode:
     user_account_mode : bool
+    # TODO: We should only serialize the user if the user selects some button
+    # like "remember me." Perhaps, we could unconditionally serialize the
+    # user_id, for ease of repeated login
```
ebrahimebrahim, line 82: I say let's keep it simple and never do a "remember me" functionality. 

Unconditionally keeping the user ID makes sense! You can remove this to-do and either do that here or open new issue to put it off. The last used user ID can be kept in the QSettings. See how the OpenLIFUData module remember what to fill in for the database path for an example of that.

ebrahimebrahim (changes requested):
- If I try to login when I am already logged in, and I provide wrong credentials and fail, then the `loginStateNotificationLabel` indicates my failure, but I appear to still be logged in.
	- It's not good that `loginStateNotificationLabel` shows the wrong state, so maybe there is some logic there to clean up, but besides that it is also a bit weird that I can attempt to re-login when I am already logged in like that. Maybe that can be polished a bit. For example if I click login when I am already logged in, a YesNoDialog could perhaps inform me that I am logged and ask whether I want to log out? Then if yes, I am logged *out* and then faced with the login prompt.
- If I log in, save the slicer scene, and then load the scene later, then when I check the loaded scene via `get_current_user` I see that the current user is still there logged in! 
	- That shouldn't happen -- after loading a scene that a logged in person had saved, you should have to re-authenticate right?
	- The logout button is enabled, as it ought to be when someone is logged in, but the `loginStateNotificationLabel` is blank. Those two things, the button enabledness states and the `loginStateNotificationLabel` text, shouldn't be out of sync like that.
- The general occurance of buttons and labels not being synced up and updated to show the true login state could perhaps be resolved by calling all the updates you need to call in an "`onParameterNodeChanged`" for the login module parameter node. That's handy because then everything just updates whenever the parameter node's `User` changes.

I reviewed all code except the big one: OpenLIFULogin.py. Since the next updates may make a few changes in there, I'll do a more thoroguh review of that next time.

Thanks!!

arhowe00:
> * If I try to login when I am already logged in, and I provide wrong credentials and fail, then the `loginStateNotificationLabel` indicates my failure, but I appear to still be logged in.
>   
>   * It's not good that `loginStateNotificationLabel` shows the wrong state, so maybe there is some logic there to clean up, but besides that it is also a bit weird that I can attempt to re-login when I am already logged in like that. Maybe that can be polished a bit. For example if I click login when I am already logged in, a YesNoDialog could perhaps inform me that I am logged and ask whether I want to log out? Then if yes, I am logged _out_ and then faced with the login prompt.

I think it may be better just to have the login/logout buttons switch their function. I intended to make it so that you could _either_ log out of the account, or while logged in, _try_ to login to a different account. I could make it so that the name of the logged in user stays there, but we are also notified of the failed attempt, but it may make more sense just to assume we are only either logging in while logged out and logging out while logged in.
 
> * If I log in, save the slicer scene, and then load the scene later, then when I check the loaded scene via `get_current_user` I see that the current user is still there logged in!

It it possible (or even sensible) to have the user associated with the parameter node, but not to serialize it in the scene?  I think it's possible as I've done something like this in the past -- but should I do it?

>   
>   * That shouldn't happen -- after loading a scene that a logged in person had saved, you should have to re-authenticate right?
>   * The logout button is enabled, as it ought to be when someone is logged in, but the `loginStateNotificationLabel` is blank. Those two things, the button enabledness states and the `loginStateNotificationLabel` text, shouldn't be out of sync like that.
> * The general occurance of buttons and labels not being synced up and updated to show the true login state could perhaps be resolved by calling all the updates you need to call in an "`onParameterNodeChanged`" for the login module parameter node. That's handy because then everything just updates whenever the parameter node's `User` changes.

Good idea.  I think this will jive well with what I suggested above?

OpenLIFULogin/OpenLIFULogin.py
arhowe00: Ok so -- together with below -- don't serialize anything, but save the previous user id in the QSettings?

ebrahimebrahim:
> It it possible (or even sensible) to have the user associated with the parameter node, but not to serialize it in the scene? I think it's possible as I've done something like this in the past -- but should I do it?

If you figure out a way to do that then it sounds good to me. If not then another approach is to just do an auto-logout upon load? Is there a way you can observe a mrml scene load event? Could be a question for Sam. Another approach is to refactor a bit and not use the parameter node to store the `User` object. Then you lose the convenient observation mechanism built into parameter nodes and would have to create you own little observation mechanism, but it's not too bad. It is somehting you can do with the @property concept we discussed not too long ago. ([Here](https://github.com/OpenwaterHealth/SlicerOpenLIFU/blob/abd484ee2de4cbe0babf752ab6e019217c642f57/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py#L382-L386) is an example where do that for the "is running" variable in sonication control)

arhowe00:
Okay, in the imminent review request, I've moved the `active_user` from the Login parameter node to the Login logic. This will be easy to integrate with all the intended behavior of the module, and because we always know exactly when a user is logging in from the Login module, updates to the enabledness of widgets will be easily done in the currently envisioned way, as it's all just done from the Login module (but that isn't part of this PR)

ebrahimebrahim (approved):
Looks great, thank you for all this!

I tested it out, everything worked smoothly. The logic added to OpenLIFULogin makes sense. I see it's not based on an observer pattern, but rather just remembering to call the right widget updaters whenever it makes sense to do so. Since it's a small module without so many moving parts, I think his approach makes sense.

OpenLIFULib/OpenLIFULib/Resources/python-requirements.txt
```diff
@@ -1 +1,2 @@
-git+https://github.com/OpenwaterHealth/OpenLIFU-python.git@f9f94cf91e8dc18540a6afc2077aa88611cdc16f
+git+https://github.com/OpenwaterHealth/OpenLIFU-python.git@40ad62bb6d29e6df775aaabab565dea1c75bc0a1
```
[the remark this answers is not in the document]
ebrahimebrahim: actually let me just go head and rebase it :)

PR #185, opened 2025-02-19
Enforce user permissions

### LINKED ISSUE #175: Enforce user permissions
# Description

Enforce the user permissions by enabling/disabling widgets based on user role. The user role should always be queried via the convenience function introduced in #172.  

Note: The logic here needs to be thought through with care, because widgets can be disabled for other reasons besides login related ones. The devised solution should be something maintainable and not too confusing. The first thing that comes to mind is each widget that needs to care about login state should observe an updateEnabledness function that gets called every time the login state changes. Many widgets already observe a function like this (possibly under a different name) so we’d want to integrate the user permission check into the existing function. Can this be streamlined in some way to reduce code duplication? The names of the “updateEnabledness” functions should be standardized to make them easy to find should we need to ever add a new user role or something like that.  

# Breakdown of this issue by module

This issue will probably best handled by multiple PRs, with each PR focusing on a module or a few modules. Let's use this checklist here to keep track of which modules are done:

- [x] login
- [x] data (the SlicerOpenLIFU data module)
- [x] preplanning
- [x] transducer tracking
- [x] planning
- [x] control
- [x] protocol config
- [x] home

# Information on user roles and permissions

This section is information from Peter -- we can request additional details as we get into it.

| Function                       |  Engineering User  | Admin User | System Operator |
|--------------------------------|:------------------:|:----------:|:---------------:|
| Define Array                   |     Python/JSON    |    FALSE   |      FALSE      |
| Run Standalone Simulations     |    Python/Slicer   |    FALSE   |      FALSE      |
| Water Tank Operation           |    Python/Slicer   |    FALSE   |      FALSE      |
| Manage Users                   |    Python/Slicer   |  Admin GUI |      FALSE      |
| Configure and Publish Protocol | Python/Slicer/JSON |  Admin GUI |      FALSE      |
| Create Session / Add Data      | Python/Slicer/JSON |     GUI    |       GUI       |
| Set Target                     | Python/Slicer/JSON |     GUI    |       GUI       |
| Virtual Fitting                |    Python/Slicer   |     GUI    |       GUI       |
| Transducer Tracking            |    Python/Slicer   |     GUI    |       GUI       |
| Create Solution                | Python/Slicer/JSON |     GUI    |       GUI       |
| Run Sonication                 |    Python/Slicer   |     GUI    |       GUI       |

> Engineering Users are not actual application users, but instead interact with the lower level Python and Slicer Components (and in some instances, could use Matlab or anything else to generate the JSON files directly). They would need to have the same OS-level access as an admin user. 
>
> Admin Users have an expanded set of privileges through widgets in the application that are only available to them. These are primarily User and Protocol management, although we may come across other app functions that should be restricted to them.
>
> System Operators are trained personnel given access to key system functionality to plan and run sonications. They cannot change the Protocol(s) that they have access to, nor modify User permissions. For particularly sensitive applications or if the IRB requires it, Operators may be interacting with the app on an OS account that does not have write access to certain data.

# Dependencies

Depends on #172
Depends on Depends on https://github.com/OpenwaterHealth/OpenLIFU-python/issues/204

arhowe00:
Closes #175

---

## Summary

This PR introduces the general framework for enforcing user roles in the login module by enclosing buttons in permissions widgets. Adding restrictions to buttons and choosing the roles that can interact should be very straightforward.

When testing, try to login and ensure that certain abilities, e.g. editing protocols, can only be done by admins (you can see the full table in [#175](https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/175)). If it works for protocols, it should work for everything else.

Currently, new users cannot be created. When testing, you can login to example users by connecting the database and referring to the credentials in the [PR where the login module was created](https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/182).

We'll have to come up with a different way of notifying users they can't do stuff when not logged in than tooltips, but this is an issue for later. For instance, operators cannot edit protocols, but the tooltip on the disabled button is related to what tooltip would have been there regardless of permission (e.g. a protocol is not selected).

---

## Missing implementations and notes

Note that the following could not yet be implemented due to buttons not being defined at this time.  Despite this, what we have is a framework for enforcing permissions through QWidgets.

### Missing implementations:

> Function - Why it's not enforceable

Define Array - Not well defined.
Run Standalone Simulations - Not well defined.
Water Tank Operation - Not well defined.
Manage Users - Buttons not yet implemented.

### Further note:

For "Transducer Tracking", the permissions control the entire widget:
image

I assume that "Create Solution" means "Compute sonication solution" in the OpenLIFUSonicationPlanner module, and so that permission only applies to the buttons here:
image

I assume that "Run Sonication" means "Run" in the OpenLIFUSonicationControl module, and so that permission only applies to the button here:
image

### DIFF
--- a/OpenLIFUData/Resources/UI/OpenLIFUData.ui
+++ b/OpenLIFUData/Resources/UI/OpenLIFUData.ui
@@ -19,7 +19,7 @@
      <property name="collapsed">
       <bool>false</bool>
      </property>
-     <layout class="QVBoxLayout" name="verticalLayout_2">
+     <layout class="QVBoxLayout" name="verticalLayout_9">
       <item>
        <widget class="QLabel" name="databaseDirectoryLabel">
         <property name="text">
@@ -45,31 +45,55 @@
        </widget>
       </item>
       <item>
-       <widget class="QPushButton" name="newSubjectButton">
-        <property name="text">
-         <string>Add New Subject</string>
+       <widget class="QWidget" name="permissionsWidget2" native="true">
+        <property name="slicer.openlifu.allowed-roles" stdset="0">
+         <stringlist>
+          <string>admin</string>
+          <string>operator</string>
+         </stringlist>
         </property>
-       </widget>
-      </item>
-      <item>
-       <widget class="QPushButton" name="addVolumeToSubjectButton">
-        <property name="text">
-         <string>Add Volume to Subject</string>
-        </property>
-       </widget>
-      </item>
-      <item>
-       <widget class="QPushButton" name="newSessionButton">
-        <property name="text">
-         <string>Create New Session</string>
-        </property>
-       </widget>
-      </item>
-      <item>
-       <widget class="QPushButton" name="addPhotoscanToSessionButton">
-        <property name="text">
-         <string>Add Photoscan to Session</string>
-        </property>
+        <layout class="QVBoxLayout" name="verticalLayout_2">
+         <property name="leftMargin">
+          <number>0</number>
+         </property>
+         <property name="topMargin">
+          <number>0</number>
+         </property>
+         <property name="rightMargin">
+          <number>0</number>
+         </property>
+         <property name="bottomMargin">
+          <number>0</number>
+         </property>
+         <item>
+          <widget class="QPushButton" name="newSubjectButton">
+           <property name="text">
+            <string>Add New Subject</string>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <widget class="QPushButton" name="addVolumeToSubjectButton">
+           <property name="text">
+            <string>Add Volume to Subject</string>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <widget class="QPushButton" name="newSessionButton">
+           <property name="text">
+            <string>Create New Session</string>
+           </property>
+          </widget>
+         </item>
+         <item>
+          <widget class="QPushButton" name="addPhotoscanToSessionButton">
+           <property name="text">
+            <string>Add Photoscan to Session</string>
+           </property>
+          </widget>
+         </item>
+        </layout>
        </widget>
       </item>
       <item>
@@ -104,7 +128,7 @@
      <property name="collapsed">
       <bool>true</bool>
      </property>
-     <layout class="QVBoxLayout" name="verticalLayout_4">
+     <layout class="QVBoxLayout" name="verticalLayout_10">
       <item>
        <widget class="QScrollArea" name="scrollArea">
         <property name="sizePolicy">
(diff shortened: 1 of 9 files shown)


peterhollender:
For these three:
Define Array - Not well defined.
Run Standalone Simulations - Not well defined.
Water Tank Operation - Not well defined.

These are all activities that are done outside of the application context by directly accessing the source code.

Manage Users... I guess this would ultimately need need some sort of panel to modify the users table.

Otherwise, you've got it right. The primary difference is the ability to create and publish new protocols. "publishing" a protocol involves making it accessible to a specified set of other users. I had discussed with @ebrahimebrahim a while ago that when you publish a protocol, you probably should not be able to further change it, but would be required to publish a new version of it (e.g. `my_protocol_v2` rather than modifying an existing one). This may require a bit deeper discussion of how to handle this - I'm not sure if we want to introduce explicit versioning - I don't want to annoy research users, but we want to make sure that as a study is conducted, we are keeping track of who got what protocol, and reduce the likelihood that an operator runs a wrong or outdated protocol.

ebrahimebrahim:
Protocol versioning is a good idea and could be introduced in future protocol module work. Right now we actually have a good failsafe for knowing exactly what protocol was used on someone: [we save a snapshot of the protocol when we write the run](https://github.com/OpenwaterHealth/OpenLIFU-python/blob/1dcff787b74a3b301edbab2f52f6d52835bd0198/src/openlifu/db/database.py#L200-L210).

ebrahimebrahim:
@arhowe00 thanks for this! Can you advise on what I should check or look at in the review?

arhowe00:
Yeah, just login to a few accounts and make sure widgets are enabled and disabled accordingly.  How about these two:
• When not logged in, you shouldn't be able to to click anything in the purple boxes above.  How about you try it for OpenLIFUSonicationPlanner.
• When logged in as operator, you shouldn't be able to edit protocols but you should be able to load protocols from anywhere.
Lastly, check that things work intuitively. Just do a random walk over login, logouts, module switches, and deactivating user account mode, about 10 steps of intuitive behavior should be good, anything not, report it.

ebrahimebrahim (changes requested):
Putting the specific permission levels into a permissionWidgets layer like that is :pinched_fingers:! Really nice simple and extensible design

Important thing I noticed which is the main requested change: After entering user acct mode and messing around to see that everything works, once I _exited_ user acct mode it seems like the permissions stuff didn't update as a consequence.

Unimportant thing that might be nice to have: Could have more meaningful names on the permissions widgets rather than numbers like `permissionsWidget2`. So for example `permissionsWidget2` in `OpenLIFUData.ui` could be named `permissionsWidgetDatabaseWriting` or something along those lines, etc. But not that important

ebrahimebrahim (approved):
Looks good now! I didn't test it after your fix, but I assume you tested it. Please go ahead and rebase+merge if you are happy with it

Issue #140, opened 2024-11-18
Integrate with sonication control code


ebrahimebrahim:
# Background

Bit fuzzy still.

There's the [`DeviceInterface`](https://github.com/OpenwaterHealth/OpenLIFU-python/blob/e17497dc7a87aa31152e7b9e769c32d314096196/src/openlifu/io/ustx.py#L1073)  API created by Peter and then there's the [`LIFUInterface`](https://github.com/OpenwaterHealth/OpenLIFU-python/pull/163#issuecomment-2483277651) concept that George talked about.

This is part of a branch [p/sonication_control](https://github.com/OpenwaterHealth/OpenLIFU-python/tree/p/sonication_control)

We have a PR for it [here](https://github.com/OpenwaterHealth/OpenLIFU-python/pull/163) but it’s not clear what everything is in that branch

# This issue

- Figure out what’s going on with that p/sonication_control branch and figure out what API we should be interfacing with
- Once George has provided an API with the needed callbacks, make our SlicerOpenLIFU sonication control module interface with it

PR #191, opened 2025-03-03
Add button to sonication control module to send solution to hardware before initiating run

### LINKED ISSUE #156: Add button to sonication control module to send solution to hardware before initiating run
Before starting sonication, there should be a separate step which loads the solution on to the hardware and makes sure that there are no hardware related issues i.e. memory error. This should not be a part of the 'Run' function call.

arhowe00:
Closes #156 

This implements the button which sends the Data parameter `loaded_solution` to the LIFU device, using the `LIFUInterface` object. The `LIFUInterface` object is initialized using `test_mode=True`, and the state is checked before confirming that the send was successful. In one example, I printed the object and state after initialization, and it looked like this after everything in the `try` block executed (left out of the PR):

```
Interface:  <openlifu.io.LIFUInterface.LIFUInterface object at 0x263c11130>
Status:  LIFUInterfaceStatus.STATUS_READY
```

Note a few things. In this particular case, I assume that hardware-related issues will be propagated through the `LIFUInterface` through exception raising. This is why the implementation for the button (`def onSendSonicationSolutionToDevicePushButtonClicked`) encloses the LIFUInterface initialization and solution setting in a try block. If things work, through a simple state update, the widget will allow the run button to be pressed.  The most recently initialized LIFUInterface is stored in the logic.

Please gloss over the changes and note anything that looks like it doesn't make sense, or note any behaviors that were unexpected.

Please make sure to update the python libraries in the Home module before testing this branch.

### DIFF
--- a/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
+++ b/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
@@ -319,16 +319,19 @@ class OpenLIFUSonicationControlWidget(ScriptedLoadableModuleWidget, VTKObservati
     def onSendSonicationSolutionToDevicePushButtonClicked(self, checked=False):
 
         try:
-            interface = openlifu_lz().io.LIFUInterface.LIFUInterface(test_mode=True)
-            interface.txdevice.enum_tx7332_devices(num_devices=2)
+            interface = openlifu_lz().io.LIFUInterface(test_mode=True)
+            interface.txdevice.enum_tx7332_devices(2) # TODO: see why can't use kwarg
             interface.set_solution(get_openlifu_data_parameter_node().loaded_solution.solution.solution)
 
             self.logic._lifu_interface = interface
 
-            if interface.get_status() != openlifu_lz().io.LIFUInterface.STATUS_READY:
+            if interface.get_status() != openlifu_lz().io.LIFUInterfaceStatus.STATUS_READY:
                 raise RuntimeError("Interface not ready")
                 
-        except Exception:
+        except Exception as e:
+            print("Exception thrown:", e)
+            import traceback
+            traceback.print_exc()
             self.logic._lifu_interface = None
             self.updateWidgetSolutionHardwareState(SolutionHardwareState.FAILED_SEND)
             return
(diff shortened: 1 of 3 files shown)


ebrahimebrahim:
@sadhana-r giving you the sonication control reviews :smile:

arhowe00:
Ready for review. Fixed some things at the last moment which you probably would have caught early on.

OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
```diff
@@ -259,21 +278,44 @@ def setParameterNode(self, inputParameterNode: Optional[OpenLIFUSonicationContro
             # ui element that needs connection.
             self._parameterNodeGuiTag = self._parameterNode.connectGui(self.ui)
 
-    def onDataParameterNodeModified(self,caller, event) -> None:
+    def onDataParameterNodeModified(self, caller=None, event=None) -> None:
+        self.updateSendSonicationSolutionToDevicePushButtonEnabled()
         self.updateRunEnabled()
         self.updateRunProgressBar()
```
sadhana-r, line 284: You need to also add a check here in the event that a new solution is computed, replacing a previously sent solution or if a solution gets set to None. In that case, you would need to call `self.WidgetSolutionHardwareState(SolutionHardwareState.NOT_SENT)`.  For this I think you need to keep track of the current solution ID and update the widget if the solution ID changes and enable/disable buttons accordingly.

OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
```diff
@@ -292,7 +334,31 @@ def onRunCompleted(self, new_sonication_run_complete_state: bool):
...
+
+            if interface.get_status() != openlifu_lz().io.LIFUInterfaceStatus.STATUS_READY:
+                raise RuntimeError("Interface not ready")
+                
+        except Exception as e:
+            print("Exception thrown:", e)
+            import traceback
+            traceback.print_exc()
+            self.logic._lifu_interface = None
+            self.updateWidgetSolutionHardwareState(SolutionHardwareState.FAILED_SEND)
```
sadhana-r, line 355: Not sure if the LIFU interface returns any other states/information now, but it would be good to add information about the kind of exception/what triggered it in the WidgetSolutionHardwareState message.

sadhana-r (changes requested):
Looks great! Noted the edge case where a user re-computes a solution that I think can be addressed in this PR.

arhowe00:
Okay, in regards to printing the state in the case of a failed send, I added a TODO and will address it as part of #140 . This is because in addressing #140 , the LIFUInterface hardware state Enum is imported, whereas with the python requirements in this version, it's not.

sadhana-r (approved):
looks great! Thanks for making the changes

PR #193, opened 2025-03-04
Integrate Hardware Interface and Test Mode Toggle

### LINKED ISSUE #140: Integrate with sonication control code

arhowe00:
Closes #140

## Summary

This pull request removes the LIFUInterface placeholders (e.g. `run()`) and integrates the hardware interface into the application, alongside a developer-oriented toggle switch between a `test_mode` interface and the real device. 

## Key Changes

- **Hardware Interface Integration (`LIFUInterface`)**
  - Handles starting and stopping sonication (`start_sonication`, `stop_sonication`).
  - Device signals are parsed and update the GUI (e.g., run progress and connectedness).

- **Test Mode Switching**
  - When switching modes, the interface is reinitialized, and the event loop holding the asynchronous `start_monitoring` is reset. When in test mode, it uses a more informative mock of start_monitoring, but this can be replaced.

## Additional Notes

- The run-status parsing logic is still subject to change pending finalization from the hardware team.

---

## For review

This will need a few stages of review. Kitware's team should do an initial pass over the code (stage 1), and then the hardware team should do a pass over the code (stage 2), and then the hardware team should actually test a release of the app (stage 3)

### Stage 1 

First, the application-level code should be reviewed for anything obvious before sending to @georgevigelette . The GUI parts of this need trivial testing; the software-hardware-bridge parts matter the most.

1. Try computing and approving a sonication solution and sending it to the device, connected in test_mode, and then hitting run.
2. Look at the code where signals from the LIFUInterface are connected to slots in the OpenLIFUSonicationControlLogic. 
3. Look at how the event loop `self.monitoring_loop` is set up and switches out the `start_monitoring` task depending on whether LIFUInterface is initialized in `test_mode`. Note that the task needs to be switched out only if we are toggling on/off test_mode, which shouldn't happen in the app.
4. It may be wise to look at all of the things the hardware team would look at in Stage 2, but it's extra work.

### Stage 2

After revisions, if any, the code should be looked over by the hardware team.

1. Look at the code under `---- LIFU Interface Connection ----`.  In this part, an initial LIFUInterface is initialized, and `start_monitoring` is scheduled in the event loop.
2. Look at `OpenLIFUSonicationControlLogic.run(self)`. That's the part where `run_sonication` is called, data received signals are connected, and the device is polled for completion status.
3. Look at `OpenLIFUSonicationControlLogic.reinitialize_lifu_interface(self, test_mode=False)`. This part reinitializes the LIFUInterface in `test_mode`. While this won't be used by the customer, its correctness is very important for the rest of our development.
5. Check that the run progress signal was parsed correctly (in `OpenLIFUSonicationControlLogic.update_run_progress_from_lifuinterface(self, descriptor, message)`.

### Stage 3

After revisions by both teams, the hardware team should install a release of the OpenLIFU-app and attempt non-test_mode runs on real hardware and report any issues.

### DIFF
--- a/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
+++ b/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
@@ -390,10 +390,10 @@ class OpenLIFUSonicationControlWidget(ScriptedLoadableModuleWidget, VTKObservati
             import traceback
             traceback.print_exc()
             self.logic.cur_lifu_interface = None
-            self.ui.manuallyGetDeviceStatusLabel.setProperty("text", "Operation failed.")
+            slicer.util.infoDisplay(text=f"Operation failed.", windowTitle="Device Status")
             return
 
-        self.ui.manuallyGetDeviceStatusLabel.setProperty("text", f"{self.logic.cur_lifu_interface.get_status()}")
+        slicer.util.infoDisplay(text=f"{self.logic.cur_lifu_interface.get_status()}", windowTitle="Device Status")
 
     def onRunningChanged(self, new_running_state:bool):
         self.updateSendSonicationSolutionToDevicePushButtonEnabled()
--- a/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
+++ b/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
@@ -218,7 +218,7 @@ class OpenLIFUSonicationControlWidget(ScriptedLoadableModuleWidget, VTKObservati
         self.addObserver(slicer.mrmlScene, slicer.mrmlScene.EndCloseEvent, self.onSceneEndClose)
 
         # Buttons
-        self.ui.connectLIFUDevicePushButton.clicked.connect(self.onConnectLIFUDevicePushButtonClicked)
+        self.ui.reinitializeLIFUInterfacePushButton.clicked.connect(self.onReinitializeLIFUInterfacePushButtonClicked)
         self.ui.sendSonicationSolutionToDevicePushButton.clicked.connect(self.onSendSonicationSolutionToDevicePushButtonClicked)
         self.ui.runPushButton.clicked.connect(self.onRunClicked)
         self.ui.abortPushButton.clicked.connect(self.onAbortClicked)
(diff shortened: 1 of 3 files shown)


OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
```diff
@@ -330,6 +371,22 @@ def updateRunEnabled(self):
...
 
+    def updateAllButtonsEnabled(self):
+        self.updateReinitializeLIFUInterfacePushButtonEnabled()
+        self.updateManuallyGetDeviceStatusPushButtonEnabled()
+        self.updateSendSonicationSolutionToDevicePushButtonEnabled()
+        self.updateRunEnabled()
+        self.updateAbortEnabled()
+
+    def updateReinitializeLIFUInterfacePushButton(self):
+        if self.logic.cur_lifu_interface._test_mode:
```
ebrahimebrahim, line 382: The fact that the `_test_mode` attribute name starts with a `_` suggests that it's "private" in spirit and we shouldn't use it from the outside like this. That being said maybe it's just not officially exposed yet and we have to do this for now, I'm just pointing it out

OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
```diff
@@ -567,3 +789,48 @@ def create_openlifu_run(self, run_parameters: Dict) -> SlicerOpenLIFURun:
...
+        self.cur_lifu_interface.stop_monitoring()
+
+        # Cancel current task if it's running
+
+        if self.current_monitoring_task and not self.current_monitoring_task.done():
+            self.current_monitoring_task.cancel()
+            try:
+                self.monitoring_loop.run_until_complete(self.current_monitoring_task)  # Ensure it's fully canceled
+            except asyncio.CancelledError:
+                pass
```
ebrahimebrahim, line 808: Having to do this makes me feel like we are doing something wrong, but maybe just worth revisiting in the future after clarifying our understanding of asyncio. It's fine for now, I don't think tying together two event loops is going to be trivial in any world!

OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
```diff
@@ -567,3 +789,48 @@ def create_openlifu_run(self, run_parameters: Dict) -> SlicerOpenLIFURun:
...
+
+        self.cur_lifu_interface.signal_connect.connect(self.on_lifu_device_connected)
+        self.cur_lifu_interface.signal_disconnect.connect(self.on_lifu_device_disconnected)
+
+    def get_lifu_device_connected(self) -> bool:
+        tx_connected = self.cur_lifu_interface.txdevice.is_connected()
+        hv_connected = self.cur_lifu_interface.hvcontroller.is_connected()
+        return tx_connected and hv_connected
+
+    async def test_mode_start_monitoring(self, interval=1):
```
ebrahimebrahim, line 831: this was a great idea, to mock start_monitoring while in test mode!

ebrahimebrahim (commented):
Stage 1 review:

Clean work, very nicely done!

1. Everything seems to work well while test mode is enabled. I couldn't manage to break the UI despite my best efforts :)
2. The connections to the LIFUSignal attributes of LIFUInterface look good to me
3. The event loop creation and cancelling all looks right as far as I can tell

Ready for stage 2 I think!

OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
arhowe00: Yes -- There is no way of getting the test_mode state of the device, and for perfect synchronization I don't want to alias it with state in our module. Let's see how the interface evolves before modifying this part

ebrahimebrahim (approved):
Adding approval since my comments were addressed and I want to leave this PR in @arhowe00's hands to merge when ready. The plan is to merge tomorrow at some point, with any further outcomes of "stage 2" of testing to be handled in future work.

Issue #157, opened 2024-12-13
Make transducer tracking placeholder for interactive registration 


sadhana-r:
As a placeholder solution till the transducer tracking library is ready, the transducer tracking algorithm can output the expected transform nodes to scene (initialized with identity), that can then be edited by manually/interactively registering the MRI/transducer in Slicer.

ebrahimebrahim:
(For anyone later looking at this issue, the description above is outdated now. The transducer tracking library never came to be, so this issue became about creating a dummy workflow that represents the end-to-end manual transducer tracking process.))

Issue #203, opened 2025-03-11
Refine end-to-end workflow for saving transducer tracking results


sadhana-r:
This includes
- checking when transducer transforms are modified or removed from the scene.
- creating a workflow for approval/revoking approval of transducer tracking results
- When a session has a transducer tracking result approved, if approval of the corresponding photoscan is revoked, the user should be notified. What should happen to the tt result? This is somewhat equivalent to the situation when a target with a virtual fit result is modified/removed. 
- Can a user proceed with transducer tracking when the photoscan is unapproved?

ebrahimebrahim:
Superceded by #259

PR #204, opened 2025-03-11
Add complete transducer tracking UI workflow

### LINKED ISSUE #199: Create UI for end-to-end transducer tracking wizard workflow
The transducer tracking wizard consists of the following steps:

1. Display the unregistered photoscan for user to annotate 3 landmarks (R/L ears, nasion) 
2. Compute and display the skin segmentation for the user to annnotate 3 landmarks 
3. Display the photoscan and skin segmentation, with the photoscan position initialized by fiducial registration. There should be a transform node attached to the photoscan surface that the user can manually adjust.
4. Display the transducer registratation surface and registered photoscan from step 3. There should be a transform node attached to the transducer surface that the user can manually adjust. The transducer surface is initialized based on  VF result. 

After step 4, the user exits the workflow and the transforms are saved.

This PR focuses on implenting the UI aspects of the transducer tracking wizard. At each step, the appropriate view node with the required models should be displayed. All other displayable nodes must be hidden. The appropriate registration/fiducial marking controls should be visible. 
For now, the photoscan position does not need to include the fiducial registration result and the transducer dos not need to take into account the VF results.

### LINKED ISSUE #200: Add functionality to annotate registration landmarks on the skin segmentation surface
User should be able to add a fiducial node to the skin segmentation surface with 3 points, similar to the photoscan landmarks. 

This can also be modified later on to use the guided workflow as described in #196

sadhana-r:
Closes #199 
Closes #200 

Few things to address:

- [x] - Currently still need to specify a skin segmentation to proceed
- [x] - Able to click "next" in the TT wizard even without placing any landmarks
- [x] - Able to click "next" in the TT wizard while int he middle of placing landmarks, before clicking "Done placing landmarks"
- [x] To fix: If transducer folder is not displayed in the main scene i.e. eyeballs off, then transducer is not visualized in the wizard.
- [x] Remove transducer tracking view node from photoscan when the user exits wizard.  
- [x] When clearing a photocan, clear the associated view node from the scene
- [ ] Can I set when the back button is enabled/disabled? What behavior is associated with cleanupPage? #203 
- [ ] Which signal is used to indicate wizard finish vs cancel. #203 
- [x] When the session is cleared or the selected TT volume node is removed from the scene - the associated facial landmarks and skin segmentation should also be cleared from the scene.

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1375,11 +1375,12 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
             for photocollection_reference_number in loaded_session.get_affiliated_photocollection_reference_numbers():
                 if photocollection_reference_number in self.getParameterNode().session_photocollections:
                     self.remove_photocollection(photocollection_reference_number)
+            
             for photoscan_id in loaded_session.get_affiliated_photoscan_ids():
                 if photoscan_id in self.getParameterNode().loaded_photoscans:
                     self.remove_photoscan(photoscan_id)
 
-            clear_transducer_tracking_results
+            clear_transducer_tracking_results(session_id = loaded_session.get_session_id())
 
     def save_session(self) -> None:
         """Save the current session to the openlifu database.
@@ -1659,7 +1660,14 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
                 return
             
         session_affiliated_photoscans = self.db.get_photoscan_ids(subject_id, session_id)
-        conflicting_loaded_photoscans =  [photoscan for photoscan in session_affiliated_photoscans if photoscan in self.getParameterNode().loaded_photoscans]
+        if loaded_session is not None:
+            # (we are okay reloading photoscans if they are just the one affiliated with the session, since user already decided to replace the session)
+            loaded_session_affiliated_photoscans = self.db.get_photoscan_ids(loaded_session.get_subject_id(), loaded_session.get_session_id())
+            conflicting_loaded_photoscans =  [photoscan for photoscan in session_affiliated_photoscans 
+                                              if photoscan in self.getParameterNode().loaded_photoscans and 
+                                              photoscan not in loaded_session_affiliated_photoscans]
+        else:
+            conflicting_loaded_photoscans =  [photoscan for photoscan in session_affiliated_photoscans if photoscan in self.getParameterNode().loaded_photoscans]
         if (
             conflicting_loaded_photoscans
         ):
@@ -1668,7 +1676,7 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
                 "Confirm replace photoscan"
             ):
                 return
-
+        
         # === Proceed with loading session ===
 
         self.clear_session()
@@ -1746,8 +1754,6 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
         self.getParameterNode().loaded_session = new_session
 
         # === Keep track of affiliated photoscans and unload any conflicting photoscans that have been previously loaded ===
-        for photoscan_id in conflicting_loaded_photoscans:
-            self.remove_photoscan(photoscan_id) 
         self.update_photoscans_affiliated_with_loaded_session()
 
         # === Load photocollections as all reference_numbers ===
--- a/OpenLIFULib/OpenLIFULib/photoscan.py
+++ b/OpenLIFULib/OpenLIFULib/photoscan.py
@@ -128,14 +128,11 @@ class SlicerOpenLIFUPhotoscan:
     def toggle_approval(self) -> None:
         self.photoscan.photoscan.photoscan_approved = not self.photoscan.photoscan.photoscan_approved 
     
-    def toggle_model_display(self, visibility_on: bool = False, viewNodes: List[vtkMRMLViewNode] = []):
-        """ If a viewNode is not specified, the model is displayed in all views by default"""
+    def toggle_model_display(self, visibility_on: bool = False):
+
         self.model_node.GetDisplayNode().SetVisibility(visibility_on)
-        self.model_node.GetDisplayNode().SetViewNodeIDs([node.GetID() for node in viewNodes])
-        
         if self.tracking_fiducial_node:
             self.tracking_fiducial_node.GetDisplayNode().SetVisibility(visibility_on)
-            self.tracking_fiducial_node.GetDisplayNode().SetViewNodeIDs([node.GetID() for node in viewNodes])
                         
     def create_tracking_fiducial_node(self, right_ear_coordinates = [0,0,0], left_ear_coordinates = [0,0,0], nasion_coordinates = [0,0,0]):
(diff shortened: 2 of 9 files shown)


sadhana-r:
@ebrahimebrahim @arhowe00 This PR is ready for review. No need for a detailed code review for now. The main things to check:

- You shold be able to load a session and click through each step of the wizard. All the buttons should work. For now, the transform approve button just toggles the status message. It does not modify the approval attribute of the associated transform node. Clicking 'run ICP registration' should turn on and off the interaction handles. You shouldn't be able to click Next/Finish while you're in the middle of performing a certain task. 
- After going through the wizard, two transducer tracking result nodes are added to the scene with the required attributes. 
- Clearing the session/deleting a volume node also removes any nodes introduced due to the TT wizard.
- The 3D views on each page are set up with correct nodes being visible. 
- I have enforced the ClassicWizard style across all platforms. I'm curious to see what the wizard looks like on Mac/Linux.
- I have found that clicking `Back` sometimes causes weird issues with the 3D view. I am still trying to figure out how to disable that button in certain situations. I also need to figure out how to catch when the wizard is finished vs. cancelled. I tried a couple different things but they caused my Slicer to crash. I thought I would handle that when refining the end -to-end workflow. 
- Please note any edge cases/weird behaviour.

ebrahimebrahim (commented):
Really amazing work, so nice to see this complicated framework coming together smoothly!

- Can we use `with BusyCursor` for time from after "Run transducer tracking" is clicked until the wizard pops up? There is some time consuming computation that blocks slicer, and busy cursor helps the user know that this is supposed to happen.
- ~~The "Run ICP registration" button can be renamed to "Run ICP registration (Not implemented - toggles transducer transform handles)" or maybe a tooltip can be added to explain the situation~~
- The two TT transform nodes that are created are added to the scene, and I saw that when I save the session they also get written to the session. But I think it's not the correct transforms -- that's intentional right? If so, have we written into one of the future issues to handle that? There is a subtlety which is that `transform_node_to_openlifu` is meant to apply specifically to tranform nodes that map the transducer into the slicer RAS space. It won't work directly to get the photoscan->MRI or the transducer->photoscan transforms. For the photoscan->MRI transform you would want to just convert the matrix in the transform node to LPS (so postcompose the matrix by RAS/LPS swap) and put "mm" as the units. For the transducer->photoscan transform it's more complicated. My suggestion to mininize having to write new transform conversion functions, two options here: Option A: in the slicer scene instead of having a transducer->photoscan transform node, have a transducer->RAS space transform node. This is more natural to use in the slicer scene and will make in-scene things more easy to handle. Then you can use `transform_node_to_openlifu` to turn that into an openlifu ArrayTransform with the appropriate units for the transducer. But then when you go to write into the session, you do the matrix math that turns it into a transducer->photoscan matrix. When loading transducer tracking results, matrix math is also done to convert transforms to the in-scene format. Option B: Same as option A except instead of doing the matrix math we change openlifu-python representations to match in-scene represenations. So in TransducerTrackingResult we replace `transducer_to_photoscan_transform` by `transducer_to_volume_transform`. Sorry for the ramble there --  please let me know if you want to discuss. Not sure how much you had already planned out -- maybe this is an already solved problem on your end in which case feel free to ignore my suggestions
- The current transducer transform doesn't update at the end -- that's also intentional right? Same question, is one of the future issues slated to handle that? Check out the SlicerOpenLIFUTransducer method `set_current_transform_to_match_transform_node` which does this, dependening on your approach to the above bullet point.
- In places where you construct an openlifu `TransducerTrackingResult`, I recommend using keyword arguments rather than listing the arguments and relying on the order of the list. (I think this comment might be more related to your already closed https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/157 but you could address it here.) If someone re-orders the dataclass attributes in openlifu python it would be difficult to catch, say, the two transform arguments getting swapped.
- Closing the scene or deleting nodes: worked cleanly!
- Back issues: yes it does seem to cause issues as you noticed. It is okay to open separate bug issues for those also. I'll get a notification wehn you open those issues and then I can think about which milestone (deadline) they belong in.
- Determining finished vs canceled: Looking at the Qt docs, QWizard inherits from QDialog, so the thing you're looking for might be at the QDialog level and so that's why you're not finding it in the QWizard doc page. I think where you call `wizard.exec_()` this may return a [QDialog.DialogCode](https://doc.qt.io/qt-5/qdialog.html#DialogCode-enum)... not certain but the QDialog doc page looks like the place to look for this level of information
- Sessionless: I also tried running through the wizard without an active session. It worked perfectly, amazing!
- No memory leaks reported after I closed slicer. I know that can be difficult :D


ClassicWizard appearance on Ubuntu with gnome 3:

Peek 2025-03-19 14-16

ebrahimebrahim:
> The "Run ICP registration" button can be renamed to "Run ICP registration (Not implemented - toggles transducer transform handles)" or maybe a tooltip can be added to explain the situation

(I just re-read my comment and noticed this shouldn't be there -- you already have a big red message there. I think I wrote that before I saw the message and then forgot to remove it)

sadhana-r:
> * The two TT transform nodes that are created are added to the scene, and I saw that when I save the session they also get written to the session. But I think it's not the correct transforms -- that's intentional right? 
> * The current transducer transform doesn't update at the end -- that's also intentional right? Same question, is one of the future issues slated to handle that? Check out the SlicerOpenLIFUTransducer method

Thanks for this insight! I have not implemented all of this transform logic yet. I'm working on that now as part of initializing the nodes using virtual fit/fiducial registration.  I think Option B (copied below) makes sense. In the end, we mainly care about the `transducer_to_volume` transform so I don't think we need to add in the extra matrix math steps of computing the `transducer_to_photoscan` transform just for the sake of writing it to the sesssion as per current openlifu representations. And to confirm my understanding, when you say `transducer->RAS space transform node`, that is the same as the `transducer_to_volume` node right? i.e the volume is in the reference Slicer RAS space? And if I clone the virtual fit result to get that intial transform, that is already a transducer to RAS space transformation? 

> * Option B: Same as option A except instead of doing the matrix math we change openlifu-python representations to match in-scene represenations. So in TransducerTrackingResult we replace `transducer_to_photoscan_transform` by `transducer_to_volume_transform`

And following this approach, at the end, I can call `set_current_transform_to_match_transform_node` on the resulting `transducer_to_volume` node.

ebrahimebrahim:
> And to confirm my understanding, when you say `transducer->RAS space transform node`, that is the same as the `transducer_to_volume` node right? i.e the volume is in the reference Slicer RAS space? And if I clone the virtual fit result to get that intial transform, that is already a transducer to RAS space transformation?

Yes exactly

sadhana-r:
@ebrahimebrahim I'm addressing the creation/saving of transform nodes in #213. Can this baseline framework be merged first or should I abandon this PR and merge once I have completed #213?

ebrahimebrahim:
> @ebrahimebrahim I'm addressing the creation/saving of transform nodes in #213. Can this baseline framework be merged first or should I abandon this PR and merge once I have completed #213?

Either way works, totally up to you. Adding approval so you can decide whether to merge now

ebrahimebrahim (approved):

PR #205, opened 2025-03-11
Fix photocollections not being removed on session unload

arhowe00:
This introduces a minor fix that removes loaded photocollections when `remove_photocollection` is called


ebrahimebrahim (approved):
I checked `slicer.modules.OpenLIFUDataWidget.logic.getParameterNode().loaded_photocollections` as I switched between sessions and indeed it loaded and unloaded as expected with `example_session`

(unrelated to this PR, but it's kinda weird to call that variable `loaded_photocollections`, since they are not loaded into memory really, it's rather a list of photocollection reference numbers that are affiliated with the session right? IDK maybe I am missing something there)

arhowe00:
That's a good point and I think I'll make that change here

PR #218, opened 2025-03-24
Fix bug in login module and reserve user ids 

### LINKED ISSUE #217: Login module needs to reserve special user_ids
Creating an account with the is "anonymous" and "default_admin" should not be allowed because they are placeholder names used internally for application logic. In the rare case that a reserved id is chosen, reject those inputs.

arhowe00:
Closes #217 

Fixes a minor bug in the login module with user mode being toggled on/off, as well as reserves user IDs.

I tested everything and the changes are atomic enough that a code glance should suffice.


ebrahimebrahim (approved):
Looked it over and it all makes sense, thanks!

PR #221, opened 2025-03-29
127 set up guided mode framework and UI

ebrahimebrahim:
This adds a guided mode framework.

The way it works is that each module that needs guided mode controls has a placeholder widget called `workflowControlsPlaceholder`. Each such module gets a mixin `GuidedWorkflowMixin` added to its inheritance chain. This provides it with a method `inject_workflow_controls_into_placeholder` that sticks the widget into place.

The workflow widgets look like this:

image

with some variability when different buttons need to be available at the start and end.

Memory-wise the workflow widgets are owned by the module widgets that they live in, but bookkeeping-wise they are tracked in a `Workflow` object that is in the OpenLIFUHome module logic. This object does the setup, linking the next and back buttons to one another and allowing you to define how the workflow is organized in one place.

The `GuidedWorkflowMixin` grants to each module widget an attribute `workflow_controls` which is the the widget in the screenshot above. Each module is responsible for controlling the widget's `can_proceed:bool` and `status_text:str` attributes. With the magic of properties, the widget updates automatically when those attributes are modified.

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -44,6 +44,7 @@ from OpenLIFULib.util import (
     ensure_list,
     add_slicer_log_handler_for_openlifu_object,
     BusyCursor,
+    replace_widget,
 )
 
 from OpenLIFULib.virtual_fit_results import (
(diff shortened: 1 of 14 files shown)


arhowe00 (approved):
Wasn't sure exactly what to look at for review, but we did look at this together and how it works makes a lot of sense. I checked out the branch and built and poked around. It seems to work great as the skeleton for the new guided mode workflow.

ebrahimebrahim:
Thanks for looking through it! I had intended for our looking over it together to cover most of the review, so I was just looking for a final sign-off after that, so this is good!

Issue #165, opened 2025-01-28
Add Additional Protocol Parameter Controls


arhowe00:
## Description

Determine which additional protocol parameter controls should be exposed and add them to the UI.

### References

- Depends on #164

ebrahimebrahim:
Adding to this issue the following:

- Write units on the protocol control module fields that represent physical quantities
- Put tooltips explaining what the fields are. A good way to do this would be to steal the docstrings from the `Protocol` class itself, [once they exist](https://github.com/OpenwaterHealth/OpenLIFU-python/issues/201).
- Enforce range limits on the numerical protocol fields (e.g. via suitable QValidators)

@peterhollender please inform about reasonable range limits

ebrahimebrahim:
Yes I think the idea is to show the units in the GUI.
So for example instead of this saying "Pulse frequency"
https://github.com/OpenwaterHealth/SlicerOpenLIFU/blob/54f857afce5d67b93937255a8b100697b5207fd6/OpenLIFUProtocolConfig/Resources/UI/OpenLIFUProtocolConfig.ui#L119
it would say "Pulse frequency (Hz)", which we know because of the docstrings and/or comments we can find in openlifu:
https://github.com/OpenwaterHealth/OpenLIFU-python/blob/322c75a3ea7b3eeba93aeb4bea7ab092412cf4a6/src/openlifu/bf/pulse.py#L12-L23

ebrahimebrahim:
> * Enforce range limits on the numerical protocol fields (e.g. via suitable QValidators)

Yes It needs some input from Peter I think, but once you have everything in place and all the QValidators in there then you can just quickly ask what would be reasonable numbers to use. These are not safety limits, I think they are more just general reasonableness limits.

peterhollender:
Hmm... enforcing of "reasonable ranges" here is actually a bit tricky. The requirement is frequency 150-500kHz (easy enough), Mechanical Index 0-1.9 (which would involve checking the center frequency and the target pressure), Pulse length is 0-100ms (easy enough) max duty cycle is 50% (which means checking pulse length and pulse interval in the sequence) and max total duration is 15 minutes (pulse interval * number of pulses * number of pulse trains * pulse train interval)...

ebrahimebrahim:
[...]

PR #225, opened 2025-04-02
Add additional protocol parameter controls

arhowe00:
Contributes to #165 

Reworked the protocol config UI to use abstract form widgets based on Python type hints, replacing many specific forms with dynamic ones that support class introspection.  Currently, the latest commit relies on `python-requirements.txt` that are unmerged, and this PR does assume that Annotated[] dataclass fields will supply both variable names and descriptions (for tooltips).

Some of the classes here are generically useful and should be noted for other parts of the app (I'm sure someone has already made these libraries before...)

- A widget that creates a form based on the type annotations of an arbitrary Python class. Supports all primitive field types including tuple and nested tuples, dicts, and lists. 
- A widget that creates a collection of forms based on classes that inherit the same ABC, which lets you select the class and then populate the form for that class 
- A widget that allows interaction with a Dictionary of `str` key and an arbitrary value (can be an entire object, lets you fill in based on form
- A widget that allows interaction with a List of arbitrary values (same as dict but without the key)


arhowe00:
Okay @ebrahimebrahim It's ready for review since the python-requirements this uses is merged

ebrahimebrahim (approved):
Really incredible work here. This is a monster PR, and yet sparkling clean. The abstract form creation classes are well documented, so we can revisit and re-use these wonderfully generic components.

There's far too much to test for me to test everything, but the overall design is good, and the few tests that I ran worked smoothly. We'll deal with any more specific issues as they get discovered.

As I played with it and looked through some tooltips, I decided I should retract [my earlier complaint](https://github.com/OpenwaterHealth/OpenLIFU-python/pull/264#pullrequestreview-2737107577) about duplication of text between docstrings and annotation fields. They really do serve different roles, and we should want to be able to make them different. Some things make sense to say in a class docstring that wouldn't make sense in a more general user-facing context. I think that OpenLIFUFieldData's documentation should make it clear that the description text inside it is meant to be end-user-facing text and not API/developer level text.

ebrahimebrahim:
Also should you set this PR so that it closes the issue #165 ?

arhowe00:
> Also should you set this PR so that it closes the issue #165 ?

It doesn't add the constraints

arhowe00:
Note that TargetConstraints is not serializable, and so even though you can add them inside the protocol config, you cannot save them to database because it relies on openlifu.
image
image

ebrahimebrahim:
> > Also should you set this PR so that it closes the issue #165 ?
> 
> It doesn't add the constraints

Ah, let me split out the constraints from #165 so that it becomes a separate issue.

PR #226, opened 2025-04-02
Add mesh reconstruction

ebrahimebrahim:
Makes it so that when you press "start photoscan generation" it calls the underlying function in openlifu that Daniel added. The infrastructure was already put together by Andrew, we just had to replace the placeholder by the function call here.

todo:

- [ ] update python reqs to something on main once https://github.com/OpenwaterHealth/OpenLIFU-python/pull/270 is merged
- [ ] make it so that the enabledness of the preview photoscan button and such updates


arhowe00 (approved):
Looks good, I have a possible alternative for generating unique photoscan_id s from the reference_number s . However, it's optional.

PR #228, opened 2025-04-03
Make targets renamable

ebrahimebrahim:
This makes it possible to set a target name by clicking on it

Peek 2025-04-02 20-43

A challenge here was to update the target dropdowns in three different modules: preplanning, transducer tracking, and sonication planning. There is no specific rename event that is invoked by fiducial nodes, so we'd have to watch for arbitrary PointModified events. But PointModified events are continuously invoked as points get dragged and dropped, and updating the input widgets across several modules continuously is ridiculous.

So the solution here was to invoke a custom vtk event when targets are renamed via SlicerOpenLIFU. This actually seemed generally useful for customizing inter-module communication in SlicerOpenLIFU, so I put the event definition into a new OpenLIFULib.events. It could come in handy!

Things to check for the review:
- The updates to the dropdowns in the transducer tracking module are done appropriately
- Optionally, you can build it and try it out. Rename a target, or create a new target and rename it, and then see that
  - the name updates in on the node and in the display
  - the name updates in the targets dropdown in the preplanning module, the TT module, and the sonication planning module

### DIFF
--- a/OpenLIFUPrePlanning/OpenLIFUPrePlanning.py
+++ b/OpenLIFUPrePlanning/OpenLIFUPrePlanning.py
@@ -36,6 +36,7 @@ from OpenLIFULib.virtual_fit_results import (
 from OpenLIFULib.targets import fiducial_to_openlifu_point_id
 from OpenLIFULib.coordinate_system_utils import get_IJK2RAS
 from OpenLIFULib.transform_conversion import transducer_transform_node_from_openlifu
+from OpenLIFULib.events import SlicerOpenLIFUEvents
 
 if TYPE_CHECKING:
     from OpenLIFUData.OpenLIFUData import OpenLIFUDataLogic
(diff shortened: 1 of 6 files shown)


OpenLIFUPrePlanning/OpenLIFUPrePlanning.py
```diff
@@ -319,6 +323,15 @@ def onTargetListWidgetCurrentItemChanged(self, current:qt.QListWidgetItem, previ
         self.updateTargetPositionInputs()
         self.updateLockButtonIcon()
 
+    def onTargetListWidgetItemDataChanged(self, item:qt.QListWidgetItem):
+        node : vtkMRMLMarkupsFiducialNode = item.data(qt.Qt.UserRole)
+        node.SetName(item.text().replace(" ", "-")) # This becomes openlifu Point ID
+        node.SetNthControlPointLabel(0, item.text()) # This becomes openlifu Point name
+        node.InvokeEvent(SlicerOpenLIFUEvents.TARGET_NAME_MODIFIED_EVENT)
```
ebrahimebrahim, line 330: Here is the _invocation_ of the custom event. In Qt terminology it's _emitting_ a signal

sadhana-r (changes requested):
Very cool solution! Nice to have that utility introduced here. I tested it out in a session-based and manual workflow and the drop downs update smoothly!
One thing that is missing is that the 'Loaded OpenLIFU Objects' table in the data module needs to be updated as well. Currently, the renaming event is not triggering that update. 
I also noticed that when I place a new target, then it updates that loaded objects table, so previously renamed targets get updated in the table. However, the new target I added i.e. 'Target_1' is incorrectly reflected in the loaded objects table as 'MarkupFiducial_X'

ebrahimebrahim:
Addressed that issue with the loaded objects view not upating:

Peek 2025-04-03 13-38

Ready for another review

sadhana-r (approved):

Issue #181, opened 2025-02-14
Handle missing db in session-based workflow


ebrahimebrahim:
... handle the errors. That is, allow for `db` to be `None` while there is a `loaded_session` -- the application won't work until you do connect to database, but it also won't break or crash on you. So allow for the dangling-session-with-missing-database-connect situation that arises when you save and load a slicer scene in the middle of a session based workflow. And whenever something in the session based workflow needs to use `db` but finds that it is `None`, the user is informed in an error/warning that they need to reconnect to the database to make things work properly. So for example when the photoscans dropdown is trying to populate, it sees that there is a session but no database, so it raises this error dialog (not a python exception just a friendly error dialog) and remains an empty dropdown.

_Originally posted by @ebrahimebrahim in https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/178#discussion_r1955049710_

Issue #128, opened 2024-11-03
Implement module-specific guided mode behaviors


ebrahimebrahim:
## Implement `can_proceed` and status message for each individual module.

Go into each module and make guided_mode_status_message and can_proceed take on the right values at all times (e.g. in the data module we are checking if there’s a loaded subject, in the pre-planning module if there’s an approved virtual fit, etc.)

## Hide and unhide

Make appropriate components of each module get hidden when guided mode is enabled and unhidden when it’s disabled to disallow things during guided mode.

## Progress on this issue

This could be done in several different PRs that all reference this issue. To help track the progress of this, let's check off modules as we finished them:

|                                       | `can_proceed` and `status_text` | hiding things in GM |
|---------------------------------------|:-------------------------------:|:-----------------:|
| database step   (needs to be split out)        |                X                 |     X              |
| login                                 |                  X               |         X          |
| data (the SlicerOpenLIFU data module) |                X                 |         X          |
| preplanning                           |                 X                |    X               |
| transducer tracking                   |                 X                |       X            |
| planning                              |                 X                |       X            |
| control                               |                 X                |          X         |

## Dependency

Depends on https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/127

PR #230, opened 2025-04-07
Make OpenLIFUDatabase module and decouple from OpenLIFUData

### LINKED ISSUE #229: Move database into separate module and add utilities for getting/setting location
We want to introduce a Database module in the Slicer extension that defines the notion of a database and includes convenience functions such as determining the default database location if one is not set. This will make the concept of initializing a default database in the custom app standalone from determining where it should be.

Interacting with the OpenLIFUDatabase module will also be the first step in the guided workflow, assuming one is not yet connected, making the guided mode more coherent by requiring a data context from the start (essential before any other steps are made). The functionality should reside in the SlicerOpenLIFU repository--not the custom app--to support platform-specific logic and database management.

**The OpenLIFU-app should then be updated to reference a merged commit of SlicerOpenLIFU that includes the new Database module.**

arhowe00:
# Closes #229 

This PR focuses on decoupling the OpenLIFUData module from direct database access and logic, moving all related functionality into the new OpenLIFUDatabase module.

- All database connection, initialization, and loading logic has been moved to OpenLIFUDatabase.
- Introduced a global accessor (`get_cur_db()`) to access the database where needed.
- Adds utilities for creating new databases in filesystem paths provided by the user 
- Prevents non-admins from setting database paths
- Adds a default database as part of the OpenLIFUDatabase module resources
- Most subject/session population logic now pulls from OpenLIFUDatabase, despite tight coupling in previous behavior.

This will simplify solving other issues significantly. For instance, [OpenLIFU-app#14](https://github.com/OpenwaterHealth/OpenLIFU-app/issues/14) is already mostly done with this PR. The Guided mode issue #128 would now have a database module to start at.

## Notes

Certain lines from OpenLIFUData were preserved to keep the UI behavior consistent with how it was previously implemented, even though they now operate in OpenLIFUDatabase. I did not want to modify parts of the Data module unrelated to the database, and so for instance, I had to keep the following lines which were in the previous implementation (database actions were tied to session/subject actions). They should be removed later:

```python
@display_errors
def onLoadDatabaseClicked(self, checked:bool):
    subject_info = self.logic.load_database(Path(self.ui.databaseDirectoryLineEdit.currentPath))
    slicer.util.getModuleWidget('OpenLIFUData').updateSubjectSessionSelector(subject_info)  # <--- looks like bad practice
    self.updateSettingFromParameter('databaseDirectory')
    slicer.util.getModuleWidget('OpenLIFUData').update_newSubjectButton_enabled()  # <--- looks like bad practice
```

```python
def load_database(self, path: Path) -> Sequence[Tuple[str,str]]:
    slicer.util.getModuleLogic('OpenLIFUData').clear_session()  # <--- looks like bad practice
    subjects = {}
    # ... (database initialization and loading)
    slicer.util.getModuleLogic('OpenLIFUData')._subjects = subjects  # <--- looks like bad practice
```

These remain to prevent regressions and UI inconsistencies but do not rely on direct access to the database internals anymore.

## For Review

This actually ended up touching a lot of the codebase, especially in regions I do not maintain. I tried not to destroy anything. For review, please run the full pipeline of widgets you authored. For instance, in my case I loaded the DB, logged in, created and edited protocols, and tried doing some sonication stuff. To speed up the review, just make sure the fullest pipeline you can think of executes with no new issues.

# Contributes to #181 

- Removes `db_is_loaded` from the data module parameter node
- Where the variable is checked, instead checks whether OpenLIFUDatabaseLogic's `db` attribute is `None`
- Introduces a property setter and callback mechanism for modules to know when the database loaded state is changed. It was implemented for OpenLIFUProtocolConfig and OpenLIFULogin
- Does not fully accomplish the following item in #181:
  > where the data module parameter node is observed for changes to detect when db_is_loaded changes, instead make sure those situations just look at db and can give a nice error if db is None without crashing the application or raising an exception -- so the user knows they just need to "reconnect" to the database to get things working again


OpenLIFUDatabase/OpenLIFUDatabase.py
```diff
@@ -0,0 +1,363 @@
...
+    """This class should implement all the actual
+    computation done by your module.  The interface
+    should be such that other python code can import
+    this class and make use of the functionality without
+    requiring an instance of the Widget.
+    Uses ScriptedLoadableModuleLogic base class, available at:
+    https://github.com/Slicer/Slicer/blob/main/Base/Python/slicer/ScriptedLoadableModule.py
+    """
+
+    db = None
```
ebrahimebrahim, line 230: Curious, why is it a class level attribute? No objection I am just curious

OpenLIFUDatabase/OpenLIFUDatabase.py
```diff
@@ -0,0 +1,363 @@
...
+    @property
+    def database_is_loaded(self) -> bool:
+        """Whether database_is_loaded"""
+        return self._database_is_loaded
+
+    @database_is_loaded.setter
+    def database_is_loaded(self, database_is_loaded_value : bool):
+        self._database_is_loaded = database_is_loaded_value
+        for f in self._on_database_is_loaded_changed_callbacks:
+            f(self._database_is_loaded)
```
ebrahimebrahim, line 260: To simplify, why not make `db` be the property rather than `database_is_loaded`. Then instead of `_on_database_is_loaded_changed_callbacks` it would be `_on_db_changed_callbacks`. Just feels a bit simpler/cleaner this way, so we are not tracking the same state in two places. Then `database_is_loaded` could just return whether `self.db is None`

(optional suggestion, and maybe ill-advised if I am missing something in this complex redesign)

OpenLIFULib/OpenLIFULib/util.py
```diff
@@ -20,6 +21,14 @@ def __exit__(self, exception_type, exception_value, traceback):
         qt.QApplication.restoreOverrideCursor()
         return False
 
+def get_openlifu_database_parameter_node() -> "OpenLIFUDatabaseParameterNode":
+    """Get the parameter node of the OpenLIFU Database module"""
+    return slicer.util.getModuleLogic('OpenLIFUDatabase').getParameterNode()
+
+def get_cur_db():
```
ebrahimebrahim, line 28: would be handy to have a return type annotation

ebrahimebrahim (approved):
Wonderful work, and much appreciation for cleaning up a convoluted mess that I introduced in the early days!

In my testing everything worked smoothly. I included the transducer tracking wizard in my test. I also tried out the database initialization functionality

OpenLIFULib/OpenLIFULib/util.py
arhowe00: I actually tried this, but what would the annotation be? openlifu.Database doesn't work

ebrahimebrahim: Optional[openlifu.Databse], in quotes?

OpenLIFUDatabase/OpenLIFUDatabase.py
arhowe00: So I agree with your idea, but I wanted this PR to do a minimal change to mimic the behavior, changing places to know when the db is changed and just checking that is obviously simpler but I considered that a refactor beyond the scope of this PR. I'll add it in here though to make it easier

OpenLIFUDatabase/OpenLIFUDatabase.py
arhowe00: I didn't have a reason, but now I think it shouldn't be class level, it should be instance level. I'll change

OpenLIFULib/OpenLIFULib/util.py
ebrahimebrahim: sorry I was rushing, I think it should be `openlifu.db.Database` in quoties, and in the you can put an import under an `if TYPE_CHECKING:`

ebrahimebrahim: (also feel free to just give up and not type-annotate if it ends up being too troublesome)

arhowe00: Solution was this:

```python
from typing import TYPE_CHECKING, Any, List, TypeVar, Type, Optional
# ...
if TYPE_CHECKING:
    from openlifu.db import Database
# ...
def get_cur_db() -> "Optional[Database]":
    """Get the current openlifu.db.Database loaded in the OpenLIFU Database module"""
    return slicer.util.getModuleLogic('OpenLIFUDatabase').db
```

arhowe00:
Okay I made some suggested changes and because I also broadly touch the code again by rebasing in the workflow controls, if reviewers could just do a random workflow with widgets they understand that would be great.

ebrahimebrahim (approved):

sadhana-r (approved):

PR #232, opened 2025-04-08
Implement module specific guided mode behaviors

### LINKED ISSUE #128: Implement module-specific guided mode behaviors

arhowe00:
Closes #128 

## Summary

This PR adds guided mode support to all modules in the Guided mode workflow by first implementing `can_proceed` and `status_text` in a coarse way. The routine is set to make this more fine as we wish.

Then, separately, guided mode hides things in two ways:
1. Static case: The dynamic property `slicer.openlifu.hide-in-guided-mode` can be set in the `.ui` file for anything to be hidden in guided mode.
2. Dynamic case: When more custom logic is required to determine what will be hidden, `enforceGuidedModeVisibility()` needs to be implemented. `Workflow.enforceGuidedModeVisibility()` will then propagate that call.

The visibility logic is centralized in the `Workflow` class. The mixin previously responsible for `enforceGuidedModeVisibility()` was removed.

## For review

Test the complete workflow in guided mode and make sure things you don't like seeing are hidden. **Note, because you will likely be testing this in SlicerOpenLIFU, before turning on Guided Mode, activate user account mode. If you don't activate user account mode, guided mode will say you can skip logging in, but this is not the experience intended in the custom app.**

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -670,7 +670,11 @@ class OpenLIFUDataWidget(ScriptedLoadableModuleWidget, VTKObservationMixin, Guid
         self.ui.loadTransducerButton.clicked.connect(self.onLoadTransducerPressed)
         self.ui.loadPhotoscanButton.clicked.connect(self.onLoadPhotoscanPressed)
 
-        # Connections
+        # ---- Inject guided mode workflow controls ----
+
+        self.inject_workflow_controls_into_placeholder()
+
+        # ---- Connections ----
 
         # These connections ensure that we update parameter node when scene is closed
         self.addObserver(slicer.mrmlScene, slicer.mrmlScene.StartCloseEvent, self.onSceneStartClose)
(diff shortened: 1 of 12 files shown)


ebrahimebrahim (approved):
Super elegant! Tested and worked well! Everything responds appropriately to toggling of guided mode

PR #227, opened 2025-04-02
Introduce guided workflow for placing facial landmarks 

sadhana-r:
Closes #196 
Partly Addresses #219. The fiducial markup nodes used within the wizard and independent of the nodes in the scene. The 'scene' nodes are only updated on Finish.  

- Introduces a guided workflow for annotating the photoscan and skin segmentation mesh
- When points are 'unset', the table shows 'Click to Place XX'. Once placed, the table lists the landmark name.
- When a point is double clicked in the table, the point is unset and the user can click to place again. The position gets reset to 0,0,0
- If the user clicks 'Cancel' at any point, even after moving landmarks, landmarks don't get created in the scene
- When user clicks Finish at the end, fiducial nodes are added to the scene and temp wizard nodes are cleared.
- Next is only enabled once all points are defined.
- Disabled the right click context menu on the markups widget but if the user does delete a control point (i.e. keyboard press or right clicking the control point), this action is blocked and the user is notified.
- If a previously computed facial landmarks node has been invalidated ie. the previously created landmark node in the scene does not have 3 control points, the user is notified before the wizard is started. 
- Control points in the list are indexed based on their names, not position in the list. The user is notified if a control point is selected that has an "invalid" name

Note: I noticed some errors when I get to the registration steps after previously running the wizard. I will fix that when updating the registration workflow as part of #220

PR #234, opened 2025-04-08
Enable scaling of photoscan-volume transform node

### LINKED ISSUE #219: Separate the TT wizard nodes state from the main Slicer scene nodes
When running the transducer tracking wizard, the nodes that are modified by the wizard are a copy of the transform nodes (if any) in the scene. If the user cancels the wizard pipeline mid-way, then the original transform nodes are maintained in the scene. If the wizard is completed till the end, when the user clicks 'Finish', the final result is saved to the scene and overwrites the transform nodes in the scene. 

The same applies to the fiducial nodes created to mark the photoscan and skin segmentation

### LINKED ISSUE #220: Add the ability to adjust the scaling factor of the photoscan-volume transform node
During transducer tracking, fiducial-based registraion is used to initialize registration of the photoscan mesh to the skin mesh. When manually refining the transform node, we should be able to scale/shrink the photoscan as needed if the fiducial initialization is not good.

sadhana-r:
Closes #219 
Closes #220 

- The photoscan-volume registration step now includes a slider that can be used during manual interactive registration to scale the photoscan mesh. Scaling factor range = [0.8,1.2]
- The transform nodes used for tt are local to the wizard and separate to the ones added to the scene at the end. These nodes are cleared when the user cancels the wizard before the end.
- On 'Finish', the transducer tracking results are added to the scene by cloning the wizard-level transform nodes. 
- This PR also includes minor bug fixes to the photoscan and skin-segmentation markup steps

### DIFF
--- a/OpenLIFULib/OpenLIFULib/photoscan.py
+++ b/OpenLIFULib/OpenLIFULib/photoscan.py
@@ -162,10 +162,3 @@ class SlicerOpenLIFUPhotoscan:
         self.model_node.GetDisplayNode().SetViewNodeIDs([node.GetID() for node in viewNodes] if viewNodes else ())
         if self.facial_landmarks_fiducial_node:
             self.facial_landmarks_fiducial_node.GetDisplayNode().SetViewNodeIDs([node.GetID() for node in viewNodes] if viewNodes else ())
-        
-    def set_transform_node(self, transform_node: vtkMRMLTransformNode):
-        
-        self.model_node.SetAndObserveTransformNodeID(transform_node.GetID())
-        
-        if self.facial_landmarks_fiducial_node:
-            self.facial_landmarks_fiducial_node.SetAndObserveTransformNodeID(transform_node.GetID())   
--- a/OpenLIFULib/OpenLIFULib/transducer_tracking_results.py
+++ b/OpenLIFULib/OpenLIFULib/transducer_tracking_results.py
@@ -33,12 +33,9 @@ def add_transducer_tracking_result(
     and will have a bunch of attributes set on it so that we can identify it
     later as a transducer tracking result node.
 
-    Note: This is a placeholder implementation. The format of this function will likely change in the future based on
-    the output of the transducer tracking algorithms. This function can be updated later to, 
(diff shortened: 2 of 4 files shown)


ebrahimebrahim (changes requested):
It seems that something may have broken with the fiducial registration!

After clicking "Initialize photoscan-volume transform" the fiducials don't line up perfectly:

image

I tried with the current main (b3f5ede55ea5b1d37c00408826d4edbfecd6a6ca, so after you merged #227 

(That being said the scaling slider appears to work well, nice!)

ebrahimebrahim:
One other thing, only if it's not too hard: the slider could be a bit finer. Currently it allows for scale values in 0.01 increments, but I can imagine a small adjustment being needed that on a smaller scale than that. Maybe 0.002 increments?

sadhana-r:
> After clicking "Initialize photoscan-volume transform" the fiducials don't line up perfectly:

I think that is expected? Since its a similarity transform, the fiducials are not going to line up perfectly based on how the fiducials are placed. For example, this is an output I get:
Screenshot 2025-04-09 092752
Screenshot 2025-04-09 092758


Like here the right ear seems well aligned, but left and nasion aren't.
If you check the fiducial registration module, there is an RMS error. I could include that in the wizard if hepful?
Screenshot 2025-04-09 092925

ebrahimebrahim:
You are right, my bad! For some reason I had it in my head that there exists an exact similarity transform. There should be an error, as the space of similarity transforms is 7D, short by 2 to guarantee a mapping from 3 generic points to 3 generic points. My tests were probably just unlucky/lucky in a misleading manner.

ebrahimebrahim (approved):

sadhana-r:
@ebrahimebrahim I found a bug. In the event that a user deletes a control point and the order of control points gets rearranged in the fiducial node. However, when running fiducial registration, the registration assumes that the control points are ordered in the same way in both the fixed and moving node. I basically need to add a check and reorder the control points before running registration to account for this. Should I just open an issue and fix this later? Or address it now.

ebrahimebrahim:
Ah good catch, yes please open an issue to fix this later (e.g. in the next milestone).

PR #242, opened 2025-04-09
Revert "Simplify algorithm input widget sub-widget hiding (#232)" (#241)

### LINKED ISSUE #241: AttributeError due to use of non-existent visibility attribute on QLabel
Commit 5c7c514c0d8e1d60c4840c210d086d7c1fddc059 introduced the following error at runtime:

```
AttributeError: 'visibility' does not exist on QLabel and creating new attributes on C++ objects is not allowed
```
This is caused by replacing .hide() / .show() calls with assignment to a non-existent .visibility attribute on Qt widgets. Since these are C++-backed objects, arbitrary attribute creation is not allowed.

Action:

Reverting the commit. Correct approach should use .setVisible(enforced) or .show() / .hide() as needed.

arhowe00:
Closes #241

This reverts commit 5c7c514c0d8e1d60c4840c210d086d7c1fddc059.

Commit 5c7c514c0d8e1d60c4840c210d086d7c1fddc059 introduced the following error at runtime:

```python
AttributeError: 'visibility' does not exist on QLabel and creating new attributes on C++ objects is not allowed
```

This is caused by replacing .hide() / .show() calls with assignment to a non-existent .visibility attribute on Qt widgets. Since these are C++-backed objects, arbitrary attribute creation is not allowed.

### DIFF
--- a/OpenLIFULib/OpenLIFULib/algorithm_input_widget.py
+++ b/OpenLIFULib/OpenLIFULib/algorithm_input_widget.py
@@ -79,12 +79,8 @@ class OpenLIFUAlgorithmInputWidget(qt.QWidget):
         # In this case we just want to hide these three combo box widgets
         for widget_key in ["Protocol", "Transducer", "Volume"]:
             if widget_key in self.inputs_dict:
-                if enforced:
-                    self.inputs_dict[widget_key].label.hide()
-                    self.inputs_dict[widget_key].combo_box.hide()
-                else:
-                    self.inputs_dict[widget_key].label.show()
-                    self.inputs_dict[widget_key].combo_box.show()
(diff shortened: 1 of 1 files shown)


ebrahimebrahim:
Andrew just try `visible` instead of visibility -- I think [it works](https://doc.qt.io/qt-6/qwidget.html#visible-prop) (but give it a quick test :grinning: )

arhowe00:
> Andrew just try `visible` instead of visibility -- I think [it works](https://doc.qt.io/qt-6/qwidget.html#visible-prop) (but give it a quick test 😀 )

Yes it did work and I tested!

ebrahimebrahim (approved):

PR #273, opened 2025-04-17
Update the home module to include database & login

### LINKED ISSUE #263: Update the home module
We've neglected the OpenLIFUHome module. Ensure there are buttons to navigate to all the modules (including database, protocol config, login, etc.)

Maybe make it visually clear how the modules are grouped, e.g. some are part of the guided workflow and some are off by themselves I think

arhowe00:
Closes #263 
Contributes to #231 

This should need minimal review, as I simply added buttons and grouped things. If you'd like to review the changes, build this branch and look at the home module and see whether things look right about the grouping of the modules.

### DIFF
--- a/OpenLIFULogin/OpenLIFULogin.py
+++ b/OpenLIFULogin/OpenLIFULogin.py
@@ -1,41 +1,38 @@
-from typing import Optional, List, Dict, TYPE_CHECKING
+# Standard library imports
 import json
 from enum import Enum
+from typing import Optional, List, Dict, TYPE_CHECKING
 
-from OpenLIFULib.user_account_mode_util import set_user_account_mode_state
+# Third-party imports
 import qt
-
 import vtk
 import numpy as np
 
+# Slicer imports
 import slicer
+from slicer import vtkMRMLScriptedModuleNode
+from slicer.ScriptedLoadableModule import *
 from slicer.i18n import tr as _
 from slicer.i18n import translate
-from slicer.ScriptedLoadableModule import *
+from slicer.parameterNodeWrapper import parameterNodeWrapper
 from slicer.util import VTKObservationMixin
 
-from slicer.parameterNodeWrapper import parameterNodeWrapper
-from slicer import (
-    vtkMRMLScriptedModuleNode,
-)
-
+# OpenLIFULib imports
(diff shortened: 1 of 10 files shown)


OpenLIFUPrePlanning/OpenLIFUPrePlanning.py
```diff
@@ -1,49 +1,61 @@
...
     get_approved_target_ids,
-    get_virtual_fit_result_nodes,
-    get_virtual_fit_approval_for_target,
-    set_virtual_fit_approval_for_target,
-    get_best_virtual_fit_result_node,
     get_approval_from_virtual_fit_result_node,
+    get_best_virtual_fit_result_node,
     get_target_id_from_virtual_fit_result_node,
+    get_virtual_fit_approval_for_target,
+    get_virtual_fit_result_nodes,
```
sadhana-r, line 51: `set_virtual_fit_approval_for_target` need to be added back.

sadhana-r (changes requested):
The updates to the home module look good! I noticed the one missing import that got lost in the re-org. I looked over all of the other modules and didn't notice anything else missing but might be worth doing a quick double check to make sure no imports were accidentaly left out.

OpenLIFUPrePlanning/OpenLIFUPrePlanning.py
arhowe00: Sounds good, nice catch, I rebased it. Maybe it wasn't the best idea to throw that commit in there. @ebrahimebrahim we can automate this stuff right?

OpenLIFULogin/OpenLIFULogin.py
```diff
@@ -1,40 +1,38 @@
...
     get_current_user,
+    openlifu_lz,
 )
-
-from OpenLIFULib.util import (
-    display_errors,
-)
-
 from OpenLIFULib.guided_mode_util import GuidedWorkflowMixin
+from OpenLIFULib.user_account_mode_util import set_user_account_mode_state
```
sadhana-r, line 28: duplicate import

sadhana-r (changes requested):

sadhana-r (approved):

PR #274, opened 2025-04-17
Update virtual fit approvals workflow

### LINKED ISSUE #257: Virtual fit approvals
Ensure that virtual fit approval works:
- Making changes in the scene to the virtual fit transform or its associated target causes the approval to be revoked
- Without an _approved_ virtual fit, the workflow doesn't allow proceeding from the preplanning module (so that would be something that is enforced when guided mode is enabled)

(Most or all of these may already be in place and working, in which case this issue can be closed after a verification of that fact.)

sadhana-r:
Closes #257 

- In guided mode, you should not be able to proceed to transducer tracking without an approved VF result.
- If a target is removed, approval is revoked (if any) and the associated VF results are cleared.
- If a target is modified (point modified or added/removed), approval is revoked (if any) and the user is asked to confirm deletion of VF results.
- In guided mode, if the virtual fit results are cleared, the workflow status updates accordingly. 

`OnNodeRemoved` is triggered when you unload a session. Since the `loaded_session` gets set to `None` before the associated target node is cleared, we don't get prompted about approval being revoked (if any) on the target node when a session is unloaded. 
However, we want to prompt the user that VF results are getting cleared (regardless of approval status). This prompt will therefore pop up when a session is unloaded, and also when a session is loaded (because OnPointModified and PointAddedRemoved are triggered when you load a session). To address this, I have added a flag to the Data module to indicate when  session loading/unloading is in progress. Approval revokation and VF result clearing only gets triggered when nodes are modified/cleared outside of  session loading/clearing.

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1370,6 +1370,9 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
         self._subjects : Dict[str, openlifu.db.subject.Subject] = {} # Mapping from subject id to Subject
         self._folder_deletion_in_progress = False
         self._timer = None 
+        
+        # To avoid triggering certain events related to nodes being added/removed.
+        self.session_loading_unloading_in_progress = False # Used in pre-planning module
 
     def getParameterNode(self):
         return OpenLIFUDataParameterNode(super().getParameterNode())
@@ -1383,6 +1386,9 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
                 it was manually loaded without the context of a session. If True then the scene
                 content is removed.
         """
+
+        self.session_loading_unloading_in_progress = True
+
         loaded_session = self.getParameterNode().loaded_session
         if loaded_session is None:
             return # There is no active session to clear
@@ -1408,6 +1414,8 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
                     self.remove_photoscan(photoscan_id)
 
             clear_transducer_tracking_results(session_id = loaded_session.get_session_id())
+
+            self.session_loading_unloading_in_progress = False
 
     def save_session(self) -> None:
         """Save the current session to the openlifu database.
@@ -1680,6 +1688,8 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
 
         self.clear_session()
 
+        self.session_loading_unloading_in_progress = True  
+
         volume_info = get_cur_db().get_volume_info(session_openlifu.subject_id, session_openlifu.volume_id)
 
        # Create the SlicerOpenLIFU session object; this handles loading volume and targets
@@ -1765,6 +1775,8 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
         for photocollection_reference_number in conflicting_session_photocollections:
             self.remove_photocollection(photocollection_reference_number) 
         self.update_photocollections_affiliated_with_loaded_session()
+
+        self.session_loading_unloading_in_progress = False
 
     # TODO: Don't Need this anymore? Observation should be on the specific TT transforms.
     # def _on_transducer_transform_modified(self, transducer: SlicerOpenLIFUTransducer) -> None:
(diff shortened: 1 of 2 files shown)


sadhana-r:
@arhowe00 When reviewing, you can test out the cases I described in the PR description (works as expected for me) and look over the code. Lmk if you think there is a better way to keep track of when events are triggered due to session loading/clearing besides using that flag in the data module.

sadhana-r:
@arhowe00 reminder to pls review when you can. ty!

arhowe00 (changes requested):
> In guided mode, you should not be able to proceed to transducer tracking without an approved VF result.

This worked for me.

> If a target is removed, approval is revoked (if any) and the associated VF results are cleared.

This initially worked for me. After I removed the target, I got the pop up window. The spacing looks a bit awkward but we can fix this later as an AP matter (@ebrahimebrahim , to what extent can we edit these windows programmatically? If not, there may be lots of them and it could be a good idea to sort the spacing out as we go.).

However, when I did click "OK" and went to place a new target, I got this window (note, there were no targets because I removed the target from example_session). Once I placed the new target, I was getting spammed with this window, and every time I hit "No" it re-appears:
image

> If a target is modified (point modified or added/removed), approval is revoked (if any) and the user is asked to confirm deletion of VF results.

I'm getting similar finicky behavior around this. When I click the lock to edit the target after approving, it does display pop ups, but they seem buggy and spammy. In case this could be specific to macOS or something, to replicate exactly: 1) Load database, 2) Load example_session, 3) Run VF, 4) Approve VF, 5) Click the target, 6) Click the lock, 7) Drag the target around

> In guided mode, if the virtual fit results are cleared, the workflow status updates accordingly.

This seems to work properly.

ebrahimebrahim:
> This initially worked for me. After I removed the target, I got the pop up window. The spacing looks a bit awkward but we can fix this later as an AP matter (@ebrahimebrahim , to what extent can we edit these windows programmatically? If not, there may be lots of them and it could be a good idea to sort the spacing out as we go.).

Regardless of whether it can be done programmatically (something I don't know for sure if it's part of the qt styling layer or not), I'm in favor of holding off on manual adjustments as we go. I'd rather have us do it all at once in a concerted effort later on, so that we agree on a consistent style and method of applying it.

sadhana-r:
Thanks for catching this!! I have re-organized it and added a check which should prevent the pop up from showing up incorrectly.

arhowe00:
Things are looking a lot better so far. The main things that I looked at earlier are fixed, but I have a few questions/comments.

1) After VF I moved the target again and it properly asked if I want to delete the VF result, which I accepted. However, I still saw the transducer positioned where it was after the VF result was made. Do you think it should be deleted?

image

2) When approving a VF and then beginning the process of moving a target, there are 2 pop up windows. There's the one that cancels the approval, and then another one that also optionally accepts the move decision and removes the VF. This seems a bit weird because that may mean that I only intended to move the target but never did.

Both of these are UI/UX and maybe don't even need changes. So regardless on your response, I'd say it's ready to merge!

I also find it interesting that we converged on adding flags during deletion events (I had a similar motif in the protocol config). I'm wondering whether there's an elegant solution that both of us missed, or if a flag is the best way.

Lastly, thanks for handling my TODO about updating the workflow results. Definitely sped things up on that change being made.

arhowe00 (approved):
Things are looking a lot better so far. The main things that I looked at earlier are fixed, but I have a few questions/comments.

1) After VF I moved the target again and it properly asked if I want to delete the VF result, which I accepted. However, I still saw the transducer positioned where it was after the VF result was made. Do you think it should be deleted?

image

2) When approving a VF and then beginning the process of moving a target, there are 2 pop up windows. There's the one that cancels the approval, and then another one that also optionally accepts the move decision and removes the VF. This seems a bit weird because that may mean that I only intended to move the target but never did.

Both of these are UI/UX and maybe don't even need changes. So regardless on whether you make the changes, I'd say it's ready to merge!

I also find it interesting that we converged on adding flags during deletion events (I had a similar motif in the protocol config). I'm wondering whether there's an elegant solution that both of us missed, or if a flag is the best way.

Lastly, thanks for handling my TODO about updating the workflow results. Definitely sped things up on that change being made.

sadhana-r:
> 2. When approving a VF and then beginning the process of moving a target, there are 2 pop up windows. There's the one that cancels the approval, and then another one that also optionally accepts the move decision and removes the VF. This seems a bit weird because that may mean that I only intended to move the target but never did.
> 
The second one doesn't accept/reject the move decision, it confirms deletion of the VF results. The target is always moved. I had initially tried to combine the two pop ups into one single message box but the second one regarding VF deletion needs to pop up regardless of whether VF is approved for the modified target. So in terms of logic, I found it easier to keep them as two separate notifications.

PR #279, opened 2025-04-21
Handle meshroom logs and errors

### LINKED ISSUE #277: Handle meshroom logs and errors
Once https://github.com/OpenwaterHealth/OpenLIFU-python/pull/281 is merged, we should handle

- forwarding Meshroom logs into the slicer log handler
- gracefully handling `CalledProcessError` now that this can be raised

ebrahimebrahim:
Closes #277 

Adds the new Meshroom logger (just a logger called "Meshroom") to the slicer logging world. (This required adding a mode to the log handler that doesn't show any dialogs, because Meshroom emits log calls from a separate thread and Qt was very unhappy with me trying to parent an error dialog from one thread to a slicer main window in another thread.)

Also catches a CalledProcessError and gives an error dialog that essentally says "hey this error isn't slicer's fault, it's meshroom!"

For review: Please just eyeball the code that this won't break anything important, and since it touches the log handler which is involved in a lot of stuff maybe just build the application and make sure nothing is obviously broken. Optionally try photoscan generation, but you don't have to. Thanks!

### DIFF
--- a/OpenLIFULib/OpenLIFULib/Resources/python-requirements.txt
+++ b/OpenLIFULib/OpenLIFULib/Resources/python-requirements.txt
@@ -1,2 +1,2 @@
-openlifu==0.2.0
+openlifu==0.2.1
 bcrypt
--- a/OpenLIFULib/OpenLIFULib/util.py
+++ b/OpenLIFULib/OpenLIFULib/util.py
@@ -56,15 +56,26 @@ def display_errors(f):
     return f_with_forwarded_errors
 
 class SlicerLogHandler(logging.Handler):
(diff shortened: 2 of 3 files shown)


ebrahimebrahim:
Oh and btw you can see the meshroom logs in Slicer's log and you will see a lot of errors show up there. That's because we currently send all of meshroom's stderr to error level logging. We should relax that to warning level. That'll be done later: https://github.com/OpenwaterHealth/OpenLIFU-python/issues/285

So no worries about too much error noise in the logs

arhowe00:
Looks good. I successfully built and ran the app.

arhowe00 (approved):

PR #283, opened 2025-04-22
Refine transducer tracking approvals workflow

### LINKED ISSUE #258: Photoscan approvals
Ensure that photoscan approval works:
- (There is nothing that should cause a photoscan approval to be revoked, right? There is no way to mutate photoscans in the scene)
- It is not possible to proceed with transducer tracking with an unapproved photoscan (we can decide whether this should always be enforced or only in guided mode). It should be clear to the user why the can't proceed in this situation.

### LINKED ISSUE #259: Transducer tracking approvals
Ensure that transducer tracking approval works:
- Making changes in the scene to the transducer tracking transforms _or the transducer transform_ (because it is what's used in the sonication planner), ~or the sonication target causes the transducer tracking approvals to be revoked.~
- Without fully _approved_ transducer tracking results, the workflow doesn't allow proceeding from the transducer tracking module (so that would be something that is enforced when guided mode is enabled)

sadhana-r:
Partially addresses #259 
Closes #258 

Photoscan:
- [x] Toggling approval updates the application state appropriately
- [x] Saving a session saves the approval state
- [x] Loading a session loads the approval state
- [x] (There is nothing that should cause a photoscan approval to be revoked, right? There is no way to mutate photoscans in the scene) - No 
- [x] In guided mode, It is not possible to proceed with transducer tracking with an unapproved photoscan.. It should be clear to the user why the can't proceed in this situation. When not in guided mode, display a warning is the photoscan is not approved.
- [x] Revoking approval on photoscan revokes approval on the corresponding transducer tracking result (if previously computed)

Transducer tracking
- [x] - Currently, unless you approve the transform during the wizard workflow, you cannot go back and approve a previously computed transform. A user should be able to click through wizard and see previous tt results and toggle approval. If a fiducial is moved, tt results are cleared and unapproved (wizard-level). **This does not work well when loading a session with a tt result and will be addressed as part of: https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/289.**
- [x] - Each step of the wizard is blocking - cannot proceed to transducer-volume tracking or click 'Finish' without approving the transform.
- [x] - In guided mode - check that the selected target has an approved VF result. Since in the case with multiple targets, you technically would be able to proceed as long as one of the targets is approved. In non-guided mode, provide a warning if VF is not approved for the selected target or the selected photoscan is not approved.
- [x] Toggling approval updates the application state appropriately (tt)
- [x] Saving a session saves the approval state (tt)
- [x] Loading a session loads the approval state (tt)
- [ ] Making changes in the scene to the transducer tracking transforms or the transducer transform (because it is what's used in the sonication planner), or the sonication target causes the transducer tracking approvals to be revoked. **Currently transducer tracking approval is revoked if the tracking transforms are modfied. Larger changes are needed to revoke approval when the target or transducer transform are modified. This is because the tracking results currently don't keep track of the affiliated target. I will address this last part in a separate PR.** 
- [x] Without fully approved transducer tracking results, the workflow doesn't allow proceeding from the transducer tracking module (so that would be something that is enforced when guided mode is enabled)

### DIFF
--- a/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
+++ b/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
@@ -58,7 +58,7 @@ from OpenLIFULib.transducer_tracking_wizard_utils import (
     get_threeD_transducer_tracking_view_node,
 )
 from OpenLIFULib.user_account_mode_util import UserAccountBanner
-from OpenLIFULib.util import add_slicer_log_handler, BusyCursor, replace_widget, clone_node
+from OpenLIFULib.util import add_slicer_log_handler, BusyCursor, replace_widget, get_cloned_node
 from OpenLIFULib.virtual_fit_results import get_best_virtual_fit_result_node,  get_virtual_fit_approval_for_target
 
 # These imports are for IDE and static analysis purposes only
@@ -146,7 +146,7 @@ class FacialLandmarksMarkupPageBase(qt.QWizardPage):
 
     def _initialize_facial_landmarks_fiducial_node(self, node_name: str, existing_landmarks_node=None) -> vtkMRMLMarkupsFiducialNode:
         if existing_landmarks_node:  # Clone the existing node if valid
-            node = clone_node(existing_landmarks_node)
+            node = get_cloned_node(existing_landmarks_node)
             node.SetName(node_name)  # Use the provided node name
             node.GetDisplayNode().SetVisibility(False)
 
@@ -465,7 +465,7 @@ class PhotoscanVolumeTrackingPage(qt.QWizardPage):
             existing_transform_node = self.wizard()._logic.get_transducer_tracking_result_node(
                 photoscan_id = self.wizard().photoscan.get_id(),
                 transform_type = TransducerTrackingTransformType.PHOTOSCAN_TO_VOLUME)
-            self.photoscan_to_volume_transform_node = clone_node(existing_transform_node)
+            self.photoscan_to_volume_transform_node = get_cloned_node(existing_transform_node)
             self.photoscan_to_volume_transform_node.GetDisplayNode().SetVisibility(False)
             self.photoscan_to_volume_transform_node.RemoveAttribute('isTT-PHOTOSCAN_TO_VOLUME')
             self.transform_approved = get_approval_from_transducer_tracking_result_node(existing_transform_node)
@@ -633,7 +633,7 @@ class TransducerPhotoscanTrackingPage(qt.QWizardPage):
             existing_transform_node = self.wizard()._logic.get_transducer_tracking_result_node(
                 photoscan_id = self.wizard().photoscan.get_id(),
                 transform_type = TransducerTrackingTransformType.TRANSDUCER_TO_VOLUME)
-            self.transducer_to_volume_transform_node = clone_node(existing_transform_node)
+            self.transducer_to_volume_transform_node = get_cloned_node(existing_transform_node)
             self.transducer_to_volume_transform_node.GetDisplayNode().SetVisibility(False)
             self.transducer_to_volume_transform_node.RemoveAttribute('isTT-TRANSDUCER_TO_VOLUME')
             self.transform_approved = get_approval_from_transducer_tracking_result_node(existing_transform_node)
@@ -878,14 +878,16 @@ class TransducerTrackingWizard(qt.QWizard):
                 transform_type= TransducerTrackingTransformType.PHOTOSCAN_TO_VOLUME,
                 approval_status=self.photoscanVolumeTrackingPage.transform_approved,
                 photoscan=self.photoscan,
-                transducer=self.transducer)
+                transducer=self.transducer,
+                clone_node = True)
         if self.transducerPhotoscanTrackingPage.transducer_to_volume_transform_node:
             self._logic.add_transducer_tracking_result(
                 transform_node=self.transducerPhotoscanTrackingPage.transducer_to_volume_transform_node,
                 transform_type= TransducerTrackingTransformType.TRANSDUCER_TO_VOLUME,
                 approval_status=self.transducerPhotoscanTrackingPage.transform_approved,
                 photoscan=self.photoscan,
-                transducer=self.transducer)
+                transducer=self.transducer,
+                clone_node = True)
 
         self.clearWizardNodes() #remove the wizard-level node
 
@@ -1871,7 +1873,7 @@ class OpenLIFUTransducerTrackerLogic(ScriptedLoadableModuleLogic):
         if volume_facial_landmarks_node is None:
             # By default, turn visibility off so that the node is visible before the 
             # appropriate view node IDs are set. 
-            volume_facial_landmarks_node : vtkMRMLMarkupsFiducialNode = clone_node(fiducial_node)
+            volume_facial_landmarks_node : vtkMRMLMarkupsFiducialNode = get_cloned_node(fiducial_node)
             volume_facial_landmarks_node.SetName(f"{volume_name}-faciallandmarks")
 
             # Ensure that visibility is turned off
@@ -1914,7 +1916,8 @@ class OpenLIFUTransducerTrackerLogic(ScriptedLoadableModuleLogic):
                                        transform_type: TransducerTrackingTransformType,
                                        approval_status: bool,
                                        photoscan: SlicerOpenLIFUPhotoscan,
-                                       transducer : SlicerOpenLIFUTransducer) -> vtkMRMLTransformNode:
+                                       transducer : SlicerOpenLIFUTransducer,
+                                       clone_node = False) -> vtkMRMLTransformNode:
         """Initializes and returns a transducer tracking result transform node. 
         This new transform node is added to the scene and assigned the necessary attributes to 
         identify it as a transducer tracking result of type 'PHOTOSCAN_TO-VOLUME' or 'TRANSDUCER_TO_VOLUME.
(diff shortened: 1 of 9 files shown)


sadhana-r:
@arhowe00 the PR description describes the different scenarios to consider. To review this, I would load a few sessions in guided mode and regular mode and play round with toggling different combinations of approvals (i.e. VF approval, photoscan approval and TT approval) and see if you notice any weird behaviour. You could also test modifying the transudcer tracking transforms i.e. the `TT transducer-volume ...` and `TT photoscan-volume ...` nodes that get added to the scene after transducer tracking and check if tracking approval gets revoked. I have added status messages to the transducer tracking module that should accurately reflect the state of photoscan and transducer tracking approvals at all times.There are a couple of cases (noted in bold) that I will address in a separate PR.

sadhana-r:
(rebased to main)

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -35,7 +35,7 @@
 )
 from OpenLIFULib.coordinate_system_utils import numpy_to_vtk_4x4
 from OpenLIFULib.events import SlicerOpenLIFUEvents
-from OpenLIFULib.guided_mode_util import GuidedWorkflowMixin
+from OpenLIFULib.guided_mode_util import GuidedWorkflowMixin, get_guided_mode_state
```
arhowe00, line 38:
```suggestion
from OpenLIFULib.guided_mode_util import get_guided_mode_state, GuidedWorkflowMixin
```

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -58,8 +58,8 @@
     get_threeD_transducer_tracking_view_node,
 )
 from OpenLIFULib.user_account_mode_util import UserAccountBanner
-from OpenLIFULib.util import add_slicer_log_handler, BusyCursor, replace_widget
-from OpenLIFULib.virtual_fit_results import get_best_virtual_fit_result_node
+from OpenLIFULib.util import add_slicer_log_handler, BusyCursor, replace_widget, get_cloned_node
```
arhowe00, line 61:
```suggestion
from OpenLIFULib.util import add_slicer_log_handler, BusyCursor, get_cloned_node, replace_widget
```

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -1010,6 +1046,12 @@ def __init__(self, photoscan : "openlifu.nav.photoscan.Photoscan"):
 
     def onFinish(self):
         self.resetViewNodes()
+
+        # Update the photoscan approval status in the underlying openlifu photoscan object
+        self.logic.update_photoscan_approval(
+            photoscan = self.photoscan,
+            approval_state = self.photoscanPreviewPage._photoscan_approved)
```
arhowe00, line 1053: Beware:
```
Argument of type "Unknown | bool | None" cannot be assigned to parameter
"approval_state" of type "bool" in function "update_photoscan_approval" 
  Type "Unknown | bool | None" is not assignable to type "bool"
    "None" is not assignable to "bool" (Pyright reportArgumentType)     
```

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -1438,17 +1492,29 @@ def checkCanRunTracking(self,caller = None, event = None) -> None:
...
+                target_id = fiducial_to_openlifu_point_id(target)
+                target_is_approved = slicer.util.getModuleLogic('OpenLIFUPrePlanning').get_virtual_fit_approval(target_id)
+
+            if transducer.surface_model_node is None: # Check that the selected transducer has an affiliated registration surface model
                 self.ui.runTrackingButton.enabled = False
                 self.ui.runTrackingButton.setToolTip("The selected transducer does not have an affiliated registration surface model, which is needed to run tracking.")
+            elif get_guided_mode_state() and (photoscan and not photoscan.photoscan_approved): # GM: Check that the selected photoscan is approved
+                self.ui.runTrackingButton.enabled = False 
+                self.ui.runTrackingButton.setToolTip("The selected photoscan has not been approved for transducer tracking.")
+            elif get_guided_mode_state() and (target and not target_is_approved): # GM: Check that virtual fit is approved for the selected target
```
arhowe00, line 1508: `target_is_approved` is possibly unbound

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -1637,6 +1763,18 @@ def get_approved_photoscan_ids(self) -> List[str]:
...
+        """Return a list of photoscan IDs that are approved for transducer tracking"""
+        session = get_openlifu_data_parameter_node().loaded_session
+        if not session and not get_openlifu_data_parameter_node().loaded_photoscans:
+            return []
+        if session:
+            approved_photoscans = [id for id, wrapped_photoscan in session.affiliated_photoscans.items() if wrapped_photoscan.photoscan.photoscan_approved]
+        elif get_openlifu_data_parameter_node().loaded_photoscans:
+            print()
+            approved_photoscans = [id for id, slicer_photoscan in get_openlifu_data_parameter_node().loaded_photoscans.items() if slicer_photoscan.is_approved()]
+        return approved_photoscans
```
arhowe00, line 1776: `approved_photoscans` is possibly unbound. Also, is the print vestigial? You could try this:

```diff
     def get_photoscan_ids_with_approval(self) -> List[str]:
         """Return a list of photoscan IDs that are approved for transducer tracking"""
         session = get_openlifu_data_parameter_node().loaded_session
+        approved_photoscans = []
         if not session and not get_openlifu_data_parameter_node().loaded_photoscans:
-            return []
+            return approved_photoscans
         if session:
             approved_photoscans = [id for id, wrapped_photoscan in session.affiliated_photoscans.items() if wrapped_photoscan.photoscan.photoscan_approved]
         elif get_openlifu_data_parameter_node().loaded_photoscans:
-            print()
             approved_photoscans = [id for id, slicer_photoscan in get_openlifu_data_parameter_node().loaded_photoscans.items() if slicer_photoscan.is_approved()]
         return approved_photoscans
```

arhowe00 (changes requested):
I did a look at both the guided mode and non-guided mode experience of making sure that when togging approvals, the application state updates in a way that I'd expect. Approvals are revoked in the right spots. One thing I'll note, which may not be a problem, is that after revoking a transducer tracking approval and the landmarks have to be replaced, they retain their previous positions in the wizard. This may be a convenience sometimes, but I also assume that the photoscan might have changed which would make those coordinates invalid. Maybe they should be reset to the origin?  Either way, it's not essential, since the user still has to replace and approve them.

So to recap: The workflow looks good. I brushed over the code and have a few comments but most of them are styling-related or linter errors. I did not do a deep code review.

sadhana-r:
Thanks for your comments! I made the style/linting changes.

> One thing I'll note, which may not be a problem, is that after revoking a transducer tracking approval and the landmarks have to be replaced, they retain their previous positions in the wizard. This may be a convenience sometimes, but I also assume that the photoscan might have changed which would make those coordinates invalid. Maybe they should be reset to the origin? Either way, it's not essential, since the user still has to replace and approve them.

Are you referring to the landmarks retaining their previous position or transforms? If landmarks, I don't think the photoscan position can change relative to the landmarks - then it would likely be a new photoscan with new photoscan ID. I kept the fiducials the same so that the user can technically go back and approve transducer tracking again without changing anything if they wanted i.e. if they revoked photoscan approval by mistake. I am going to refine the workflow a bit as part of https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/289 and could add some status messages within the wizard to make things more clear.

arhowe00:
> Are you referring to the landmarks retaining their previous position or transforms?

This may be way too nitpicky and beyond the scope of this issue (or even not an issue). But just to be clear, let's say I place the landmarks on the photoscan:

image

If I then finish both image registration steps and approve transducer tracking, and then revoke the photoscan approval, the previous landmarks remain when I re-run transducer tracking.

image

However, I reckon that if I load a different photoscan, they will disappear.  The reason I initially complained is because I assumed that revoking photoscan approval means that the photoscan should also would not be used for transducer tracking, or the landmarks would be reset.

arhowe00 (approved):
Great work on the changes. Applied to the confusion above. Feel free to make changes about it and merge or don't make changes and merge.

sadhana-r:
> However, I reckon that if I load a different photoscan, they will disappear. The reason I initially complained is because I assumed that revoking photoscan approval means that the photoscan also would not / could not be used for transducer tracking, and so the landmarks would be reset. I'd say if loading a different photoscan resets them, then it's a non-issue. But are the landmarks linked to the photoscan so that they would disappear if using a different one?

Yep, if a different photoscan is loaded, the landmarks would reset and not show up in the wizard. My thinking is that approval/unapproval is more about whether the quality of the photoscan is good enough for transducer tracking, but unapproving the photoscan does not neccesarily invalidate the position of the landmarks relative to the photoscan. The user may want to still keep those landamrks and maybe compare registration with different photoscans etc. 
But you bring up a good point that for guided mode (where we are most restrictive) we need to catch the case where the user unapproves a photoscan after running tracking. I have updated the guided mode workflow controls to only allow the user to proceed with sonicaton if the approved transducer tracking result is associated with an approved photoscan. Before it was just checking for an approved tt result, but not checking the photoscan state.

PR #286, opened 2025-04-23
Bump openlifu to v0.3.0

ebrahimebrahim:
Bumping openlifu version for a release.

@arhowe00 I tested a workflow with each module and everything worked, with one exception: no solution analysis table was displayed. This is possibly totally acceptable because you are still working on #284 but I just want to confirm with you first.


arhowe00 (approved):
That is correct -- the solution analysis table should be broken. We need to merge the PR which adds support for the complete solution analysis, which would be after this version bump PR (if we merge this before that)

PR #287, opened 2025-04-23
Dont clutter the log with lifu interface

### LINKED ISSUE #265: Don't clutter the log with LIFUInterface
> once the device is connection and monitoring starts, the python console gets clogged with messages, preventing effective use. There should be a way to disconnect.

This comment applies to both test mode and non-test-mode.

When in test mode, we can just remove those print statements that currently just show us that LIFUInterface is being observed. Maybe we can comment them out for now and put a little code comment explaining that these would be useful to uncomment if the event loop stuff needs to be debugged.

When in non-test-mode, there are log messages coming from LIFUInterface from the openlifu layer. We may be able to suppress those (or reduce them) in SlicerOpenLIFU if we set the logging level appropriately for LIFUInterface's logger. If the clutter is in the root logger, however, then we should open an issue on OpenLIFU-python to adjust that.

arhowe00:
Closes #265 

- [x] Since we can't test the HW, let's wait for @ebrahimebrahim to finish messing around with logging for meshroom stuff, so maybe after combining our PRs we assume everything will work out.

## For Review

Make sure the meshroom stuff is finished, then briefly inspect code. Note: Because there were not specifics in #265, I'm not sure whether I commented all the stuff out that was desired. Furthermore, is it enough just to set the logger for "LIFUInterface"?

You probably should not have to build and test anything.

### DIFF
--- a/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
+++ b/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
@@ -722,16 +722,18 @@ class OpenLIFUSonicationControlLogic(ScriptedLoadableModuleLogic):
             f(self._run_hardware_status)
 
     def on_lifu_device_connected(self, *args, **kwargs):
-        print("on_lifu_device_connected called in sonication control module. The arguments given in were:") # TODO: Remove when hardware device works without issues
-        print("Positional arguments:", args) # TODO: Remove when hardware device works without issues
-        print("Keyword arguments:", kwargs) # TODO: Remove when hardware device works without issues
+        # This would be useful to uncomment if debugging hardware/software integration
+        #print("on_lifu_device_connected called in sonication control module. The arguments given in were:")
+        #print("Positional arguments:", args)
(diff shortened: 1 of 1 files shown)


ebrahimebrahim:
(I am finished messing around with logging so I checked it off)

ebrahimebrahim:
(Heads up that I am rebasing since this actually needs openlifu v0.3.0 to work with the dvc data it's associated with)

ebrahimebrahim (approved):
Nice, I think this will work. We'll find out when they test on hardware

ebrahimebrahim:
(I did build and test btw, it worked when I went to openlifu v0.3.0)

ebrahimebrahim:
> Furthermore, is it enough just to set the logger for "LIFUInterface"?

Oh, hmm looking at the openlifu code maybe we also would have wanted to include the logger called "UART".

ebrahimebrahim:
I also found LIFUTXDevice and LIFUHVController

PR #290, opened 2025-04-24
Update workflow for revoking transducer tracking approval

### LINKED ISSUE #259: Transducer tracking approvals

### LINKED ISSUE #289: Make it possible to skip to the TT wizard tracking pages when a tt result exists
When you load a session that has a previously computed transducer tracking result, you cannot view the transforms in the wizard since the fiducials are no longer in the scene ( fiducials don't get saved to the session) and the wizard doesn't allow you to proceed without placing new fiducials. If there is a previous result in the scene, the wizard should instead direct you to the `PhotoscanVolumeTrackingPage` by default. We can add a button like `Modify Fiducials` that the user can click to go back to page 1 and modify the landmarks.

sadhana-r:
Closes #259 
Closes #289 

- When a transduer tracking result is computed, the result is copied to the current transducer transform. An observer is added to the current transducer transform so that it revokes tt approval for the associated "matching_transform" when modified.
- When approving a photoscan is a session-based workflow, added a check to enforce that only one photoscan can be approved per session.
- When loading a session, if there is a photoscan (could be approved or unapproved) associated with an approved transducer tracking result, we check whether the current transducer transform matches the tracking result and set the "matching_transform" attribute on that transform if it is. This will then revoke approval on the tracking result if the transform is moved. 
- Only one transducer tracking result can be approved within a session. If a user computes a second transducer tracking result, approval of the previously computed result is revoked.
- The wizard now includes clear messaging when a previously computed result is being modified within the wizard. In order to make any edits to an existing result, the user has to first click to revoke approval before being able to make changes. 

When reviewing, it would be helpful if you could just look over the code and play around with the wizard and test out creating a new result from scratch, editing a previous result, computing a second result etc to see if there are any edge cases I missed.

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1835,7 +1835,7 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
             tt_results_openlifu = session_openlifu.transducer_tracking_results,
             session_id = session_openlifu.id,
             transducer = newly_loaded_transducer.transducer.transducer,
-            replace=True, # If there happen to already be some virtual fit result nodes that clash, loading a session will silently overwrite them.
+            replace=True, # If there happen to already be some transducer tracking result nodes that clash, loading a session will silently overwrite them.
         )
 
         for (transducer_to_volume_node, photoscan_to_volume_node) in newly_added_tt_result_nodes:
@@ -1876,7 +1876,32 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
             self.remove_photocollection(photocollection_reference_number) 
         self.update_photocollections_affiliated_with_loaded_session()
 
-        self.session_loading_unloading_in_progress = False
+        # Check if there are any approved photoscans with approved transducer tracking results
+        approved_photoscan_id = self.getParameterNode().loaded_session.get_transducer_tracking_approvals()
+        if not approved_photoscan_id:
+            return  # No approved photoscan ID, so we can stop here
+
+        approved_photoscan_id = approved_photoscan_id[0]
+        approved_tracking_result = None
+
+        for transducer_to_volume_node, _ in newly_added_tt_result_nodes:
+            current_photoscan_id = get_photoscan_id_from_transducer_tracking_result(transducer_to_volume_node)
+            if current_photoscan_id == approved_photoscan_id:
+                if newly_loaded_transducer.is_matching_transform(transducer_to_volume_node):
+                    approved_tracking_result = transducer_to_volume_node
+                    break  # Found a match, no need to continue searching
+
+        if approved_tracking_result:
+            newly_loaded_transducer.set_current_transform_to_match_transform_node(approved_tracking_result)
+        else:
+            # revoke transducer tracking approval
+            transducer_tracking_widget = slicer.modules.OpenLIFUTransducerTrackerWidget
+            transducer_tracking_widget.revokeTransducerTrackingApprovalIfAny(
+                photoscan_id=approved_photoscan_id,
+                reason="The transducer transform does not match the approved tracking result."
+            )
+
+        self.session_loading_unloading_in_progress = False  
 
     # TODO: This should be a widget level function
     def _on_transducer_transform_modified(self, transducer: SlicerOpenLIFUTransducer) -> None:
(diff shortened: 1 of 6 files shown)


OpenLIFUData/OpenLIFUData.py
```diff
@@ -1874,23 +1876,47 @@ def load_session(self, subject_id, session_id) -> None:
...
+
+    # TODO: This should be a widget level function
+    def _on_transducer_transform_modified(self, transducer: SlicerOpenLIFUTransducer) -> None:
+
+        matching_transform_id = transducer.transform_node.GetAttribute("matching_transform")
+        if matching_transform_id:
+            # If its a transducer tracking node, revoke approval if approved
+            transform_node = slicer.mrmlScene.GetNodeByID(matching_transform_id)
+            if transform_node:
+                photoscan_id = get_photoscan_id_from_transducer_tracking_result(transform_node)
```
ebrahimebrahim, line 1914: I am getting a ValueError "Invalid transducer tracking result type" at this line whenever I try to move the current transducer transform after performing a virtual fit.

This is with the demo_4_tt session, and after only doing a virtual fit.

It seems that after performing a virtual fit the mathcing_transform attribute is set.
But there's no "TT:photoscanID" which causes a complaint at this line

sadhana-r: good catch! I have pushed a fix now

OpenLIFUData/OpenLIFUData.py
```diff
@@ -1956,7 +1985,8 @@ def load_transducer_from_file(self, filepath:str) -> None:
             ]
         }
         
-        self.load_transducer_from_openlifu(transducer, transducer_abspaths_info)
+        newly_loaded_transducer = self.load_transducer_from_openlifu(transducer, transducer_abspaths_info)
+        newly_loaded_transducer.observe_transform_modified(self._on_transducer_transform_modified)
```
ebrahimebrahim, line 1989: (*)

OpenLIFUData/OpenLIFUData.py
```diff
@@ -1803,6 +1805,7 @@ def load_session(self, subject_id, session_id) -> None:
             transducer_matrix_units = session_openlifu.array_transform.units,
             replace_confirmed = True,
         )
+        newly_loaded_transducer.observe_transform_modified(self._on_transducer_transform_modified)
```
ebrahimebrahim, line 1808: You could stick this `observe_transform_modified` into the end of the definition of `load_transducer_from_openlifu` -- then you'd only have to include it once and you wouldn't need (*).

That would also better guarantee your observer getting added if anyone uses `load_transducer_from_openlifu` in some other way later.

ebrahimebrahim (approved):
I tested each of your bullet points and everything worked very cleanly!

> Only one transducer tracking result can be approved within a session. If a user computes a second transducer tracking result, approval of the previously computed result is revoked.

This worked well but the revoking happened silently -- I wonder if it makes sense to alert the user about it? Well... maybe it's good as is because maybe it's the natural expectation of the user... let's leave it as is for now.


---

One thing I noticed in the code that isn't really related to this PR but I thought you should be aware of: I saw in `PhotoscanPreviewWizard` that you have a line

```
self.logic = OpenLIFUTransducerTrackerLogic()
```

So this is actually creating a _new instance_ of OpenLIFUTransducerTrackerLogic, which is different from _the_ OpenLIFUTransducerTrackerLogic that belongs to the widget (and that you'd retrieve by doing `slicer.util.getModuleLogic` or whatever). It doesn't seem to break anything here, but if you ever start seriously using OpenLIFUTransducerTrackerLogic to record application state then it could bring up some weird bugs later. Nothing to solve just to be aware of

sadhana-r:
> So this is actually creating a _new instance_ of OpenLIFUTransducerTrackerLogic, which is different from _the_ OpenLIFUTransducerTrackerLogic that belongs to the widget (and that you'd retrieve by doing `slicer.util.getModuleLogic` or whatever). It doesn't seem to break anything here, but if you ever start seriously using OpenLIFUTransducerTrackerLogic to record application state then it could bring up some weird bugs later. Nothing to solve just to be aware of

Thanks for pointing this out! I addressed it in this PR to avoid further issues down the line. Since it was being initialized again in the wizard as well.

sadhana-r:
> This worked well but the revoking happened silently -- I wonder if it makes sense to alert the user about it? Well... maybe it's good as is because maybe it's the natural expectation of the user... let's leave it as is for now.

Yep, I also considered adding in an alert but decided against it since it seems like the natural expectation. And anyways, revoking approval doesn't remove the previously computed tracking result nodes from the scene. So the user can easily go back and re-approve the result associated with the intial photoscan if they decide to.

PR #307, opened 2025-04-28
Make it possible to import a photocollection from disk into the database

### LINKED ISSUE #268: Make it possible to import a photocollection from disk into the database
Add a button that imports a photocollection from disk.
So it's the same thing that the import-from-android functionality currently does, but just gives a file dialog to browse for any old folder.
(Probably this should share code with the import-from-android-functionality?)
This should only be visible in SlicerOpenLIFU, not the app.
So it can just be one the "hidden in guided mode" things.

arhowe00:
Closes #268 

## For review

This should need minimal review, but you can check that you are able to import a photocollection as a directory of photos to a subject & session. I made it so you can choose any alphanumeric reference number.

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -278,7 +278,7 @@ class AddNewVolumeDialog(qt.QDialog):
 class StartPhotocollectionCaptureDialog(qt.QDialog):
     """ Add new photocollection dialog """
 
-    MINIMUM_NUMBER_OF_PHOTOS_FOR_PHOTOSCAN=20
+    MINIMUM_NUMBER_OF_PHOTOS_FOR_PHOTOSCAN=1
 
     def __init__(self, parent="mainWindow"):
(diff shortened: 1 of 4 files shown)


OpenLIFUData/OpenLIFUData.py
```diff
@@ -472,6 +472,77 @@ def customexec_(self):
         }
         return (returncode, photoscan_dict)
 
+class ImportPhotocollectionFromDiskDialog(qt.QDialog):
+    """Import photocollection from disk dialog."""
+
+    MINIMUM_NUMBER_OF_PHOTOS_FOR_PHOTOSCAN = 20
```
sadhana-r, line 478: Same for this

sadhana-r (approved):
Looks good to me! I included reminders to reduce the minimum number of photoscans requirement. 
I'm going to approve but one UI design thought: might be better to have the `Start Photocollection Capture` and `Import Photocollctions from Disk` buttons be side by side since they have the same action.  And the `Start Photoscan Generation` can be the larger button below that.  Basically swap the import photocollection and photoscan generation button.

PR #309, opened 2025-04-28
Enforce solution approval to proceed in guided mode

### LINKED ISSUE #260: Solution approvals
Ensure that solution approval works:
- Toggling approval updates the ~~application state~~ in-scene and in-database `Solution` appropriately
- ~~Saving a session saves the approval state~~ (Solutions are saved upon generation, and there is no mechanism for explicitly saving solutions, so never mind!)
- ~~Loading a session loads the approval state~~ (The approval is contained in the Solution and there is not a concept of loading solutions at the moment, so never mind!)
- Revoking of solution approval will be handled by #261, not here
- Without an _approved_ solution, the workflow doesn't allow proceeding from the preplanning module (so that would be something that is enforced when guided mode is enabled)

sadhana-r:
Closes #260


sadhana-r:
This is a quick one. You can check that in guided mode, you can only proceed to sonication control if the solution is approved.

arhowe00 (approved):
Checked it out and tested. Works as stated.

Issue #25, opened 2024-07-08
Volume Review/Adjust Color/Alpha maps (SOFTREQ-33)


peterhollender:
The user must be able to adjust the false-color and transparency maps for scalar volumes

Notes:
nan

Updated on 2024-07-08 15:31:11

PR #310, opened 2025-04-28
Propagate param node updates to QSettings and GUI (#308)

### LINKED ISSUE #308: Database parameter node should propagate updates
Several concepts share values that are stored in the Database parameter node (namely, the databaseDirectory parameter value). When the parameter node is updated, all related elements should also have those updates propagated to:

- The QSettings stored in the app config (usually somewhere in `~/.config`)
- The Database module GUI

This will make it easier to complete [OpenLIFU-app #25](https://github.com/OpenwaterHealth/OpenLIFU-app/issues/25), which attempts to silently connect to the database given the QSettings. Instead of setting the QSettings and then attempting to silent connect, it should set the Database parameter node details, which would propagate updates to the settings and GUI, and then silently connect using `OpenLIFUDatabaseWidget.onLoadDatabaseClicked()`. It's important to propagate updates to the GUI, because `OpenLIFUDatabaseWidget.onLoadDatabaseClicked()` uses the text in the line edit (`OpenLIFUDatabaseWidget.ui.databaseDirectoryLineEdit.currentPath`) to load the database.

arhowe00:
Closes #308 
Contributes to OpenLIFU-app #25

This PR makes it so that changes to the Database parameter node updates the QSettings and the GUI (the GUI part was already happening through the `self._parameterNodeGuiTag = self._parameterNode.connectGui(self.ui)` construct). This will streamline setting the value from a default in external modules (external modules only need to update the Database parameter node) and default loading of those locations.

## For review

Just check the code. If you want, try also loading databases at custom paths and making sure the line edit changes.

### DIFF
--- a/OpenLIFUDatabase/OpenLIFUDatabase.py
+++ b/OpenLIFUDatabase/OpenLIFUDatabase.py
@@ -120,7 +120,6 @@ class OpenLIFUDatabaseWidget(ScriptedLoadableModuleWidget, VTKObservationMixin, 
         self.initializeParameterNode()
 
         # Call the routine to update from data parameter node
-        self.onDataParameterNodeModified()
         self.updateWorkflowControls()
 
     def cleanup(self) -> None:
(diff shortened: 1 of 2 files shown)


sadhana-r (approved):
looks good!

Issue #282, opened 2025-04-22
Add setting parameter constraints and API for bounding values in protocol config


arhowe00:
After having merged the solution analysis PR in [OpenLIFU-python PR#250](https://github.com/OpenwaterHealth/OpenLIFU-python/pull/250), a clear definition of Parameter constraints was made. The Protocol Config module should create a custom form for this that allows definition of `ParameterConstraint`s. This definition should allow selection from operators and whether ranges or fixed values issue warnings or errors.

In addition, creating an API for fixing bounds of the dynamically created forms from [SlicerOpenLIFU PR#225](https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/225) will be helpful, as currently the bounds are chosen from a generic collection of magic numbers.

---

**Traceability Information:**
- Issue: [SlicerOpenLIFU#282](https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/282)
- SOFTREQ: [SlicerOpenLIFU#12](https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/12)
- Comment: This issue is related to the requirements defined in [SlicerOpenLIFU#12](https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/12).

_This information was populated automatically by a script on 2025-06-09T18:05:39Z._

ebrahimebrahim:
@arhowe00 Is this complete? Close if so

arhowe00:
Closed as part of #311 .

PR #311, opened 2025-04-28
Add improvements to protocol config

arhowe00:
Contributes to #282 

## Summary (for future reference)

This PR adds setting parameter constraints and an API for bounding values. Initially, the fields of a protocol would be defined in forms created by the abstract form creator, e.g. `OpenLIFUAbstractDataclassDefinitionFormWidget`.  Here, an API is introduced in those abstract forms for modifying the specific ranges of configured values in the protocol config.  For instance, it may be desired that when configuring `SimSetup` for using k-wave to simulate wave propagation, the Axial Simulation Range (mm) (`z_extent`) be bounded between 1-200. This PR introduces a simple API for manually adjusting those bounds. 

### Using the API to set bounds

To use the API effectively, each class that wants custom bounds first:

1) Subclasses `OpenLIFUAbstractDataclassDefinitionFormWidget`, for example in the `SimSetup` example, we have `OpenLIFUSimSetupDefinitionFormWidget(OpenLIFUAbstractDataclassDefinitionFormWidget)`
2) Constructs the parent
3) Set the bounds manually

### Subclassing for custom definitions (e.g. for `ParameterConstraint`s)

The subclass-and-modify pattern is also used for making a user-friendly way of defining parameter constraints (`ParameterConstraint` objects). Since the parameter constraints on a protocol is a dictionary mapping the parameter to a `ParameterConstraint` object, we:

1) Make the following subclass: `OpenLIFUParameterConstraintsWidget(DictTableWidget)`
2) Override dialogs, functions, and widgets to have the desired, hardcoded behavior.  For instance, you'll notice that constraints are defined by choosing from a list of parameters and constraint operators. The default forms would not create this type of an interface.
image

## For review

Test out loading and unloading protocols, and then loading one, adding some parameter constraints, and then saving it.

## What remains to complete #282 

After this is merged, we need to merge @peterhollender 's [OpenLIFU-python PR#289](https://github.com/OpenwaterHealth/OpenLIFU-python/pull/289) into openlifu and then change how `SegmentationMethods` are defined in the protocol config.

### DIFF
--- a/OpenLIFULib/OpenLIFULib/util.py
+++ b/OpenLIFULib/OpenLIFULib/util.py
@@ -150,20 +150,15 @@ def create_noneditable_QStandardItem(text:str) -> qt.QStandardItem:
             return item
 
 def replace_widget(old_widget: qt.QWidget, new_widget: qt.QWidget, ui_object=None):
-    """Replace a widget by another. Meant for use in a scripted module, to replace widgets inside a layout.
+    """Replace a widget by another, supporting both standard layouts and QFormLayout.
 
     Args:
         old_widget: The widget to replace. It is assumed to be inside a layout.
         new_widget: The new widget that should replace old_widget.
         ui_object: The ui object from which to erase the replaced widget, if there is one.
-            This is referring to the `ui` attribute that is often defined in the setup of scripted
-            modules and constructed via `slicer.util.childWidgetVariables`.
-
     """
     parent = old_widget.parentWidget()
     layout = parent.layout()
-    index = layout.indexOf(old_widget)
-    layout.removeWidget(old_widget)
 
     if ui_object is not None:
         ui_attrs_to_delete = [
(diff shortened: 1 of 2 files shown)


arhowe00:
@sadhana-r This will allow you to make parameter constraints very easily in reviewing [PR#284](https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/284). You can just do the "For Review" suggestions and not do a thorough code review.  The summary is long-winded here so that it can help in the future when using the API.

sadhana-r (changes requested):
The parameter constraints widget shows up after the `Segmentation Method` collapsible and under `Solution Analysis Options`. I'm assuming its not meant to be aduplicated? Cause right now I can select the same parameter and place different constraints in each widget. 

After Segmentation Methods, there is Parameter Constraints (maybe not meant to be there) and Target constraints. Can the layout of target constraints be changed to form layout to be consistent with the rest of the UI?

arhowe00:
> I opened an issue about this weirdness in openlifu: [OpenwaterHealth/OpenLIFU-python#293](https://github.com/OpenwaterHealth/OpenLIFU-python/issues/293) When that is fixed hopefully the fix propagates automatically to the protocol config form

Yes, and if it's not automatic we can always remove these things through special logic -- the goal is to stay generic until we are sure..

sadhana-r (approved):

arhowe00:
> After Segmentation Methods, there is Parameter Constraints (maybe not meant to be there) and Target constraints. Can the layout of target constraints be changed to form layout to be consistent with the rest of the UI?

FYI: I opened an issue for AP matters in the slicer extension. See #313

PR #312, opened 2025-04-29
Add status update during photoscan generation

### LINKED ISSUE #264: Add words of patience encouragement during photoscan generation
> Photoscan generation puts the app into "Not Responding" while it runs - better background running and messaging would be helpful. "This might take 20 minutes" is a bit light when it looks like the app crashes.

sadhana-r:
Closes #264 

(@ebrahimebrahim does this close this issue or should we keep it open till we connect meshroom updates?)

- Adds dummy connections to the photoscan generation progress bar. It updates to 100% once phtoscan geenration is finished, and resets to 0 when `Start Photoscan Generation` is clicked.
- Adds a status update message that should only show up while photoscan generation is running

**For review:**
I don't have meshroom set up on my computer so wasn't able to test it out while the pipeline running. I think it should work but it would  be great if you could look over the code and test it out.

### DIFF
--- a/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
+++ b/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
@@ -1300,6 +1300,7 @@ class OpenLIFUTransducerTrackerWidget(ScriptedLoadableModuleWidget, VTKObservati
         self.ui.startPhotocollectionCaptureButton.clicked.connect(data_module.onStartPhotocollectionCaptureClicked)
         self.ui.startPhotoscanGenerationButton.clicked.connect(self.onStartPhotoscanGenerationButtonClicked)
         self.ui.importPhotocollectionFromDiskButton.clicked.connect(data_module.onImportPhotocollectionFromDiskClicked)
+        self.ui.photoscanGenerationStatusMessage.hide() # Only displayed when running photoscan generation
         # ------------------------------------------
 
         # Replace the placeholder algorithm input widget by the actual one
@@ -1444,6 +1445,10 @@ class OpenLIFUTransducerTrackerWidget(ScriptedLoadableModuleWidget, VTKObservati
             openlifu_lz().nav.photoscan.get_meshroom_pipeline_names()
         )
         if photoscan_generation_options_dialog.exec_() == qt.QDialog.Accepted:
+            self.updatePhotoscanGeneratorProgressBar(new_photoscan_generator_progress_value = 0)
+            self.ui.photoscanGenerationStatusMessage.show()
+            self.ui.photoscanGenerationStatusMessage.text = ("Please note that this process can take up to 20 minutes."
+            "The progress bar is currently not fully functional and will update only upon completion.")
             try:
                 self.logic.generate_photoscan(
                     subject_id = subject_id,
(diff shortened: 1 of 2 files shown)


OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -1444,6 +1445,10 @@ def onStartPhotoscanGenerationButtonClicked(self):
             openlifu_lz().nav.photoscan.get_meshroom_pipeline_names()
         )
         if photoscan_generation_options_dialog.exec_() == qt.QDialog.Accepted:
+            self.updatePhotoscanGeneratorProgressBar(new_photoscan_generator_progress_value = 0)
+            self.ui.photoscanGenerationStatusMessage.show()
+            self.ui.photoscanGenerationStatusMessage.text = ("Please note that this process can take up to 20 minutes."
+            "The progress bar is currently not fully functional and will update only upon completion.")
```
ebrahimebrahim, line 1451:
```suggestion
            self.ui.photoscanGenerationStatusMessage.text = ("Generating mesh... this process can take up to 20 minutes.")
```

ebrahimebrahim:
Thanks for hooking up the progress bar!

Yes this would close the issue

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -1455,11 +1459,24 @@ def onStartPhotoscanGenerationButtonClicked(self):
             except CalledProcessError as e:
                 slicer.util.errorDisplay("The underlying Meshroom process encountered an error.", "Meshroom error")
                 raise e
+            self.updatePhotoscanGeneratorProgressBar(new_photoscan_generator_progress_value = 100)
+            self.ui.photoscanGenerationStatusMessage.hide()
```
ebrahimebrahim, line 1463: this hide is not reached if we hit the `except` just above. I think you can fix this by putting these UI update lines inside a `finally` after the `try` and `except`

ebrahimebrahim (changes requested):
The message did show up when I was running meshroom, but it is kind of hard to see I felt:

image


Not sure if there's an easy solution to this. The UI is just really busy. Maybe it should be an info dialog that you have to "Ok" through? But then it's an extra click.
I think maybe it's fine as is unless you can think of something. Making it bold or italic feels tacky. Maybe once the progress bar starts moving in a later change the attention will be drawn to that area and it will be readable

---

Okay but here is why a change is requested: the message is hidden only if mesh generation succeeds. If there is an error, then the message gets stuck there, visible

sadhana-r:
> Not sure if there's an easy solution to this. The UI is just really busy. Maybe it should be an info dialog that you have to "Ok" through? But then it's an extra click. I think maybe it's fine as is unless you can think of something. Making it bold or italic feels tacky. Maybe once the progress bar starts moving in a later change the attention will be drawn to that area and it will be readable
> 

I think for now we can leave it as is or I could change the text to display in red, similar to the warning messages? Since we plan to update this later to make the progress bar functional/provide meshroom updates, it will become more noticeable. Too many info boxes could get annoying.

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -1448,6 +1448,7 @@ def onStartPhotoscanGenerationButtonClicked(self):
             self.updatePhotoscanGeneratorProgressBar(new_photoscan_generator_progress_value = 0)
             self.ui.photoscanGenerationStatusMessage.show()
             self.ui.photoscanGenerationStatusMessage.text = ("Generating mesh... this process can take up to 20 minutes.")
+            self.ui.photoscanGenerationStatusMessage.styleSheet = "color:red;"
```
ebrahimebrahim, line 1451: the red is much easier to see, thanks
ugly, but we are okay with ugly for now :)

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -1456,11 +1457,12 @@ def onStartPhotoscanGenerationButtonClicked(self):
                     meshroom_pipeline = photoscan_generation_options_dialog.get_selected_meshroom_pipeline(),
                     image_width = photoscan_generation_options_dialog.get_entered_image_width(),
                 )
+                self.updatePhotoscanGeneratorProgressBar(new_photoscan_generator_progress_value = 100)
             except CalledProcessError as e:
                 slicer.util.errorDisplay("The underlying Meshroom process encountered an error.", "Meshroom error")
                 raise e
-            self.updatePhotoscanGeneratorProgressBar(new_photoscan_generator_progress_value = 100)
-            self.ui.photoscanGenerationStatusMessage.hide()
+            finally:
```
ebrahimebrahim, line 1464: Tested a case where it fails and a case where it succeeds -- indeed this worked, ty!

ebrahimebrahim (approved):

PR #314, opened 2025-04-30
Add management and enforcement of protocol specific user permissions

### LINKED ISSUE #262: Add management and enforcement of protocol-specific user permissions
Once https://github.com/OpenwaterHealth/OpenLIFU-python/issues/239 is done, we will have some scheme for saying which user role or which user has access to which protocols. Whatever scheme we come up with, we should add (1) a method for managing that in the SlicerOpenLIFU user management interface and (2) enforcement of that information based on who is the currently logged in user. So whether it's visibility of protocols that we decide should be enforced, or whether it's the ability to use them to initiate a session, or whether it's both of those things, we should enforce that.

There is also (3): a method for managing this in the protocol configuration module when creating or editing protocols. However that's not part of this issue here as it's to be included in a future milestone.

Depends on https://github.com/OpenwaterHealth/OpenLIFU-python/issues/239

arhowe00:
Closes #262 

This PR adds:

- Adding user roles in the user management console (note: you can't set these when creating a new user; you just choose between operator and admin still)
- Allowing allowed roles to be configured on protocols
- Using the user roles of the currently logged in user to determine which protocols can be seen at any given time

This also implements the important considerations in #262, such as:
- Loaded data are removed when changing users

Note, the following item remains to be addressed at a future date:

[@ebrahimebrahim said](https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/262#issuecomment-2828112194):
> Agree that triggering a scene close (using the existing functions that do that) on change of user makes sense. But there should be a warning asking for confirmation, when someone tries to switch user while there are loaded openlifu objects around, so they don't unknowingly lose their work

Currently, the data are wiped when changing users, and there is no warning that it's about to happen.

## For review

I already tested how the protocols are selectively loaded based on the user roles and making sure that data are wiped when changing users. For review, I'd say to do the same thing I did: add some custom role names, make a new protocol with some of those custom role names, and see that some users can access it but others can't.  Admins should be able to access everything at all times.

## Before merge

- [x] Make sure to edit the commit which updates python-requirements to reference a commit merged into the main branch of openlifu from the [PR implementing the fix to non-serializable Protocols](https://github.com/OpenwaterHealth/OpenLIFU-python/pull/295).


OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py.orig
sadhana-r, line 1: This file should be removed from the commit

OpenLIFULogin/OpenLIFULogin.py
```diff
@@ -245,6 +246,101 @@ def customexec_(self):
...
+        """
+        create_password_text = self.createPasswordField.text
+        confirm_password_text = self.confirmPasswordField.text
+
+        if not create_password_text or len(create_password_text) < 6:
+            slicer.util.errorDisplay("Password must be at least 6 characters.", parent=self)
+            return
+        if create_password_text != confirm_password_text:
+            slicer.util.errorDisplay("Passwords do not match.", parent=self)
+            return
```
sadhana-r, line 323: Not important to address now but for consistency, I think it would be good to include this double confirm when creating a new user and setting a password in the beginning.

sadhana-r (changes requested):
Great work. I think everything works as expected! I created a new user with a custom role and created a new protocol with that custom role. And then checked that another user role type (except admin) is not able to load that protocol. And only admin is allowed to create new protocols. 

Requesting changes to remove the misc. file.

OpenLIFULogin/OpenLIFULogin.py
arhowe00: I'll append this to our AP issue

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py.orig
arhowe00: Good catch

arhowe00:
[Great, before merging I'm going to wait for this to get merged](https://github.com/OpenwaterHealth/OpenLIFU-python/pull/295) and then rebase the `python-requirements.txt`

sadhana-r (approved):
Going to approve so you can rebase and merge once the python requirements are updated.

PR #315, opened 2025-05-01
Impose solution analysis safety limits

### LINKED ISSUE #261: Impose solution analysis safety limits
A solution that does not satisfy the constraints defined in OpenLIFU-python is automatically not approved and cannot be approved.

So this is two things:
- Controlling the enabledness of the solution approval button: it is only enabled if there exists a solution analysis that passes the checks.
- Automatically revoking any solution approval in the event that the solution analysis becomes invalidated (i.e. stops exist or stops passing the checks)

But there's a bit more to it to guarantee safety:
- The solution analysis needs to be recomputed and updated whenever the inputs that were used to generate it change. Either that, or it needs to be invalidated if the inputs used to generate it change. So I think that means just if the transducer is moved or if the target is moved.

The saving and loading of solution approval, and its effect on workflow, will be handled in #260 and not here. This issue is about the relationship between solution analysis and solution approval.

Depends on #98.
Depends on [P#230](https://github.com/OpenwaterHealth/OpenLIFU-python/issues/230).

arhowe00:
Closes #261 

This PR implements safety protections for sonication solutions and sonication solution analyses.

- Delete solution and solution analysis when sonication target or transducer transform is modified
- Update solution approval logic based on solution analysis validity (exists & not in error state)
- Revoke solution approval when the solution analysis is modified (cannot currently occur, but logically important)

## For Review

Make sure to check the parts where I inject `OpenLIFUSonicationPlanner` related callbacks and invocations from external modules. For instance, the callback for the sonication target being modified was _not_ registered, it was called directly in `OpenLIFUPrePlanning/OpenLIFUPrePlanning.py`. I did a similar thing in `OpenLIFUData/OpenLIFUData.py` for the transducer, except the callback registration is more explicit through `observe_transform_modified` in `load_transducer_from_openlifu`.

If you want, you can also attempt loading the example session, computing the solution, approving it, and moving a target, but I already checked that it works.

## Before merge

- [ ] Make sure [SlicerOpenLIFU PR#284](https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/284) is merged to main and then rebase this branch to main before merging.

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1906,6 +1906,8 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
     # TODO: This should be a widget level function
     def _on_transducer_transform_modified(self, transducer: SlicerOpenLIFUTransducer) -> None:
 
+        slicer.util.getModuleWidget('OpenLIFUSonicationPlanner').onTransducerTransformModified
+
         matching_transform_id = transducer.transform_node.GetAttribute("matching_transform")
         if matching_transform_id:
             # If its a transducer tracking node, revoke approval if approved
@@ -2034,7 +2036,9 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
             transducer_matrix_units=transducer_matrix_units,
         )
         self.getParameterNode().loaded_transducers[transducer.id] = newly_loaded_transducer
+
         newly_loaded_transducer.observe_transform_modified(self._on_transducer_transform_modified)
+
         return newly_loaded_transducer
 
     def remove_transducer(self, transducer_id:str, clean_up_scene:bool = True) -> None:
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -68,6 +68,8 @@ if TYPE_CHECKING:
     import openlifu
     import openlifu.nav.photoscan
     from OpenLIFUPrePlanning.OpenLIFUPrePlanning import OpenLIFUPrePlanningWidget
+    from OpenLIFUSonicationPlanner.OpenLIFUSonicationPlanner import OpenLIFUSonicationPlannerLogic
+
 #
 # OpenLIFUData
 #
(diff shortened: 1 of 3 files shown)


OpenLIFUData/OpenLIFUData.py
```diff
@@ -2006,6 +2006,10 @@ def load_transducer_from_openlifu(
             transducer_matrix_units=transducer_matrix_units,
         )
         self.getParameterNode().loaded_transducers[transducer.id] = newly_loaded_transducer
+
+        # ---- Add observers ----
+        newly_loaded_transducer.observe_transform_modified(slicer.util.getModuleWidget('OpenLIFUSonicationPlanner').onTransducerTransformModified)
```
sadhana-r, line 2011: This will not be needed after you rebase. I have added an observer to `load_transducer_from_openlifu`

OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
```diff
@@ -309,6 +309,26 @@ def onPointAddedOrRemoved(self, caller, event):
...
+            data_logic.clear_solution(clean_up_scene=False)
+            self._parameterNode.solution_analysis = None
+            slicer.util.infoDisplay(
+                text= "Computed solution has been deleted due to moving the target.",
+                windowTitle="Solution deleted"
+            )
+
+    def onTransducerTransformModified(self, transducer):
+        data_logic : "OpenLIFUDataLogic" = slicer.util.getModuleLogic('OpenLIFUData')
+        if self.logic.solution_analysis_exists():
```
sadhana-r, line 324: After you rebase to the current main, this function can be deleted and the code can be moved to ` _on_transducer_transform_modified` in the data module.

sadhana-r (changes requested):

OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
```diff
@@ -309,6 +309,26 @@ def onPointAddedOrRemoved(self, caller, event):
...
 
+    def onTargetModified(self):
+        data_logic : "OpenLIFUDataLogic" = slicer.util.getModuleLogic('OpenLIFUData')
+        if self.logic.solution_analysis_exists():
+            data_logic.clear_solution(clean_up_scene=False)
+            self._parameterNode.solution_analysis = None
+            slicer.util.infoDisplay(
+                text= "Computed solution has been deleted due to moving the target.",
+                windowTitle="Solution deleted"
+            )
```
sadhana-r, line 320: You could create a function called `clearSolutionAnalysis` similar to 
https://github.com/OpenwaterHealth/SlicerOpenLIFU/blob/3c4af3901add0fbc9ba6683d8af8b92ba8a0e558/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py#L1721
and
https://github.com/OpenwaterHealth/SlicerOpenLIFU/blob/3c4af3901add0fbc9ba6683d8af8b92ba8a0e558/OpenLIFUPrePlanning/OpenLIFUPrePlanning.py#L339
that takes a 'reason' as input and triggers the info box and calls a logic function to clear the solution. That way you don't need to repeat the code when the target or transform is modified.

arhowe00: Added!

sadhana-r (approved):

OpenLIFUPrePlanning/OpenLIFUPrePlanning.py
```diff
@@ -315,6 +316,7 @@ def onPointModified(self, node:vtkMRMLMarkupsFiducialNode, caller, event):
             reason = "The target was modified."
             self.revokeApprovalIfAny(node, reason=reason)
             self.clearVirtualFitResultsIfAny(node, reason = reason)
+            slicer.util.getModuleWidget('OpenLIFUSonicationPlanner').onTargetModified()
```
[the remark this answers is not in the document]
sadhana-r: Hmm, I guess I like the suggested approach since it makes it easier to see in one place all of the actions that get triggered by either a target or transducer modification, while maintaining the logic related to approvals/deletion within the relevant module. But I'll leave the final decision up to you!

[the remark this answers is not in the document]
arhowe00: They are both pretty convincing, but I think I am won over by your solution due to the `reason` being naturally unified in that place =)

PR #319, opened 2025-05-02
Prevent unallowed protocols being loaded through creating or loading sessions

### LINKED ISSUE #262: Add management and enforcement of protocol-specific user permissions

arhowe00:
Closes #262 

#262 was reopened due to some [incorrect behavior](https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/262#issuecomment-2847781778) around enforcement of allowed protocols given the user roles.

This PR prevents unallowed protocols from being shown in the session creator, and attempts to load sessions with unallowed protocols will fail. I tested this example by adding "operator" to the allowed roles of the example session and using the example restricted user and example operator to compare. Things worked as expected.

## For review

You can try modifying the allowed roles on some protocols programmatically or in the JSON files (e.g. add "operator" to a protocol's list). Then, using the accounts below or a custom account, you can test that you can't see any unallowed protocols.

**### Credentials:**
```
complex_user/complex_user_hash
example_admin/example_admin_hash
example_operator/example_operator_hash
example_restricted/example_restricted_hash
```

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -49,7 +49,7 @@ from OpenLIFULib.transducer_tracking_results import (
     get_photoscan_id_from_transducer_tracking_result,
     is_transducer_tracking_result_node,
 )
-from OpenLIFULib.user_account_mode_util import UserAccountBanner
+from OpenLIFULib.user_account_mode_util import get_current_user, get_user_account_mode_state, UserAccountBanner
 from OpenLIFULib.util import (
     BusyCursor,
     create_noneditable_QStandardItem,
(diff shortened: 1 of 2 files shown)


ebrahimebrahim (approved):
Thank you for the fix! Also I agree with the simplification to the code where you stopped filtering dropdowns unnecessarily in the protocol config module

PR #321, opened 2025-05-02
Improve layout and scaling of transducer tracking wizard 

### LINKED ISSUE #269: Better scale the transducer tracking wizard
> the scaling on the tracking wizard is wonky, leaving blank space below the landmark table instead of just making the rendering window larger

Currently the view is 50/50 vertical space taken by the 3D view and the controls.

But the controls don't need all that much space, so it should be more like 70/30 or 80/20.

sadhana-r:
Closes #269 

### For review

Checkout the branch and see if the layout out behaves and looks as expected when you adjust the size of the wizard window on the different pages. The layout should prioritize maximizing the 3D view widget over the control widgets. (You can also take a quick look over the code).

### DIFF
--- a/OpenLIFULib/OpenLIFULib/transducer_tracking_wizard_utils.py
+++ b/OpenLIFULib/OpenLIFULib/transducer_tracking_wizard_utils.py
@@ -16,14 +16,20 @@ def initialize_wizard_ui(wizard: qt.QWizard):
     return slicer.util.childWidgetVariables(uiWidget)
 
 def set_threeD_view_widget(ui):
-     
+
     viewWidget = slicer.qMRMLThreeDWidget()
     viewWidget.setMRMLScene(slicer.mrmlScene)
     viewWidget.setMinimumHeight(200)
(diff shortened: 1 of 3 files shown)


arhowe00:
Hey, what do you want me to review for this?

sadhana-r:
> Hey, what do you want me to review for this?

I updated the PR description

arhowe00 (approved):
Checked it out and the scaling looks great! I am approving because I don't know if below is a bug, but in the tt wizard, after placing the landmarks I don't seem to be able to continue. Again, I wasn't asked to test this so I'll assume it's something wrong on my end.
image

ebrahimebrahim:
(it's not supposed to let you continue until you approve -- that is intended behavior)

arhowe00:
> (it's not supposed to let you continue until you approve -- that is intended behavior)

Oh right, I didn't even start with photoscan-volume stuff. I went right to "Run transducer tracking".

sadhana-r:
I actually found a bug in the workflow, which I think is what you are referring to. If there is an existing result and then you revoke approval and place new landmarks, the `Initialize transform` button does not get enabled. I have fixed this here: e03dac617b0e87ae0ba4dc6899f75f27cc340eb9

PR #328, opened 2025-05-05
Move protocol config status

### LINKED ISSUE #254: protocol config status view usually not visible
Now that the protocol config module has become a large form that you have to scroll through, the screaming status message "You have unsaved changes!" is almost always out of view. Probably the status message should be pulled out of the main form scrollable area widget so that it's always visible

arhowe00:
Closes #254 

This PR encloses protocol configurations in a separate section and uses the `ctkDynamicSpacer` to always show the status message along with other protocol utilities (e.g. "Save Protocol to Database")

## For review

Code review alone should be enough, but you can also  pull this branch and check out whether the protocol configuration module looks intuitive.

### DIFF
--- a/OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
+++ b/OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
@@ -426,10 +426,12 @@ class OpenLIFUProtocolConfigWidget(ScriptedLoadableModuleWidget, VTKObservationM
         if protocol.id in self.logic.cached_protocols:
             cached_protocol = self.logic.cached_protocols[protocol.id]
             self.updateProtocolDisplayFromProtocol(cached_protocol)
+            self.ui.scrollArea.verticalScrollBar().setValue(0)
             self.setProtocolEditorEnabled(True)
             self.updateWidgetSaveState(SaveState.UNSAVED_CHANGES)
         else:
             self.updateProtocolDisplayFromProtocol(protocol)
+            self.ui.scrollArea.verticalScrollBar().setValue(0)
             self.setProtocolEditorEnabled(False)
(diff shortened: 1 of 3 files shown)


sadhana-r (changes requested):
The enclosure within the scrolling window looks good and makes the module much easier to work with. It would be nice if when you change the active protocol i.e. either by loading a different protocol, creating a new one, changing the actively selected protocol, you also reset the scroll bar so that its at the top of the window. It will make things a little easier instead of having to scroll back to the top.

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -426,10 +426,12 @@ def onProtocolSelectorIndexChanged(self):
         if protocol.id in self.logic.cached_protocols:
             cached_protocol = self.logic.cached_protocols[protocol.id]
             self.updateProtocolDisplayFromProtocol(cached_protocol)
+            self.ui.scrollArea.verticalScrollBar().setValue(0)
```
sadhana-r, line 429: You also just plug this line in at the end of the `updateProtocolDisplayFromProtocol` function since its called whenever the display is updated. Then you won't need to add this line multiple times.

sadhana-r (approved):

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
arhowe00: In fact, the way I looked for where to put that line was everywhere `updateProtocolDisplayFromProtocol` is put, but I did not add it into the function in case there is a reason we would want to update the display but not move the scroll bar.

PR #329, opened 2025-05-05
Update workflow controls when adding target (#252)

### LINKED ISSUE #252: Preplanning workflow status message does not update when a target is added
When you have no targets, after a target is added the status message in the preplanning module workflow controls still tells you that you need to add a target to proceed. It should instead tell you that you need to do virtual fit.

arhowe00:
Closes #252 

Adding a sonication target now updates the workflow controls to prompt for running a virtual fit.

## For review

Code review should be enough -- I only modified the `onPointAddedOrRemoved` to update workflow controls. I would like the reviewer's opinion on my **only** adding the update there (for instance, I did not add it to `onPointModified`). This is because while we could issue updates on every `on..` routine call, it may make the code less maintainable because it's probably redundant (would the workflow controls ever update due to a target modification)?


sadhana-r (approved):
Looks good! I played around with this module and I there isn't a situation other than `PointAddedOrRemoved` that the `updateWokflowControls` call needs to be added.

PR #332, opened 2025-05-07
Validate protocols in the protocol config module

### LINKED ISSUE #324: Validate protocols in the protocol config module
In the protocol configuration module, a newly created protocol should run through a validation algorithm before the protocol can be published to the database via an instantiation validation in a Protocol's ` _post_init_` method, so that an error will be shown if an instantiated Protocol uses "out of bounds" attribute values.

arhowe00:
Closes #324 

Newly created protocols run through validation algorithms defined in the `__post_init__`  method, both for the top-level `Protocol` class and all sub-objects. This PR officially handles thrown exceptions (i.e. for "out of bounds" attribute values) by displaying them to the researcher and failing to publish the Protocol to the database.

Caches skip the validation through an introduced utility `instantiate_without_post_init` that does not call `__init__`.

## For review

I already tested this by configuring protocols with values that cause `__post_init__` to fail ("erroneous values"). Caching protocols does not fail (via the utility with `post_init=False`). Then, when any "save" routine is performed, an explanation for a failed save is given via `errorDisplay`.

If you would like to test this, you can alternate between edited protocols with erroneous values. For instance, a `Sequence` with the following settings will fail `__post_init__`:
```
pulse_interval=1.0
pulse_count=4
pulse_train_interval=1.0
pulse_train_count=1
```

### DIFF
--- a/OpenLIFULib/OpenLIFULib/class_definition_widgets.py
+++ b/OpenLIFULib/OpenLIFULib/class_definition_widgets.py
@@ -940,9 +940,9 @@ class CreateAbstractClassDialog(qt.QDialog):
         top_level_layout.addWidget(self.buttonBox)
 
         self.buttonBox.rejected.connect(self.reject)
-        self.buttonBox.accepted.connect(self.objectidateInputs)
+        self.buttonBox.accepted.connect(self.validateInputs)
 
-    def objectidateInputs(self):
+    def validateInputs(self):
         """
         Ensure object is valid
         """
--- a/OpenLIFULib/OpenLIFULib/class_definition_widgets.py
+++ b/OpenLIFULib/OpenLIFULib/class_definition_widgets.py
@@ -1,6 +1,6 @@
 # Standard library imports
 import inspect
-from dataclasses import fields, is_dataclass
+from dataclasses import fields, is_dataclass, MISSING
 from typing import (
     Annotated,
     Any,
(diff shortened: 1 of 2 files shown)


sadhana-r (dismissed):
I tested creating a new protocol, and saving a protocol with the Sequence you suggested. I get the error message when I try to save to file or the database as expected.

I think you plan to do this in a separate PR, addressing some of the UI changes that Peter suggested but when there are multiple post-init exceptions i.e. negative pulse interval and negative pulse count, the error message only shows the first exception. Does openlifu return all of the errors for us to show all the protocol validation errors?

Otherwise, I looked over the code and it looks good to me. The distinction between instantiating with and without post-init makes sense. 

Not related to this PR but I found some small bugs. When I create a new protocol, enter some settings/descriptions and then create a second new protocol, without saving the first one, the first protocol does not get cached. It gets cached if I create a protocol and then load a protocol from the database. 

Also, if I change the name of a protocol, and then create a new protocol or switch to a different protocol in the drop down,  the protocol gets cached when I change the drop down selection, but the dropdown list doesn't update to reflect the new protocol name. It only updates when I load a protocol. Ideally, creating a new protocol/loading a protocol or changing the selection should trigger the same UI updates/caching behavior?

arhowe00:
> I think you plan to do this in a separate PR, addressing some of the UI changes that Peter suggested but when there are multiple post-init exceptions i.e. negative pulse interval and negative pulse count, the error message only shows the first exception. Does openlifu return all of the errors for us to show all the protocol validation errors?

This is a good point. It is infeasible to check on a granular basis which definitions throw exceptions. I think that the   . One way you can see **all the issues at once** is to collect errors and raise them together, e.g.:

```python
def __post_init__(self):
    errors = []

    if self.pulse_interval <= 0:
        errors.append("Pulse interval must be positive")
# ...
    if self.pulse_train_count <= 0:
        errors.append("Pulse train count must be positive")

    if errors:
        raise ValueError("Validation errors:\n" + "\n".join(errors))
```

Based on the discussion we had with Peter, however, this will not be necessary because as soon as we know that one field is erroneous we can [report this in the requested validity indicator](https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/334#issue-3049993727). Peter suggested not having class definition-specific warnings and instead just report at the bottom where the Protocol was invalid.

I'm going to open an issue regarding the protocol config errors you noted as a bug report, and when triaging I'll see if I can reproduce the sequence of steps. In the second case ("Also, if I change [...] updates/caching behavior?"), I think in this case we, as developers, need to make a choice on what we cache and show the user and showing the old protocol name and ID made sense because this could result in an ID name being the same as another as ID checks are not made until the protocol is officially saved (i.e., we confirm that the ID they finally typed in is unique in the database)

ebrahimebrahim (approved):
Approving and merging as the two issues pointed out by Sadhana are acceptable:
- Showing only one exception message is the intended and expected behavior
- The protocol caching bugs are tracked in in a separate issue

PR #331, opened 2025-05-06
Add Photoscan-MRI ICP registration

sadhana-r:
Closes #326 

This PR updates the UI to include a 'Run ICP registration' button and includes the logic for performing ICP when the button is clicked. The ICP result fine-tunes the fiducial registration  result. ICP is run between the  full photoscan and skin segmentation mesh.

Running ICP involves interpolating additional control points on the photoscan surface using the 3 manually placed facial landmarks, using the interpolated points to extract a submesh of the face from the photoscan model, and running ICP using the submesh. 

There was a bug in the order in which transforms were being composed when applied to the photoscan mesh i.e. when composing the fiducial transform result with the scaling transform. This PR addresses that issue. 

### For review

- Test out the ICP button (only on the photoscan to volume tracking page) and check that the ICP runs and updates the photoscan mesh as expected.
- You should be able to manually edit the transform after fiducial registration or ICP
- A look over the code would also be help. In particular, error handling when calling the ICP registration logic. 
-(Feel free to play around with the wizard a little and make sure nothing was broken by the changes in this PR).

Types of isses to look out for: new wizard-related nodes visible in the slicer scene while the wizard is running, any temp wizard nodes remaining in the scene after exiting the wizard.

PR #336, opened 2025-05-11
Add transducer-photoscan and photoscan-MRI ICP registration

### LINKED ISSUE #325: Transducer-photoscan ICP registration
Implement ICP registration and make it happen when the ICP button is clicked on the transducer-photoscan transducer tracking wizard page.

If the registration doesn't immediately work well on the example, then add a workflow for subsetting the photoscan mesh to an ROI that can be used for registration.

Depends on https://github.com/OpenwaterHealth/OpenLIFU-python/issues/299

sadhana-r:
Closes #325

Should be merged after #331. 

### For Review

This should be a quick review since it builds on the ICP logic introduced in #331. You can look over the code and test out ICP registration between the photoscan and transducer (final page of wizard).

### DIFF
--- a/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
+++ b/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
@@ -247,6 +247,7 @@ class FacialLandmarksMarkupPageBase(qt.QWizardPage):
         # invalidates the result and resets previously initialized transform nodes
         if self.wizard()._valid_tt_result_exists:
             self.wizard()._valid_tt_result_exists = False
+            self.wizard().photoscanVolumeTrackingPage.resetScalingTransform()
             self.wizard().photoscanVolumeTrackingPage.photoscan_to_volume_transform_node = None
             self.wizard().transducerPhotoscanTrackingPage.transducer_to_volume_transform_node = None
 
@@ -338,7 +339,6 @@ class PhotoscanMarkupPage(FacialLandmarksMarkupPageBase):  # Inherit from the ba
         if self.facial_landmarks_fiducial_node is None:
             self._initialize_facial_landmarks_fiducial_node(node_name = "photoscan-wizard-faciallandmarks")
             self.setupMarkupsWidget()
-            # self.wizard()._valid_tt_result_exists = False # TODO: Do I need this here?
 
         if self.ui.placeLandmarksButton.text == "Place/Edit Registration Landmarks":
             self.facial_landmarks_fiducial_node.SetLocked(False)
@@ -474,6 +474,7 @@ class PhotoscanVolumeTrackingPage(qt.QWizardPage):
 
         self.ui.approveTransformButton.clicked.connect(self.onTransformApproveClicked)
         self.ui.enableManualPVRegistration.clicked.connect(self.onManualRegistrationClicked)
+        self.ui.runICPRegistrationPV.clicked.connect(self.onRunICPRegistrationClicked)
         self.ui.initializePVRegistration.clicked.connect(self.onInitializeRegistrationClicked)
         self.runningRegistration = False
 
@@ -487,11 +488,10 @@ class PhotoscanVolumeTrackingPage(qt.QWizardPage):
         self.ui.scalingTransformMRMLSliderWidget.pageStep = 1.0
         self.ui.scalingTransformMRMLSliderWidget.setToolTip(_('Adjust the scale of the photosan mesh."'))
         self.ui.scalingTransformMRMLSliderWidget.connect("valueChanged(double)", self.updateScaledTransformNode)
-        self.ui.scalingTransformWidget.hide()
 
-        self.scaledTransformNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLTransformNode")
-        self.scaledTransformNode.SetName("wizard_photoscan_volume-scaled")
         self.photoscan_to_volume_transform_node: vtkMRMLTransformNode = None
+        self.scaling_transform_node = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLTransformNode")
+        self.scaling_transform_node.SetName("wizard_photoscan_volume-scaling_factor")
     
     def initializePage(self):
         """ This function is called when the user clicks 'Next'."""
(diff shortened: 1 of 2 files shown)


arhowe00 (dismissed):
I tested out the ICP registration, and the transducer doesn't appear to be perfectly aligned with the photoscan. However, I don't doubt that I could have made mistakes in the process of aligning this, I just clicked through the wizard. Since this is to be merged after #331 is this behavior to be expected? I jumped on this review because it was requested recently but maybe it is too soon for this feature to work:
image

Using the handles is nice, but I am not able to see the photoscan very well in demo_transducertracking. Is there a different subject/session you used in development that would be easier to use?
image

ebrahimebrahim:
Thanks for reviewing

> Is there a different subject/session you used in development that would be easier to use?

Yes, the Soren example is the one to try here.

> but maybe it is too soon for this feature to work:

No, it's supposed to work! However it is understood that transducer alignment only works if the initialization is close enough -- so it's okay if it's failing when the initialization is poor.

ebrahimebrahim:
Tests were very successful! For both photoscan-MRI and transducer-photoscan ICP, on the soren example.

Indeed one does have to give the transducer a good initialization for the ICP to work on the transducer.

The transducer registration surface could perhaps be improved here -- currently it's just a rectangle and it doesn't capture the sharp curvature at the edges of the outside surface of the device, which could be helpful for ICP.

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -2121,6 +2212,156 @@ def run_fiducial_registration(self,
...
+            if transformType == 2:
+                icpTransform.GetLandmarkTransform().SetModeToAffine()
+            icpTransform.SetMaximumNumberOfIterations( numIterations )
+            icpTransform.SetMaximumNumberOfLandmarks( numLandmarks )
+            icpTransform.Modified()
+            icpTransform.Update()
+
+            icp_result_node.SetMatrixTransformToParent( icpTransform.GetMatrix() )
+
+            if slicer.app.majorVersion >= 5 or (slicer.app.majorVersion >= 4 and slicer.app.minorVersion >= 11):
```
ebrahimebrahim, line 2354: Woah, I think it's safe to assume we are on a least version 5 :sweat_smile: 

sooo many things would probably break if we tried any of this with Slicer 4...

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -2121,6 +2212,156 @@ def run_fiducial_registration(self,
...
+                icp_result_node.SetNodeReferenceID(slicer.vtkMRMLTransformNode.GetMovingNodeReferenceRole(), input_moving_model.GetID())
+                icp_result_node.SetNodeReferenceID(slicer.vtkMRMLTransformNode.GetFixedNodeReferenceRole(), input_fixed_model.GetID())
+
+        except Exception as e:
+            print("Exception thrown:", e)
+            return None
```
ebrahimebrahim, line 2360: Arbitrary exceptions get swallowed up here and reduced to a print statement, but then in places where `run_icp_model_registration` is used we do not check if `None` was returned and do error handling.

I think it would be better to allow the exception to be raised, or re-raise it here. And ensure that if ICP fails then there is a reasonable user-facing outcome, not just a traceback or a print in the console. Could be as simple as an error dialog saying "ICP failed, check the logs."

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -815,31 +878,56 @@ def setupTransformNode(self):
...
+        photoscan_hardened.HardenTransform()
+
+        # Clone transducer surface model node and harden tansform after virtual fit initialization
+        transducer_hardened = get_cloned_node(self.wizard().transducer_surface)
+        transducer_hardened.SetAndObserveTransformNodeID(self.transducer_to_volume_transform_node.GetID())
+        transducer_hardened.HardenTransform()
+ 
+        self.transducer_to_photoscan_icp_transform_node = self.wizard()._logic.run_icp_model_registration(
+            input_fixed_model = photoscan_hardened,
+            input_moving_model = transducer_hardened)
```
ebrahimebrahim, line 920: Should the transducer actually be registered using a _rigid_ transform (`transformType=0` here)?

Because the CAD model of the transducer comes with units and we use those to place it in the Slicer RAS space with mm as the units.

ebrahimebrahim (changes requested):
A couple of small changes requested:

- Error handling in case ICP fails
- Consider whether transducer transform should be rigid

I looked over a lot of code, but not all of it -- it looks very good!

(Let me know if there is anything special you wanted to make sure I look at.)

By the way, this is a combined review of both #331 (which I closed without merge) and the additional things here.

And heads up -- I rebased this to main. So make sure to pull and reset before resuming on this branch.

Really wonderful work on this ICP integration!

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
ebrahimebrahim: (Update: Since they have asked for an early dev version of this to try out, I am going to make the change to `transformType=0` because I'm 80% sure that's what it should be -- but we can keep discussing when you get to this review)

sadhana-r: Sounds good. You're right, the transducer registration should be a rigid transform. Going back to the point Peter brought up in our last meeting, since the photoscan is scaled to the skin segmentation, the photoscan transducer scale may be slightly different from the transducer scale so using a rigid transform here may not give the best fit right now, but it is the correct choice. 
For future reference, I think we need to maybe do a similarity transform between the photoscan and transducer to get the photoscan scaled correctly, and then rigid transform between the scaled photoscan and skin segmentation. Something like that..

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
sadhana-r: True 😅 I copied whatever was in the logic from the Model Registration module but I'll remove this.

sadhana-r:
> (Let me know if there is anything special you wanted to make sure I look at.)

Thanks for the review! The error handling was the main thing I wanted someone to look over so this was helpful! I have updated the exception handling now.

I also re-organized the commits introducing icp registration and squashed a couple commits so none of the commits introduce the Model Registration dependency.

ebrahimebrahim (approved):
Error handling looks good now!

> For future reference, I think we need to maybe do a similarity transform between the photoscan and transducer to get the photoscan scaled correctly, and then rigid transform between the scaled photoscan and skin segmentation. Something like that..

Agreed. Maybe an automated fixup step at the end that similarity-ICP-registers photoscan to transducer and then rigid-ICP-registers photoscan ROI to MRI again, and then finally again rigid-ICP-registers transducer to photoscan.

Or...

The MRI and the transducer both provide information that should help determine the scale of the photoscan. I wonder if we can come up with a way to simultaneously use both of them to optimally scale the photoscan?

Anyway, we can talk about that later

Merging!

PR #346, opened 2025-05-13
Harmonize TT wizard approval workflow

### LINKED ISSUE #322: Harmonize transducer tracking wizard UI locking and approvals experience
(Add description once we figure out exactly how we want to change it)

sadhana-r:
Closes #322 

- Removes approval toggling from the photoscan preview wizard
- Adds a lock button to each page of the wizard which acts as an approval button. The final tracking result is saved and approved when the user clicks `Approve` on the last page. This also approves the associated photoscan
- Updates session-workflow to allows for multiple approved photoscans per session
-  Removes warning label when modifying previous tracking result

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1905,30 +1905,27 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
             self.remove_photocollection(photocollection_reference_number) 
         self.update_photocollections_affiliated_with_loaded_session()
 
-        # Check if there are any photoscans with approved transducer tracking results
-        # TODO: This assumes a single approved photoscan per session. This check needs to be udpated to allow for 
-        # approved_photoscan_id containing multiple photoscan IDs. 
-        approved_photoscan_id = self.getParameterNode().loaded_session.get_transducer_tracking_approvals()
-        if approved_photoscan_id:
-            approved_photoscan_id = approved_photoscan_id[0]
-            approved_tracking_result = None
+        # If there are any *approved* transducer tracking results that we have just loaded in newly_added_tt_result_nodes,
+        # then we check to see if any of them match the current transducer transform in terms of matrix values.
+        # If there is a match in matrix values, then the first such matching TT result that we encounter in the loop is 
+        # "officially" linked to the current transform by setting the "matching_transform" attribute, thereby ensuring that
+        # TT approval is revoked if the transducer is moved.
+        # Additionally, any other transducer tracking results whose matrix does not match current transducer get their approval revoked.
+        transducer_tracking_widget = slicer.modules.OpenLIFUTransducerTrackerWidget
+        approved_photoscan_ids = self.getParameterNode().loaded_session.get_transducer_tracking_approvals()
+        # approved_photoscan_ids is a list of photoscan IDs for which there is an approved TT result in the openlifu session
+        for approved_photoscan_id in approved_photoscan_ids:
 
             for transducer_to_volume_node, _ in newly_added_tt_result_nodes:
                 current_photoscan_id = get_photoscan_id_from_transducer_tracking_result(transducer_to_volume_node)
                 if current_photoscan_id == approved_photoscan_id:
                     if newly_loaded_transducer.is_matching_transform(transducer_to_volume_node):
-                        approved_tracking_result = transducer_to_volume_node
-                        break  # Found a match, no need to continue searching
-
-            if approved_tracking_result:
-                newly_loaded_transducer.set_current_transform_to_match_transform_node(approved_tracking_result)
-            else:
-                # revoke transducer tracking approval
-                transducer_tracking_widget = slicer.modules.OpenLIFUTransducerTrackerWidget
-                transducer_tracking_widget.revokeTransducerTrackingApprovalIfAny(
-                    photoscan_id=approved_photoscan_id,
-                    reason="The transducer transform does not match the approved tracking result."
-                )
+                        newly_loaded_transducer.set_current_transform_to_match_transform_node(transducer_to_volume_node)
+                    else:
+                        transducer_tracking_widget.revokeTransducerTrackingApprovalIfAny(
+                            photoscan_id=approved_photoscan_id,
+                            reason="The transducer transform does not match the approved tracking result."
+                        )
 
         self.session_loading_unloading_in_progress = False  
 
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1906,6 +1906,8 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
         self.update_photocollections_affiliated_with_loaded_session()
 
         # Check if there are any photoscans with approved transducer tracking results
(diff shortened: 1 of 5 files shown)


OpenLIFUData/OpenLIFUData.py
```diff
@@ -1906,6 +1906,8 @@ def load_session(self, subject_id, session_id) -> None:
         self.update_photocollections_affiliated_with_loaded_session()
 
         # Check if there are any photoscans with approved transducer tracking results
+        # TODO: This assumes a single approved photoscan per session. This check needs to be udpated to allow for 
+        # approved_photoscan_id containing multiple photoscan IDs. 
         approved_photoscan_id = self.getParameterNode().loaded_session.get_transducer_tracking_approvals()
```
ebrahimebrahim, line 1911: Maybe there was previously a bug here? `get_transducer_tracking_approvals` is used without passing `approved_photoscans_only=True`, and so what it's doing here is to to get _all_ photoscan IDs for which there is an approved tt result, rather than getting _only_ the ones where the photoscan itself is also approved.

(Although I guess it shouldn't ever happen that a tt result is approved while the photoscan isn't, but still.)

OpenLIFULib/OpenLIFULib/transducer.py
```diff
@@ -194,7 +194,7 @@ def is_matching_transform(self, query_transform_node: vtkMRMLTransformNode) -> b
         query_transform_node.GetMatrixTransformToParent(query_transform_matrix)
 
         is_matching = (
-            slicer.util.arrayFromVTKMatrix(current_transform_matrix) == slicer.util.arrayFromVTKMatrix(query_transform_matrix)
-            ).all()
+            np.isclose(slicer.util.arrayFromVTKMatrix(current_transform_matrix), slicer.util.arrayFromVTKMatrix(query_transform_matrix))
+        ).all()
```
ebrahimebrahim, line 198: Note that I've modified the exact equality check here

Since there could be rounding or precision issues when the floats are read in and out, they could match without being literally equal bit for bit. So here it is better to use `np.isclose` which does a "fuzzy" equality check based on a resonable default for relative and absolute tolerance.

OpenLIFUData/OpenLIFUData.py
```diff
@@ -1905,28 +1905,27 @@ def load_session(self, subject_id, session_id) -> None:
...
-        approved_photoscan_id = self.getParameterNode().loaded_session.get_transducer_tracking_approvals()
-        if approved_photoscan_id:
-            approved_photoscan_id = approved_photoscan_id[0]
-            approved_tracking_result = None
+        # If there are any *approved* transducer tracking results that we have just loaded in newly_added_tt_result_nodes,
+        # then we check to see if any of them match the current transducer transform in terms of matrix values.
+        # If there is a match in matrix values, then the first such matching TT result that we encounter in the loop is 
+        # "officially" linked to the current transform by setting the "matching_transform" attribute, thereby ensuring that
+        # TT approval is revoked if the transducer is moved.
+        # Additionally, any other transducer tracking results whose matrix does not match current transducer get their approval revoked.
```
ebrahimebrahim, line 1913: Note here the way this works now

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -1674,9 +1648,9 @@ def checkCanRunTracking(self,caller = None, event = None) -> None:
             if transducer.surface_model_node is None: # Check that the selected transducer has an affiliated registration surface model
                 self.ui.runTrackingButton.enabled = False
                 self.ui.runTrackingButton.setToolTip("The selected transducer does not have an affiliated registration surface model, which is needed to run tracking.")
-            elif get_guided_mode_state() and (photoscan and not photoscan.photoscan_approved): # GM: Check that the selected photoscan is approved
-                self.ui.runTrackingButton.enabled = False 
-                self.ui.runTrackingButton.setToolTip("The selected photoscan has not been approved for transducer tracking.")
+            # elif get_guided_mode_state() and (photoscan and not photoscan.photoscan_approved): # GM: Check that the selected photoscan is approved
+            #     self.ui.runTrackingButton.enabled = False
+            #     self.ui.runTrackingButton.setToolTip("The selected photoscan has not been approved for transducer tracking.")
```
ebrahimebrahim, line 1653: @sadhana-r 
I commented these out temporarily, because otherwise it will be impossible in the custom app to proceed (you can't get the photoscan approved without opening the wizard, and this doesn't let you open the wizard until the photoscan is approved)

But we should decide how we want things to work, and then delete these lines or modify them accordingly.

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -72,6 +72,7 @@
 class FacialLandmarksMarkupPageBase(qt.QWizardPage):
     def __init__(self, parent=None):
         super().__init__(parent)
+        # TODO: This placelandmarks flag can probably be removed now. Since the self.page_locked flag conveys that same information.
```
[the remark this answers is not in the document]
sadhana-r: I have removed this flag now

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
sadhana-r: I hae removed this check since a photoscan no longer needs to be approved for transducer tracking. It is automatically approved when affiliated with an approved tracking result.

ebrahimebrahim (commented):
Everything is looking great except I found one bug:

When I enable manual transform interaction in photoscan-mri page and then disable it, it can unlock ICP button even when there are no fiducials. Clicking the ICP button then results in chaos.

I think I can fix this myself right now, so heads up that I will be pushing to this branch shortly, hopefully!

ebrahimebrahim:
Another small bug I noticed is that if you enable manual transform and then toggle the lock, the transform handles stay enabled

ebrahimebrahim:
I think I can fix this one quickly as well

> Another small bug I noticed is that if you enable manual transform and then toggle the lock, the transform handles stay enabled

sadhana-r:
Your fix/clean up looks good to me!

> Another small bug I noticed is that if you enable manual transform and then toggle the lock, the transform handles stay enabled

Hmm good catch. This bug will also be there when you enable manual transform on the transducer/photoscan registation page.

ebrahimebrahim:
Thanks for checking it! I just pushed a fix for the manual registration / lock bug. I'll do the same thing for the transducer/photoscan page.

ebrahimebrahim (approved):

ebrahimebrahim:
One last bug that I found but could not reproduce is that when I chaotically mess around with the back and next buttons, while also messing with the lock, I sometimes get locked into a mode where "next" is permanently disabled. Then I have to cancel and start over. It seems like a rare thing though, and I couldn't figure out how to reliably reproduce it. I'd like to let that bug through and open a bug report if we can actually come up with steps to reproduce it.

ebrahimebrahim:
@sadhana-r I've tested and am super happy with the current state of this! If you approve of the last changes I just added, please go ahead and hit the merge button.

sadhana-r:
Changes look good to me! And your plan to let though that bug for now makes sense to me. I'll also try to mess around and see if I can break things

PR #348, opened 2025-05-13
Fix protocols saved via the protocol configuration module not being able to be used for solution computation

### LINKED ISSUE #318: Protocols saved via the protocol configuration module cannot be used for solution computation
It seems that when we save a protocol using the protocol config module, there are then errors if we try to use the protocol for solution computation.

The specific error I ran into was

```
  File "/home/ebrahim/openwater/openlifu-superbuild/python-install/lib/python3.9/site-packages/openlifu/seg/seg_methods/seg_method.py", line 62, in _map_params
    param = xa.DataArray(np.zeros(seg.shape), coords=seg.coords, attrs={"units": info["units"], "long_name": info["name"], "ref_value": ref_mat.get_param(param_id)})
AttributeError: 'NoneType' object has no attribute 'shape'
```

but @arhowe00 I think you already knew something was up with how protocol config module handles seg methods and it's being worked on in OpenLIFU-python?

arhowe00:
Closes #318 

Addresses the new structure of `SegmentationMethod` and some cases in which an existing protocol had decimal values set to 0 due to a lack of precision in the GUI. Note that for this last case, an issue was opened in [OpenLIFU-python#302](https://github.com/OpenwaterHealth/OpenLIFU-python/issues/302), which is the correct place to address the possibility of setting a `Pulse.duration` to 0 in the protocol config module. Protocols configured in the protocol config module can now be used for solution computation.

## For review

Try using the protocol configuration module to set some parameters and possibly experiment with some random interesting parameter constraints. Then compute a sonication solution.

### DIFF
--- a/OpenLIFULib/OpenLIFULib/Resources/python-requirements.txt
+++ b/OpenLIFULib/OpenLIFULib/Resources/python-requirements.txt
@@ -1,2 +1,2 @@
-openlifu==0.3.1
+git+https://github.com/OpenwaterHealth/OpenLIFU-python.git@0e09d41ae8eb862319de0598187687db21b3081b
 bcrypt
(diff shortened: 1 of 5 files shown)


ebrahimebrahim (approved):
It worked well after rebasing, fixing merge conflicts, and addressing a rebase issue and a bug.

Great work addressing this seg method situation!

PR #353, opened 2025-05-13
Add protocol validity indicator

### LINKED ISSUE #324: Validate protocols in the protocol config module

### LINKED ISSUE #334: Add protocol validity indicator
Add a more responsive validity indicator that gives the researcher prior knowledge that a protocol may not be published to the database due to `__post_init__` failures.

arhowe00:
Closes #334 

This adds a simple validity indicator which checks that when a field is edited, whether it would produce an erroneous value. 

## For review:

A code review should be good enough, but you can also try dynamically changing some values in the protocol configuration and checking that any exceptions that would be thrown are reported at the bottom. You can try switching between protocols that have/don't have errors and fixing your errors, and *before saving*, you should know that your protocol has at least one error.

**Note**: We do not add the ability to list all the errors because some exceptions are thrown in control flow within `__post_init__`; some exceptions involving objects such as `Sequence` are raised in a mutually exclusive manner. If desired, we can still attempt to handle exceptions by collecting them in a list and then print them all, but this was omitted for now. This also potentially leads to an arbitrary long list of exceptions that would stretch the label until blocking other UI elements. Therefore the language is clear by [saying "at least one error,"](https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/353/commits/e112a4194bb85e6dfc59deab1dbed5df5ad7e815#diff-0416f7e8759becf4cc7996173be0ce7660701c1c218c3e38dd9b878f465c154fR834) and in this PR, we only list one exception.

## Before merging:

- [x] Make sure to merge PR #332 , which closes #324 . This branches off of that code as they handle very similar functions in the codebase.  Once PR #332 is merged, rebase this to main and then merge.

### DIFF
--- a/OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
+++ b/OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
@@ -197,6 +197,7 @@ class OpenLIFUProtocolConfigWidget(ScriptedLoadableModuleWidget, VTKObservationM
 
         self._cur_protocol_id: str = ""  # important if WIPs change the ID
         self._cur_save_state = SaveState.NO_CHANGES
+        self._editor_is_enabled: bool = False
         self._parameterNode: Optional[OpenLIFUProtocolConfigParameterNode] = None
         self._parameterNodeGuiTag = None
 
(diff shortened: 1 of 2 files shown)


OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -289,23 +290,25 @@ def setup(self) -> None:
...
-        self.parameter_constraints_widget.table.itemChanged.connect(lambda *_: trigger_unsaved_changes())
-        self.target_constraints_widget.table.itemChanged.connect(lambda *_: trigger_unsaved_changes())
-        self.solution_analysis_options_definition_widget.add_value_changed_signals(trigger_unsaved_changes)
-        self.virtual_fit_options_definition_widget.add_value_changed_signals(trigger_unsaved_changes)
+        trigger_invalid_protocol_check = lambda: self.updateWidgetProtocolValidityIndicator()
+
+        for f in [trigger_unsaved_changes, trigger_invalid_protocol_check]:
+            self.ui.protocolNameLineEdit.textChanged.connect(f)
+            self.ui.protocolIdLineEdit.textChanged.connect(f)
+            self.ui.protocolDescriptionTextEdit.textChanged.connect(f)
```
ebrahimebrahim, line 298: on textChanged is very eager and will run these triggers a lot; consider using the editingFinished signal?

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
```diff
@@ -289,23 +290,25 @@ def setup(self) -> None:
...
+            self.sequence_definition_widget.add_value_changed_signals(f)
+            self.abstract_focal_pattern_definition_widget.add_value_changed_signals(f)
+            self.sim_setup_definition_widget.add_value_changed_signals(f)
+            self.abstract_delay_method_definition_widget.add_value_changed_signals(f)
+            self.abstract_apodization_method_definition_widget.add_value_changed_signals(f)
+            self.abstract_segmentation_method_definition_widget.add_value_changed_signals(f)
+            self.parameter_constraints_widget.table.itemChanged.connect(lambda *_: f)
+            self.target_constraints_widget.table.itemChanged.connect(lambda *_: f)
+            self.solution_analysis_options_definition_widget.add_value_changed_signals(f)
+            self.virtual_fit_options_definition_widget.add_value_changed_signals(f)
```
ebrahimebrahim, line 311: similarly here, you can change some of the signals used in `OpenLIFUAbstractDataclassDefinitionFormWidget.add_value_changed_signals` (and probably rename that function slightly)

ebrahimebrahim (approved):
It's really nice! It's a bit eager to have the triggers happen on valueChanged/textChanged rather than on editingFinished though. Like if I am inputting a number it would be eagerly attempting to construct protocols while I am still typing and not finished entering the number.

But I'll leave it to you to decide whether to try using the editingFinished signals instead -- I am not sure how much it would break things to do that.

If you did want to try it, I have indicated in my review comments where you'd probably have to make the changes.

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
arhowe00: Yes, there is an editingFinished signal that we could use and would make a very trivial edit to this signal.  I think if you are veering toward this change being the intuitive choice, we should make it. It is quite reactive to show the errors while editing, but the errors would also disappear equally as rapidly as when the error is fixed. \

So, if somebody is interacting directly with that error box (i.e. guiding their movements based on its current text in real time), the more reactive choice would be superior.  However, this depends on whether that is a common interaction.  Do you think that the user is more likely to acknowledge their changes after an edit has been made and they have focused on something else?

OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
arhowe00: Agreed, depending on the answer to above, we should just change what the function does. yes we could have both, but this is the only place where it's used, so whether we go with `__Changed` or `editingFinished`, we should just change what that is called.

ebrahimebrahim:
Accepting these changes as is; we can change the reactivity in later work if it turns out to be an issue. Thanks!

PR #355, opened 2025-05-13
Give default admin when no db connected

### LINKED ISSUE #354: No connected db should imply superuser privileges
Since the custom app automatically tries connection of DB, if there is no connected DB we must assume superuser privileges. The current app behavior only grants superuser privileges if there is a connected DB and no admin in the database. It likely went undiscovered because you can load the database as a restricted (anonymous) user, but you can't set the database location, and we usually always started user account mode after connecting the database.

When updating the widget login state, if there is no connected database, the widget login state should act as the default admin user.

This will deal with [OpenLIFU-app#39](https://github.com/OpenwaterHealth/OpenLIFU-app/issues/39)

arhowe00:
Closes #354 

This also makes simple refactorings and edits to comments and typos in displayed strings.

## For review

Code review should be enough. However, if interested in trying, start the app, skip connecting a db, and activate user account mode. You should be a superuser (i.e. have admin privileges despite a DB not being connected).

## After merging
- [ ] Close [OpenLIFU-app#39](https://github.com/OpenwaterHealth/OpenLIFU-app/issues/39) in a commit that updates the SlicerOpenLIFU revision, as this PR should address it

### DIFF
--- a/OpenLIFULogin/OpenLIFULogin.py
+++ b/OpenLIFULogin/OpenLIFULogin.py
@@ -569,8 +569,15 @@ class OpenLIFULoginWidget(ScriptedLoadableModuleWidget, VTKObservationMixin, Gui
                 password_hash = "",
                 roles = [],
                 name = "Anonymous",
-                description = "This is the default role set when the app opens, without anyone logged in, and when user account mode is deactivated. It has no roles, and therefore is the most restricted."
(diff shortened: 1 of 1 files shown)


ebrahimebrahim (approved):
Code looks good! Just updated the message to make it a bit more clear what's going on in "default admin" mode

PR #377, opened 2025-05-27
Hide user account mode banner in guided mode (#344)

### LINKED ISSUE #344: Hide signed-in bar when not necessary
From Peter:

> Signed in bar at the top is not necessary if signin is the initial landing page and switching users mid session is not a thing

We can achieve this by tagging the sign in bar as "hidden in guided mode". That way it is still visible in the rare case where someone enables user acct mode while using the extension.

arhowe00:
Closes #344 

This adds the `slicer.openlifu.hide-in-guided-mode` property set to `True` to all `UserAccountBanner`s at the end of their construction. The current intended guided mode experience is that the banner is hidden.

Note: If `enforceGuidedModeVisibility()` is called before the `UserAccountBanner`s are inserted into each module, they will fail to be hidden.

## For review:

Read the "Note" above. Then, review the code.

I have noticed in testing that the banners are always removed when enabling guided mode, so things work as expected. In the custom app, I suspect things will behave likewise, as the algorithm inputs *also* are added in via `replace_widget` (so the "Note" above applies to algorithm input widgets as well). The algorithm inputs do not fail to enforce guided mode visibility in the custom app.

### DIFF
--- a/OpenLIFULib/OpenLIFULib/user_account_mode_util.py
+++ b/OpenLIFULib/OpenLIFULib/user_account_mode_util.py
@@ -59,6 +59,8 @@ class UserAccountBanner(qt.QWidget):
         group_layout.addWidget(self.go_to_login_button)
 
         top_level_layout.addWidget(group_box)  # Add the group box to the top_level_layout
+
+        self.setProperty("slicer.openlifu.hide-in-guided-mode", True)  # dynamic property to hide in guided mode
     
     def change_active_user(self, new_active_user: Optional["openlifu.db.User"]):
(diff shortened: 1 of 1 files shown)


ebrahimebrahim (approved):
I am also optimisitc that this will work in the custom app for the reasons you give. Either before or after merging, can you test that it does indeed work in the custom app?

arhowe00:
> I am also optimisitc that this will work in the custom app for the reasons you give. Either before or after merging, can you test that it does indeed work in the custom app?

Confirmed.

image

Issue #364, opened 2025-05-23
Verify that virtual fit transform isn't getting flipped incorrectly


ebrahimebrahim:
It was observed that the virtual fit transform might be getting applied in a flipped fashion somehow. Debug and make sure this isn't happening, and fix it if it is.

PR #381, opened 2025-05-28
VF debugging utilities, VF transform bug fix, Mesh recon controls, and VF and mesh recon progress bars

ebrahimebrahim:
- Adds progress bars for both virtual fitting and mesh reconsruction
- Exposes controls for things that Daniel added in https://github.com/OpenwaterHealth/OpenLIFU-python/pull/315
- Adds visualizations of the virtual fit debug info
- Fixes the bug #364, which was indeed a bug

# For review

- Try the virtual fit debugging and progress bar, checking the following
  - The debug checkbox is hidden when you turn on guided mode.
  - When you run virtual fit, a progress bar shows up and progresses through the steps.
  - When you run virtual fit with the debug checkbox enabled, a bunch of model nodes are added to the scene: a skin mesh, an interpolated skin mesh, and three "search point" displays. The points should be colored.
- Mesh generation checks
  - When you click "start mesh generation" (or whatever it was called) and select a photoscan, the dialog that shows up now lets you select whether you want sequential image matching, and also lets you put in a photo sampling rate.
  - Running the actual mesh generation is more difficult to test because you need meshroom, and I don't want to waste your time on setting that up. You don't need to test it if you haven't set up meshroom. I tested it, you can trust me :). The progress bar is super nice.
- For code you could only look at the short commit "Fix volume array axis permutation in virtual fit call". I think this was causing the bug #364
  - You could also try plopping the target somewhere interesting and running virtual fit. Hopefully the transducer does something more sensible than it used to (it's subtle to notice sometimes, but it definitely used to be a little wonky) and this should indicate that #364 has been addressed.

# Before merging

- [x] Ensure that https://github.com/OpenwaterHealth/OpenLIFU-python/pull/323 is merged
- [x] Update the python requirements

### DIFF
--- a/OpenLIFUTransducerTracker/Resources/UI/OpenLIFUTransducerTracker.ui
+++ b/OpenLIFUTransducerTracker/Resources/UI/OpenLIFUTransducerTracker.ui
@@ -107,8 +107,8 @@
          </item>
          <item>
           <widget class="QLabel" name="photoscanGenerationStatusMessage">
-           <property name="text">
-            <string>(placeholder_photoscangeneration_status_label)</string>
+           <property name="styleSheet">
+            <string notr="true">font-style: italic</string>
            </property>
            <property name="wordWrap">
             <bool>true</bool>


ebrahimebrahim:
> I think `clear_virtual_fit_results` needs to be updated to also clear debug information associated with the result. If we re-run virtual fit, its not apparent which result the debug info is assocciated with. Also, since we can have multiple virtual fit results for different targets, I think it would be helpful to include the target ID in the debug node names, or as a node attribute? Whatever is easier to keep track of.

The debugging info is not meant to be part of the normal application workflow. It's just sort of a disorganized dump, and you are meant to exit the SlicerOpenLIFU world and use the full power of Slicer to inspect and debug. I want to avoid adding any infrastructure around the debug info. I see it as the mrml scene equivalent of sticking a print statement in the code to work out a bug. So I want to leave it as is, where it doesn't get cleaned up with the session, as though you used slicer to import some data unrelated to SlicerOpenLIFU. What do you think?

OpenLIFUTransducerTracker/Resources/UI/OpenLIFUTransducerTracker.ui
```diff
@@ -107,8 +107,8 @@
          </item>
          <item>
           <widget class="QLabel" name="photoscanGenerationStatusMessage">
-           <property name="text">
-            <string>(placeholder_photoscangeneration_status_label)</string>
+           <property name="styleSheet">
+            <string notr="true">font-style: italic</string>
```
ebrahimebrahim, line 111: Post-review change: I made this italic to match the other one. I feel like progress status looks better in italic since it makes it feel like a dynamic text somehow

sadhana-r (approved):

PR #382, opened 2025-05-28
Improve display options within transducer tracking wizard

sadhana-r:


sadhana-r:
@arhowe00 also ready for review now!

arhowe00 (changes requested):
In response to:

> Check that the wizard displays the transducer body correctly, without affecting main scene display

The main scene display doesn't get modified throughout the process other than it only showing the gridlines during tracking (this is the "surface model" ?).
image

Everything else works. I have some minor suggestions in the code. Since you branched off `upstream/361_transducer_model_color`, I made sure my comments only reference the added commits

PR #385, opened 2025-06-05
Add quantitative metric comparing vf and tracked result

### LINKED ISSUE #362: Show quantitative measure of virtual fitted vs tracked transducer poses
In the TT module info messages area, show a quantitative measure of how different the VF result is from the tracked position. Linear distance between origins should do.

sadhana-r:
Closes #362 

- [x] Add logic and label for displaying a quantitative measure comparing the virtual fit and transducer tracking transform.
- [x] Update when the label is shown/hidden. It should be shown when a approved transducer tracking result is added to the scene. When approval is revoked i.e. the transducer is moved, hide the label since the result is no longer valid. 

For review:

-Relevant code changes are the last 2 commits.
-Check that when there is an approved tracking result in the scene (either when loaded in or after running tracking), the distance from virtual fit metric is displayed. If there is an approved tracking result, but no virtual fit result, then an error message is displayed. Otherwise, the label is hidden.

### DIFF
--- a/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
+++ b/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
@@ -1575,12 +1575,7 @@ class OpenLIFUTransducerTrackerWidget(ScriptedLoadableModuleWidget, VTKObservati
         self.algorithm_input_widget = OpenLIFUAlgorithmInputWidget(algorithm_input_names)
         replace_widget(self.ui.algorithmInputWidgetPlaceholder, self.algorithm_input_widget, self.ui)
         self.updateInputOptions()
-        self.algorithm_input_widget.connect_combobox_indexchanged_signal(self.checkCanRunTracking)
-        self.algorithm_input_widget.connect_combobox_indexchanged_signal(self.updateApprovalWarningsIfAny)
-        # Update the enabled-ness of the virtual fit rendering checkbox
-        self.algorithm_input_widget.connect_combobox_indexchanged_signal(self.updateVirtualFitResultForDisplay) 
-        # Update the enabled-ness of the moderl rendering options
-        self.algorithm_input_widget.connect_combobox_indexchanged_signal(self.updateModelRenderingSettings)
+        self.algorithm_input_widget.connect_combobox_indexchanged_signal(self.updateInputRelatedWidgets)
 
         # ---- Model rendering options ----
         self.ui.viewVirtualFitCheckBox.stateChanged.connect(self.updateVirtualFitResultForDisplay)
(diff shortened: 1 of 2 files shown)


OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -1839,7 +1972,46 @@ def revokeTransducerTrackingApprovalIfAny(self, photoscan_id: str, reason:str):
             )
             self.logic.revoke_transducer_tracking_approval(photoscan_id = photoscan_id)
             self.updateApprovalStatusLabel()
+            self.updateDistanceFromVFLabel()
```
arhowe00, line 1975: Here, you propagate an update to the VF distance label only if transducer tracking is approved. However, this seems like it makes some of the logic within the update to the VF label redundant. Both check against `get_transducer_tracking_approval`, for instance. 

Though the message is hidden when approval is revoked, the notion of "distance" is more directly tied to the position of the transform. What are your thoughts on propagating updates every time the transform is moved instead of only when there is an approval? That way, the approval status is checked against independently in `updateDistanceFromVFLabel`, and if code is replicated or repurposed, the most relevant logic is directly in context of the distance update?

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -1564,15 +1574,22 @@ def setup(self) -> None:
...
+        self.ui.photoscanOpacitySlider.valueChanged.connect(self.updateModelRendering)
+        
+        # ---------------------------------
         self.ui.runTrackingButton.clicked.connect(self.onRunTrackingClicked)
         self.ui.previewPhotoscanButton.clicked.connect(self.onPreviewPhotoscanClicked)
 
+        # These ui elemeents are not specific to the currently selected input options
         self.updatePhotoscanGenerationButtons()
         self.updateApprovalStatusLabel()
-        self.updateApprovalWarningsIfAny()
```
arhowe00, line 1575: Why was this also removed?

arhowe00 (approved):
Looks great. I have one question about a choice with the update routine, but even without the change everything still works great. Approving for merging.

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -1591,16 +1586,11 @@ def setup(self) -> None:
...
-            "Euclidean distance between the virtual fit and tracked transform origins.")
-        #self.ui.quantitativeTransducerTrackingMetricLabel.hide()
 
-        self.updateVirtualFitResultForDisplay()
+        # These ui elemeents are not specific to the currently selected input options
         self.updatePhotoscanGenerationButtons()
         self.updateApprovalStatusLabel()
-        self.updateApprovalWarningsIfAny()
         self.updateWorkflowControls()
-        self.updateModelRenderingSettings()
```
sadhana-r: Same thing here. This is called as part in `updateInputRelatedWidgets`

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
sadhana-r: I'm not sure I am following. `revokeTransducerTrackingApprovalIfAny` is called whenever the transform is moved i.e. in OpenLIFUData, `on_transducer_transform_modified`. Are you suggesting I call `updateDistanceFromVFLabel` directly from the data module or within `revokeTransducerTrackingApprovalIfAny`, but outside of the if satement? I kept it within the if statement because only revoking approval from a previously approved result would need to trigger a change to the label (i.e. line 1973). Calling it when the tracking result approval state isn't changed would be redundant. 

> Though the message is hidden when approval is revoked, the notion of "distance" is more directly tied to the position of the transform. What are your thoughts on propagating updates every time the transform is moved instead of only when there is an approval?

Also, a user shouldn't be able to edit a valid transducer transform outside the context of the wizard. So the distance shouldn't update dynamically as the transform is moved.

PR #390, opened 2025-06-06
Push notifications instead of raising dialogs when revoking approvals

### LINKED ISSUE #335: Smooth out UX around auto-revoking of approvals
Currently if someone tries to move something that affects some kind of approval (e.g. moving a target after a solution was approved), the approval is revoked and an info dialog pops up explaining as much.

Having something pop up that you have to interact with as soon as you try to move something is quite intrusive. If it's going to require interaction, then it should at least give the choice to not proceed with the modification and not revoke approval (and e.g. not destroying a solution).

Two possible solutions:

(1) Remove the dialog and reduce it to a non-intrusive message. The status bar could be used for the message, but it might be better to create our own push-notification-like widget with noticeable but non-interrupting disappearing messages.

OR

(2) Instead of raising an _info_ dialog when the object is _moved_, raise a _question_ dialog when the movement is _unlocked_ (e.g. the lock button is pressed for a target) explaining that approval would be revoked and confirming whether the user wants to proceed.

ebrahimebrahim:
When approvals get revoked or when virtual fit results need to be cleared out, we now just show a notification rather than interrupting the user with a dialog that they have to interact with.

This adds a handy notification system that we can make use of elsewhere too

Closes #335

For review: Try loading the Soren example, which already comes with an approved virtual fit, and then move the target around. Instead of interrupting you, it should show two notifications. Make sure the notifications don't look too ridiculous on your system. They look nice on my system. Give the code changes a cursory look.

### DIFF
--- a/OpenLIFULib/OpenLIFULib/notifications.py
+++ b/OpenLIFULib/OpenLIFULib/notifications.py
@@ -1,0 +1,85 @@
+from typing import Optional
+import logging
+import slicer
+import qt
+
+LAST_NOTIFICATION = None # globally tracks the last notification that was shown
+
+class Notification(qt.QWidget):
+    """
+    Notification widget that can be shown temporarily on top of its parent.
+    """
+    def __init__(self, message:str, parent=None, previous_notification:"Optional[Notification]"=None):
+        """
+        If parent is not provided then it will be the Slicer layout container that contains the slice and 3D views.
+
+        Another Notification object can optionally be provided as `previous_notification`.
+        If one is provided then this notification will be positioned below `previous_notification`, rather than 
+        being at the top-middle-ish of the parent widget.
+        """
+        if parent is None:
+            parent = slicer.app.layoutManager().parent()
+
+        super().__init__(parent)
+        
+        self.previous_notification = previous_notification
+
+        self.setWindowFlags(
+            qt.Qt.Tool | # tooltip like behavior
+            qt.Qt.FramelessWindowHint |
+            qt.Qt.WindowStaysOnTopHint
+        )
+        self.setAttribute(qt.Qt.WA_ShowWithoutActivating) # prevents widget from grabbing focus
+        self.setAttribute(qt.Qt.WA_TranslucentBackground)
+
+        self.setStyleSheet("""
+            QWidget {
+                background-color: rgba(30, 30, 30, 220);
+                color: white;
+                border-radius: 5px;
+            }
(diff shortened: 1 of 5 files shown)


OpenLIFULib/OpenLIFULib/notifications.py
```diff
@@ -0,0 +1,85 @@
...
+
+        if self.previous_notification is not None and self.previous_notification.isVisible():
+            self.move_to_below_previous_notification()
+        else:
+            self.move_to_top_center_of_parent()
+
+        self.show()
+        qt.QTimer.singleShot(duration_ms, self.close)
+
+def notify(message:str) -> None:
```
ebrahimebrahim, line 73: this is very easy to use!

```
from OpenLIFULib.notifications import notify

notify("a message")
```

OpenLIFUPrePlanning/OpenLIFUPrePlanning.py
```diff
@@ -326,20 +327,16 @@ def onLockModified(self, caller, event):
...
     def clearVirtualFitResultsIfAny(self,target: vtkMRMLMarkupsFiducialNode, reason:str):
-        """Clear virtual fit results for the target from the scene if any, and show a message dialog for the user to confirm deletion.
+        """Clear virtual fit results for the target from the scene if any.
         """
         target_id = fiducial_to_openlifu_point_id(target)
         session = get_openlifu_data_parameter_node().loaded_session
         session_id = None if session is None else session.get_session_id()
         
         if list(get_virtual_fit_result_nodes(target_id, session_id)):
-            if slicer.util.confirmYesNoDisplay(
```
ebrahimebrahim, line 336: Notice that we now delete the virtual fit results without confirmation! I think this makes sense to do. It wouldn't make sense to keep the virtual fit results after the user invalidated them, unless we want to somehow undo the user's move -- not a path I want to go down! If we want to prevent this from happening accidentally then that needs to be done at the level of unlocking a target for movement, which is a separate matter from what this PR is handling

arhowe00 (approved):
This notification infrastructure is excellent. In future work, we should explore a decorator or an additional parameter with that allows certain levels of notification. We have a suppressed level, the non-obtrusive implementation of `notify`, and possibly one that uses the error display.

Note: Depending on the chosen layout, the notification can look slightly awkward. 

Screenshot 2025-06-08 at 11 51 53 PM

Approved for merging.

PR #391, opened 2025-06-06
Mask the background out of loaded MRIs (#387)

### LINKED ISSUE #387: Mask out background of MRI slices
Mask out background of MRI slices using the foreground mask that can be generated using the skin seg tools in openlifu

ebrahimebrahim:
Closes #387

One thing I slightly dislike about this method is that it modifies the values in the volume array: it makes the background voxels take a value that is 1 less than the minimum value in the volume. This way they can simply be thresholded out. There should be other ways to do this that involve creating a segmentation node out of the foreground mask and then using it as a "cropping ROI" for the volume. This would be nicer but then it's another piece of data to manage. This method  keeps things simple, and it doesn't seem to harm the downstream skin segmentation computation.

Another thing I slightly dislike about this method is that it takes some time to compute the foreground mask, so this makes loading a volume into the scene a slower process. But it isn't too bad, at least on my system, and the result is definitely very pleasing to look at.

For review: Load a volume into the scene by loading a session, or by doing it manually, and observe that it is immediately masked after being loaded in. Give the code a cursory look.

### DIFF
--- a/OpenLIFULib/OpenLIFULib/volume_thresholding.py
+++ b/OpenLIFULib/OpenLIFULib/volume_thresholding.py
@@ -1,0 +1,48 @@
+"""Volume thresholding tools"""
+
+import logging
+import vtk
+import slicer
+from slicer import vtkMRMLScalarVolumeNode
+from OpenLIFULib.util import BusyCursor
+from OpenLIFULib.lazyimport import openlifu_lz
+
+def cast_volume_to_float(volume_node:vtkMRMLScalarVolumeNode) -> None:
+    """Converts a volume node to float, replacing the underlying vtkImageData."""
+    image_cast = vtk.vtkImageCast()
+    image_cast.SetInputData(volume_node.GetImageData())
+    image_cast.SetOutputScalarTypeToDouble()
+    image_cast.Update()
+    volume_node.SetAndObserveImageData(image_cast.GetOutput())
+
+    # I am not certain that the display node will know to update itself to handle the new image data type,
+    # so I hope poking `CreateDefaultDisplayNodes` here makes it do the right thing. If it's not needed then it's harmless anyway:
+    volume_node.CreateDefaultDisplayNodes()
+
+def threshold_volume_by_foreground_mask(volume_node:vtkMRMLScalarVolumeNode) -> None:
+    """Compute the foreground mask for a loaded volume and threshold the volume to strip out the background.
+    This modifies the values of the background region in the volume and sets them to 1 less than the minimum value in the volume.
+    This way we can simply enable volume thresholding to remove
+    It can take a moment to actually compute the foreground mask.
+    """
+    volume_array = slicer.util.arrayFromVolume(volume_node)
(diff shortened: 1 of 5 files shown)


ebrahimebrahim:
Oh wow, glad you tried those examples! I will think about how to deal with the integer overflow. Maybe always forcing float type

> My understanding from the meeting yesterday was that Peter wanted the volume to get masked when the skin segmentation is computed and displayed during virtual fit (or tracking) since it was anyways computing the foreground mask. That way the computation does get repeated when loading and then running those pipelines. But this works too! The loading is a little slower but not too annoying.

Yeah I think your understanding of the meeting is correct. I am cheating because I didn't want to add so much logic for sharing the mask result. And then if someone saves the session after pre-planning and then loads again later then the mask would be gone, unless we also add logic for writing it to the database somehow. This seemed like a good tradeoff. I wonder if we can speed up the `compute_foreground_mask` function a bit...

sadhana-r:
> And then if someone saves the session after pre-planning and then loads again later then the mask would be gone, unless we also add logic for writing it to the database somehow. This seemed like a good tradeoff. I wonder if we can speed up the `compute_foreground_mask` function a bit...

Hmmm true. Related to this, if a user loads a session where tracking or planning has already been run, the skin-segmentation is not shown unless the user re-runs virtual fit or goes through and 're-approves' the tracking result. This isn't ideal but I left it as is. 
One idea:
We could on `load_session`, compute the skin_segmentation if the loaded session includes an approved tracking result or virtual fit result (we already check for that). This will then resolve the need to save the mask etc. Basically, whenver you have a VF result to tracking result, compute the skin surface. And we can update `openlifu` to return the mesh and foreground mask when `compute_skin_mesh_from_volume` is called.

OpenLIFULib/OpenLIFULib/volume_thresholding.py
```diff
@@ -0,0 +1,48 @@
...
+    This way we can simply enable volume thresholding to remove
+    It can take a moment to actually compute the foreground mask.
+    """
+    volume_array = slicer.util.arrayFromVolume(volume_node)
+    volume_array_min = volume_array.min()
+
+    background_value = volume_array_min - 1
+    if background_value < volume_node.GetImageData().GetScalarTypeMin(): # e.g. if volume_array_min is 0 and it's an unsigned int type
+        logging.info("Casting volume to float for the sake of `threshold_volume_by_foreground_mask`.")
+        cast_volume_to_float(volume_node)
```
ebrahimebrahim, line 34: This is how I addressed the problem you pointed out: if the value I am trying to use for background (e.g. `-1`) is less than the minimum for the data type (e.g. 0 for unsigned integer types), then we cast to float before proceeding.

sadhana-r (approved):
looks good to me!

ebrahimebrahim:
Merging this without adding #399 fixes like I had suggested in a meeting. I will actually do #399 in a separate PR to keep things simpler.

PR #396, opened 2025-06-09
Remove all observers added by the wizard

### LINKED ISSUE #272: vtkDebugLeaks detected on exit on Windows
On windows, debug leaks are caught when exiting the app. This only happens when interacting with the transducer tracking wizard before exiting.

Image

sadhana-r:
Closes #272 

As Sam suggested, the debug leaks seems to be caused by vtkObservers remaining after node removal.  I tested building the custom app on Windows using this branch and this seems to fix the issue. 
I also found that simply using `node.RemoveAllObservers()` does not work. I had to tag and remove the individual observers. 

When testing, I found a bug in the wizard workflow which allows the user to click `Next`/`Approve` even if there wasn't a tracking transform defined. So I include a fix for that in this PR. 

### For Review
- Look over the code
- Confirm that this changes resolves the debug leaks error on the custom app built on Linux (git tag: 1e1340c7fa68aae7efaef130a7388d552135d899)

### DIFF
--- a/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
+++ b/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
@@ -1,4 +1,5 @@
 # Standard library imports
+from collections import defaultdict
 import itertools
 import warnings
 from subprocess import CalledProcessError
@@ -12,6 +13,7 @@ import vtk
 # Slicer imports
 import slicer
 from slicer import (
+    vtkMRMLNode,
     vtkMRMLMarkupsFiducialNode,
     vtkMRMLModelNode,
     vtkMRMLScalarVolumeNode,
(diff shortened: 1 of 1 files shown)


sadhana-r:
I got an error once related to `unwatch_fiducial_node` in the PrePlanning module, that includes functionality that gets called whenever markupsFiducual nodes are added/removed (https://github.com/OpenwaterHealth/SlicerOpenLIFU/blob/5c4d2987bd205abb4cfb85374a8a777c27a05f30/OpenLIFUPrePlanning/OpenLIFUPrePlanning.py#L275). I wasn't able to re-produce the error again, so not sure if its a major issue but I wonder if I should either reference the `self._running_wizard` flag in the Pre-planning module so that it ignores any fiducial nodes added by the wizard, or as Andrew suggested, tag all wizard-level nodes with an attribute that the pre-planning module ignores. What do you think?

PR #397, opened 2025-06-10
Refactor subject/session management in data module

### LINKED ISSUE #369: Allow viewing subject level info before loading a session
From Peter:

> Loading a subject without a session doesn't tell you any information about the subject. I should be able to see their name, what volumes they have attached, etc.

This would actually be a big task right now because we never really load subjects as persistent scene data (there is no "SlicerOpenLIFUSubject" and there's no such thing as a "currently loaded subject"). But here is a way to make it small: after selecting a subject in the subject/session selector tree view, there should be a button and/or context menu option to show the subject's info. The button could be called "view subject details" or something like that, and it can result in a dialog that just displays some info temporarily.

### LINKED ISSUE #370: Allow creating new session when session is selected
Create New Session should probably be accessible even when selecting a session in the selector. The distinction of which type of row you have selected is subtle

### LINKED ISSUE #371: Show subject name in create new session dialog
In the "create new session" dialog, we should show the Subject Name (not editable) so that the user isn't confused and puts subject Name under session Name.

arhowe00:
Closes #369 
Closes #370 
Closes #371 

This range of commits remodels the Data module in accordance the Openwater team's preferences. Not all the changes made are reflected in the issues above. The old subject/session tree was replaced with a cleaner, isolated workflow. Some logic was cleaned up. The result is a more reliable way of managing subjects, sessions, and their associated data.

I can squash lots of the incremental changes if they are not good for the history.

## For review

This PR has a lot of code changes and commits, so the easiest way to review this it to test out a workflow on a recently merged module. Try to unload some sessions, create new subjects and sessions, and run through the workflow. A random walk that goes through virtual fitting, transducer tracking, and sonication planning once or twice and involves at least three different sessions and two different subjects total would be good enough. Please report any related errors.

Lastly, take a close look at the new functionality in the data module (and what happens when you open certain dialogs). Did I miss something?

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -136,8 +136,6 @@ class CreateNewSessionDialog(qt.QDialog):
 
     def setup(self) -> None:
 
-        self.setMinimumWidth(200)
-
         formLayout = qt.QFormLayout()
         self.setLayout(formLayout)
 
@@ -223,8 +221,6 @@ class AddNewVolumeDialog(qt.QDialog):
 
     def setup(self):
 
-        self.setMinimumWidth(400)
-
         formLayout = qt.QFormLayout()
         self.setLayout(formLayout)
 
@@ -302,7 +298,6 @@ class LoadSubjectDialog(qt.QDialog):
 
         self.setWindowTitle("Load Subject")
         self.setWindowModality(qt.Qt.WindowModal)
-        self.resize(800, 400)
 
         self.db = db
         self.selected_subject: Optional["openlifu.db.Subject"] = None
@@ -312,9 +307,6 @@ class LoadSubjectDialog(qt.QDialog):
     def setup(self) -> None:
         self.boxLayout = qt.QVBoxLayout()
         self.setLayout(self.boxLayout)
-        self.setMinimumSize(800, 400)
-        self.setMaximumSize(1000, 600)
-
         # ---- Subjects Table ----
         cols = [
             "Subject Name",
@@ -338,9 +330,8 @@ class LoadSubjectDialog(qt.QDialog):
 
         header = self.tableWidget.horizontalHeader()
         header.setSectionResizeMode(0, qt.QHeaderView.Interactive)
-        header.resizeSection(0, 250)
-        header.setSectionResizeMode(1, qt.QHeaderView.Interactive)
-        header.resizeSection(1, 200)
+        header.setSectionResizeMode(0, qt.QHeaderView.ResizeToContents)
+        header.setSectionResizeMode(1, qt.QHeaderView.ResizeToContents)
         header.setSectionResizeMode(2, qt.QHeaderView.ResizeToContents)
         header.setSectionResizeMode(3, qt.QHeaderView.Stretch)
 
@@ -390,7 +381,7 @@ class LoadSubjectDialog(qt.QDialog):
             self.tableWidget.setItem(row, 1, qt.QTableWidgetItem(subject.id))
             self.tableWidget.setItem(row, 2, qt.QTableWidgetItem(str(num_volumes)))
             self.tableWidget.setItem(row, 3, qt.QTableWidgetItem(str(num_sessions)))
-            self.tableWidget.setRowHeight(row, 36)
+        self.tableWidget.resizeRowsToContents()
 
     def appendSubjectToList(self, subject: "openlifu.db.subject.Subject") -> None:
         row = self.tableWidget.rowCount
@@ -403,7 +394,7 @@ class LoadSubjectDialog(qt.QDialog):
         self.tableWidget.setItem(row, 1, qt.QTableWidgetItem(subject.id))
         self.tableWidget.setItem(row, 2, qt.QTableWidgetItem(str(num_volumes)))
         self.tableWidget.setItem(row, 3, qt.QTableWidgetItem(str(num_sessions)))
-        self.tableWidget.setRowHeight(row, 36)
+        self.tableWidget.resizeRowsToContents()
 
     def on_add_subject_clicked(self, checked:bool) -> None:
         subjectdlg = AddNewSubjectDialog()
@@ -483,7 +474,6 @@ class LoadSessionDialog(qt.QDialog):
 
         self.setWindowTitle(f"View Sessions for {subject.name} ({subject_id})")
         self.setWindowModality(qt.Qt.WindowModal)
-        self.resize(700, 400)
 
         self.db = db
         self.subject_id = subject_id
@@ -497,8 +487,6 @@ class LoadSessionDialog(qt.QDialog):
     def setup(self):
         self.box_layout = qt.QVBoxLayout()
         self.setLayout(self.box_layout)
-        self.setMinimumSize(1200, 700)
-        self.setMaximumSize(1600, 1000)
 
         # ---- Sessions Table ----
         cols = [
@@ -517,11 +505,8 @@ class LoadSessionDialog(qt.QDialog):
         self.table_widget.setSelectionBehavior(qt.QAbstractItemView.SelectRows)
         self.table_widget.setEditTriggers(qt.QAbstractItemView.NoEditTriggers)
         self.table_widget.setSizePolicy(qt.QSizePolicy.Expanding, qt.QSizePolicy.Expanding)
-        self.table_widget.horizontalHeader().setDefaultSectionSize(120)
-        self.table_widget.horizontalHeader().setMinimumSectionSize(120)
         self.table_widget.horizontalHeader().setHighlightSections(False)
-        self.table_widget.horizontalHeader().setFixedHeight(30)
-        self.table_widget.verticalHeader().setDefaultSectionSize(18)
+        self.table_widget.horizontalHeader().setStretchLastSection(True)
         self.table_widget.verticalHeader().setVisible(False)
         self.table_widget.setShowGrid(False)
         self.table_widget.setFocusPolicy(qt.Qt.NoFocus)
@@ -599,7 +584,7 @@ class LoadSessionDialog(qt.QDialog):
             self.table_widget.setItem(row, 5, qt.QTableWidgetItem(session.date_created.strftime('%Y-%m-%d %H:%M')))
             self.table_widget.setItem(row, 6, qt.QTableWidgetItem(session.date_modified.strftime('%Y-%m-%d %H:%M')))
 
-            self.table_widget.setRowHeight(row, 48)
+        self.table_widget.resizeRowsToContents()
 
     @display_errors
     def append_session_to_list(self, session: "openlifu.db.session.Session") -> None:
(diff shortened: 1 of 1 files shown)


OpenLIFUData/OpenLIFUData.py
```diff
@@ -768,358 +1179,178 @@ def setup(self) -> None:
...
-        """Update whether the session loading button is enabled based on whether any subject or session is selected."""
-        if self.ui.subjectSessionView.currentIndex().isValid():
-            self.ui.sessionLoadButton.setEnabled(True)
-            if self.itemIsSession(self.ui.subjectSessionView.currentIndex()):
-                self.ui.sessionLoadButton.toolTip = 'Load the currently selected session'
-            else:
-                self.ui.sessionLoadButton.toolTip = 'Query the list of sessions for the currently selected subject'
+            # The volumes section should be automatically opened only if there
+            # are no volumes for the subject
+            subject_has_no_volumes = len(get_cur_db().get_volume_ids(self.logic.subject.id)) <= 0
```
sadhana-r, line 1244: You could also disable the sessions collapsible/buttons if the subject has no volumes. Becase you can't have/create a session without volumes. This doesn't cause any error, but the user is only prompted about not being able to create a session after they open the session dialog and try to create a session.

OpenLIFUData/OpenLIFUData.py
```diff
@@ -768,358 +1179,178 @@ def setup(self) -> None:
...
-        # Add new volume to subject
-        self.ui.addVolumeToSubjectButton.clicked.connect(self.onAddVolumeToSubjectClicked)
+        # Volumes collapsible section
+        self.ui.addVolumeButton.clicked.connect(self.on_add_volume_clicked)
 
-        # Add new session
-        self.ui.newSessionButton.clicked.connect(self.onCreateNewSessionClicked)
-        self.update_subjectLevelButtons_enabled()
+        # Session collapsible section
+        self.ui.chooseSessionButton.clicked.connect(self.on_load_session_clicked)
```
sadhana-r, line 1199:
```suggestion
self.ui.saveSessionButton.clicked.connect(self.on_save_session_clicked)
```

The button connection is missing! But I tested after adding this and the save_session logic works as expected.

OpenLIFUData/OpenLIFUData.py
```diff
@@ -768,358 +1179,178 @@ def setup(self) -> None:
...
-        # Selecting an item and clicking sessionLoadButton is equivalent to doubleclicking the item:
-        self.ui.sessionLoadButton.clicked.connect(
-            lambda : self.on_item_double_clicked(self.ui.subjectSessionView.currentIndex())
-        )
-
-        self.update_sessionLoadButton_enabled()
-        self.ui.subjectSessionView.selectionModel().currentChanged.connect(self.update_sessionLoadButton_enabled)
-
-        # Session management buttons
-        self.ui.unloadSessionButton.clicked.connect(self.onUnloadSessionClicked)
```
sadhana-r, line 817: Are we sure we don't want to keep the `Unload Session` button? Atleast in the Slicer extension, it may be helpful if a user is simultaneously interacting with session level and manually loaded data, and just wants to unload session level data. Right now, a session can only unloaded be unloaded by changing the subject or session, or they do something that triggers an unload.

OpenLIFUData/OpenLIFUData.py
```diff
@@ -278,6 +287,397 @@ def customexec_(self):
...
+
+        self.addButton = qt.QPushButton("Add Subject")
+        self.addButton.setToolTip("Add a new subject")
+        self.addButton.clicked.connect(self.on_add_subject_clicked)
+        buttonRowLayout.addWidget(self.addButton)
+
+        self.loadSubjectButton = qt.QPushButton("Load Subject")
+        self.loadSubjectButton.setToolTip("Load the selected subject")
+        self.loadSubjectButton.clicked.connect(self.onLoadSubjectClicked)
+        self.tableWidget.doubleClicked.connect(self.onLoadSubjectClicked)
```
sadhana-r, line 357: It would be nice if a singleClick highlighted the `Load Subject` button (i.e. simmilar to how the `Add Subject` button is highlighted in blue by default.

sadhana-r (commented):
Really great work with this PR!! It was a big effort to re-do the whole workflow and its so much better now. I looked over the code as much as I could and tested out a few different sessions with saving/adding data. It mostly works well! I have noted a few bug fixes and some UI/workflow suggestions that are optional for you to decide on. 

I was wondering, is there a reason why you changed the widet-level function names to be lowercase with underscores, as opposed to capitalized after the first word, as they were before? (i.e. `on_save_session_clicked` vs `onSaveSessionClicked`). I'm not sure what standard practice is but now the function naming convention is inconsistent across buttons. It might be better to use the same convention throughout?

sadhana-r:
> I can squash them if this is not good for the history.

Maybe you can squash a few of the early "experimental" commits?

ebrahimebrahim:
(I am in favor of not squashing any commits. This commit history looks pretty clean)

arhowe00:
> I was wondering, is there a reason why you changed the widet-level function names to be lowercase with underscores, as opposed to capitalized after the first word, as they were before? (i.e. `on_save_session_clicked` vs `onSaveSessionClicked`). I'm not sure what standard practice is but now the function naming convention is inconsistent across buttons. It might be better to use the same convention throughout?

Yes. I decided that with revamped functionality, it would be a subtle, non-technical way of communicating that the code is newer. In many cases I rewrote functions, so there is a rare chance it could help. I also noticed the inconsistency was already prevalent in our codebase, so as a tossup I chose snake case. I also believe it's more <u>***"pythonic"***</u>

OpenLIFUData/OpenLIFUData.py
```diff
@@ -278,6 +287,397 @@ def customexec_(self):
...
+        self.box_layout.addWidget(self.button_box)
+
+        # ---- Populate Table ----
+        self.update_sessions_list()
+
+    def update_sessions_list(self):
+        session_ids = self.db.get_session_ids(self.subject_id)
+
+        self.table_widget.clearContents()
+        self.table_widget.setRowCount(0)
```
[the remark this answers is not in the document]
arhowe00: I think this effectively clears the items, but I'm going to see if I can make new additions append rather than reset the list.

OpenLIFUData/OpenLIFUData.py
```diff
@@ -278,6 +287,397 @@ def customexec_(self):
...
+            self.table_widget.setItem(row, 2, qt.QTableWidgetItem(protocol_text))
+            self.table_widget.setItem(row, 3, qt.QTableWidgetItem(volume_text))
+            self.table_widget.setItem(row, 4, qt.QTableWidgetItem(transducer_text))
+
+            # Date created, modified
+
+            self.table_widget.setItem(row, 5, qt.QTableWidgetItem(session.date_created.strftime('%Y-%m-%d %H:%M')))
+            self.table_widget.setItem(row, 6, qt.QTableWidgetItem(session.date_modified.strftime('%Y-%m-%d %H:%M')))
+
+            self.table_widget.setRowHeight(row, 48)
```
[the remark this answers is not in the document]
arhowe00: I'm so-so on the idea of not including any styling because the changes (i.e. ones where we `set.*`) are all greppable and removable if there is a sweeping styling for all tables and widgets. I tried to set things so they look just pretty enough that if we delivered and those things never changed, we would still be satisfied with how things looked. However, given that things (especially related to dimensions) seem not even to be consistent, I'll remove all the row heights, it doesn't even look too bad:

image

Also, I know that setting things goes against the suggested approach of leaving all stylings to the end. If you guys would like (@ebrahimebrahim @sadhana-r ), I can remove every line where I set a styling choice and only leave things like number of columns, etc. However, the result would be, in my opinion, pretty ugly (especially some of the focusing choices I removed which give the false impression that a particular cell may be editable or selectable). I also worry the extent or ease to which dynamic widgets are stylable -- I assume they are probably easily stylable.

So anyway, in addition to removing the dimensions, let me know if you'd also like me to remove these decisions like the alignment of rows, expandability of columns (so text is readable), and visibility of row numbers:
```
        self.tableWidget.horizontalHeader().setDefaultAlignment(qt.Qt.AlignLeft | qt.Qt.AlignVCenter)
        self.tableWidget.setSelectionBehavior(qt.QAbstractItemView.SelectRows)
        self.tableWidget.setEditTriggers(qt.QAbstractItemView.NoEditTriggers)
        self.tableWidget.setSizePolicy(qt.QSizePolicy.Expanding, qt.QSizePolicy.Expanding)
        self.tableWidget.horizontalHeader().setHighlightSections(False)
        self.tableWidget.horizontalHeader().setStretchLastSection(True)
        self.tableWidget.verticalHeader().setVisible(False)
```

OpenLIFUData/OpenLIFUData.py
arhowe00: The "Add Subject" button is highlighted when the dialog opens, I don't know how to make it not highlighted, but this makes keyboard-only workflows convenient (Tab to change button focus + Enter to click).  I think making clicking a button highlight it first would be pretty un-natural, as it would make the `Load *` buttons the only ones in the app you have to double click. Is there a way I can make it so that opening the dialog does not highlight the "Add Subject" button?

OpenLIFUData/OpenLIFUData.py
arhowe00: Update: This has been resolved by @peterhollender and I will add an "Exit Session" button.

OpenLIFUData/OpenLIFUData.py
```diff
@@ -278,6 +287,397 @@ def customexec_(self):
...
+
+            # Protocol, volume, transducer text
+            def safe_call(func, fallback="NA"):
+                try:
+                    return func()
+                except Exception:
+                    return fallback
+
+            protocol_name = safe_call(lambda: self.db.load_protocol(session.protocol_id).name)
+            volume_name = safe_call(lambda: self.db.get_volume_info(session_id, session.volume_id)["name"])
```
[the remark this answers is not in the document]
sadhana-r: I guess no harm in leaving it in. In case there are missing fields in the database for some reason

sadhana-r:
> Adding some context: This problem has plagued every python based slicer module I've worked with. The conflict is between Slicer's convention of camel case and python's convention of snake case. I gravitate towards snake case because I'm thinking about this as python programming. But sometimes my brain switches to slicer programming mode and some camel case slips though as well. It doesn't help that the templates created by the slicer extension wizard all start with camel case.
> 
> (We can do a mass rename if we care, but no strong need to make that part of this PR)

> Yes. I decided that with revamped functionality, it would be a subtle, non-technical way of communicating that the code is newer. In many cases I rewrote functions, so there is a rare chance it could help. I also noticed the inconsistency was already prevalent in our codebase, so as a tossup I chose snake case. I also believe it's more **_`_"pythonic"_`_**

Thanks for the explanations! I think I've been trying to do camel case at the widget level, and snake case at the logic level but I'll gravitate towards snake case moving forward :)

OpenLIFUData/OpenLIFUData.py
```diff
@@ -278,6 +287,397 @@ def customexec_(self):
...
+        self.selected_session = None
+
+        self.setup()
+
+        self._enforce_user_permissions()
+
+    def setup(self):
+        self.box_layout = qt.QVBoxLayout()
+        self.setLayout(self.box_layout)
+        self.setMinimumSize(1200, 700)
```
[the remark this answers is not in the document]
arhowe00: I'm just going to remove minimums overall

arhowe00:
Okay, I've made all the changes based on resolutions above. I removed all dimensions-based resizing as it's unreliable to set those things. It does feel uncomfortably unconfigured, but it's hard to decide on these things so either I'll go with what looks good on Windows, what is in explicit feedback, or we will delay until AP efforts.

PR #398, opened 2025-06-10
Simplify workflow save/exit buttons (#388)

### LINKED ISSUE #388: Workflow controls "save" button improvement
Workflow control panel save buttons should be simply "Save" and "Exit", where the exit button prompts if you want to save.

arhowe00:
Closes #388 

- Simplifies the buttons to save and exit buttons only.
- Refactor save functionality into `save_session`.
- Confirm whether user wants to save before exiting.
- Save button no longer closes but simply calls `save_session`.

## For review

A close code review should be easy and enough.

### DIFF
--- a/OpenLIFULib/OpenLIFULib/guided_mode_util.py
+++ b/OpenLIFULib/OpenLIFULib/guided_mode_util.py
@@ -131,10 +131,22 @@ class WorkflowControls(qt.QWidget):
 
     @display_errors
     def on_save(self, clicked:bool):
+        data_module_parameter_node = slicer.util.getModuleLogic('OpenLIFUData').getParameterNode()
+
+        if data_module_parameter_node.loaded_session is None:
+            slicer.util.errorDisplay("There is no loaded session.")
+            return
+
         self.save_session()
 
     @display_errors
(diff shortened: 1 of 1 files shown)


OpenLIFULib/OpenLIFULib/guided_mode_util.py
```diff
@@ -130,19 +130,26 @@ def on_jump_ahead(self, clicked:bool):
...
+        self.save_session()
 
     @display_errors
-    def on_close(self, clicked:bool):
-        self.close_session(save=False)
+    def on_exit(self, clicked:bool):
+        if slicer.util.confirmYesNoDisplay("Do you want to save your progress before exiting?", windowTitle="Save Confirmation"):
+            self.close_session(save=True)
+        else:
+            self.close_session(save=False)
```
sadhana-r, line 141: For the `on_save` and `on_exit`, maybe add a check for a valid loaded session first, since the buttons are enabled even when there is no active session. Unlikely edge case but you could just `return` if the user clicks those buttons, without raising an error.

sadhana-r (approved):
Added minor suggestions for tooltips/error handling but otherwise looks good and works as expected!

PR #405, opened 2025-06-12
Fix new protocols not being cached properly

### LINKED ISSUE #190: Handle changes to the ID of new and cached protocols
When caching a protocol, we need to account for the case where the user changes the protocol id. When that happens, `self._cur_prototocol_id` and the id keys in `self.logic.cached_protocols` use the 'original' protocol id, but `protocol.id` will be different. This becomes an issue for new protocols, for example the check at line 376. In the case I tested, if I create a new protocol, change the ID , load a session in the data module (which causes the new protocol to get cached and reloads the list), and then select the new protocol in the drop down, it isn't recognized correctly as a new protocol i.e. the `Delete Protocol From Database` option is enabled. 

The fix might be to just replace the original ID in the list of `new_protocol_ids` and include the new one, if the id changes. I think the key in `cached_protocols` would need to be changed as well?   Let me know your thoughts. It would be good to think about any other situations where a change to the ID may break things.

_Originally posted by @sadhana-r in https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/170#discussion_r1969779533_

arhowe00:
Closes #190 

Thanks for the helpful bug report @sadhana-r !

## For review

Confirm that the bug identified in #190 is fixed

### DIFF
--- a/OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
+++ b/OpenLIFUProtocolConfig/OpenLIFUProtocolConfig.py
@@ -463,9 +463,9 @@ class OpenLIFUProtocolConfigWidget(ScriptedLoadableModuleWidget, VTKObservationM
         self.updateProtocolDisplayFromProtocol(protocol)
         self.ui.scrollArea.verticalScrollBar().setValue(0)
 
-        self._cur_protocol_id = protocol.id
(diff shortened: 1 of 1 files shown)


sadhana-r (approved):
Looks good! The issue I previously described seems to be fixed!

PR #408, opened 2025-06-13
Disable right-click context menus while using wizard

### LINKED ISSUE #244: Disable context menu on control points within the TT wizard
I like how it says deletion blocked ✋ when you try to delete a node by right clicking in the 3D view. I did notice that you can still wreak havoc by right clicking and deleting the entire point list... maybe we should open an issue to somehow disable that context menu altogether? Can be addressed later

_Originally posted by @ebrahimebrahim in https://github.com/OpenwaterHealth/SlicerOpenLIFU/pull/227#pullrequestreview-2751319296_

sadhana-r:
Closes #244 

### For Review

Confirm that context menus (via right click) are not available when using the wizard. Anywhere in the 3D view or on the control points when editing. This functionality is restored after exiting the wizard.

### DIFF
--- a/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
+++ b/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
@@ -935,6 +935,12 @@ class TransducerTrackingWizard(qt.QWizard):
 
         self._logic = slicer.util.getModuleLogic('OpenLIFUTransducerTracker')
         
+        pluginHandler = slicer.qSlicerSubjectHierarchyPluginHandler.instance()
+        pluginLogic = pluginHandler.pluginLogic()
+        self.current_allowed_context_menu_actions = pluginLogic.allowedViewContextMenuActionNames
+        # Hide all context menu items
+        pluginLogic.allowedViewContextMenuActionNames = ["NoActionsAllowed"]
+
         with BusyCursor():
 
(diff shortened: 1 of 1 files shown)


arhowe00 (approved):
This is great and it works as stated. I found a bug while testing. When the Wizard is open, you can hit "Cancel" while placing a fiducial, and this caused the following traceback.

```python
>>> 
  File "lib/Slicer-5.8/qt-scripted-modules/OpenLIFUTransducerTracker.py", line 1172, in onCancel
    self.clean_up()
  File "lib/Slicer-5.8/qt-scripted-modules/OpenLIFUTransducerTracker.py", line 1188, in clean_up
    self.clearWizardNodes()
  File "lib/Slicer-5.8/qt-scripted-modules/OpenLIFUTransducerTracker.py", line 1207, in clearWizardNodes
    self.clean_up_observers(node)
  File "lib/Slicer-5.8/qt-scripted-modules/OpenLIFUTransducerTracker.py", line 1220, in clean_up_observers
    for tag in self.node_observations.pop(node.GetID()):
KeyError: 'vtkMRMLMarkupsFiducialNode9'
```

I noticed similar behavior when hitting ESC, which likely just hits cancel.  I can open a separate issue about this or you could sneak a fix in here, it's up to you.  I'll approve so you can merge regardless!

Issue #183, opened 2025-02-19
Clearing slicer scene causes errors


ebrahimebrahim:
Clearing the slicer scene causes a bunch of errors coming from widgets trying to update themselves by accessing some parameter nodes and finding that those parameter nodes don't exist.

Can fix this by adding a check to those widget-updater-methods that the parameter node exists before we use it

PR #409, opened 2025-06-13
Again fix errors that occur during scene cleanup

ebrahimebrahim:
The same issues from #183 popped up again, and this should fix them


arhowe00 (approved):
Great job, approved for merging.

Why is it that the parameter node can be `None` if it is always initialized in `setup()`?

PR #414, opened 2025-06-16
Scale photoscan mesh from model centroid

### LINKED ISSUE #372: Mesh scaling origin issues
This is regarding the scaling slider that can be used to manually scale the mesh in the transducer tracking wizard.

The origin of mesh scaling appears to move around relative to the photoscan when the photoscan is translated.

It should instead be a fixed point in photoscan space.

sadhana-r:
Closes #372


OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -737,11 +726,41 @@ def disable_manual_registration(self):
...
+            (bounds[2] + bounds[3]) / 2,
+            (bounds[4] + bounds[5]) / 2
+        ]
+
+        center_local = [0.0,0.0,0.0]
+        transform_from_world = vtk.vtkGeneralTransform()
+        self.photoscan_to_volume_transform_node.GetTransformFromWorld(transform_from_world)
+        transform_from_world.TransformPoint(center_world,center_local )
+
+        return center_local
```
ebrahimebrahim, line 763: I am confused by the naming of things here. It returns center_local which sounds like it would be in local photoscan coordinates, but the function is called `getPhotoscanCentroidRAS` which sounds like it's in global coordinates. I think it's returning something in _world_ coordinates.. right?

sadhana-r: I'll update the naming - it is confusing. You're right its returning the photoscan coordinates in world coordinates. The local here is referring to the fact that the coordinates after transformed bythe photoscan-volume transform but its confusing. I'll update the names/docstrings to clarify.

ebrahimebrahim: much clearer thank you!

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -737,11 +726,44 @@ def disable_manual_registration(self):
...
+        self.wizard().photoscan.model_node.GetRASBounds(bounds)
+        centroid_world = [
+            (bounds[0] + bounds[1]) / 2,
+            (bounds[2] + bounds[3]) / 2,
+            (bounds[4] + bounds[5]) / 2
+        ]
+
+        centroid_transformed = [0.0,0.0,0.0]
+        transform_from_world = vtk.vtkGeneralTransform()
+        self.photoscan_to_volume_transform_node.GetTransformFromWorld(transform_from_world)
```
ebrahimebrahim, line 763: When locking and unlocking things in the transducer tracking wizard, here I sometimes get `AttributeError: 'NoneType' object has no attribute 'GetTransformFromWorld'`, so it seems `getTransformedPhotoscanCentroid` can get called while `self.photoscan_to_volume_transform_node` is `None`

ebrahimebrahim (changes requested):
It seems to still not be centered: when I drag the photoscan off to the side and drag the scaling slider, the centroid moves around

(No need to keep working on this atm unless you really want to -- we can leave the PR up!)

ebrahimebrahim:
Closing this PR as stale for now. If #372 comes back up in maintenance, then this PR can still be referenced as a starting point.

Issue #357, opened 2025-05-21
Enable manual adjustment of virtual fit result


ebrahimebrahim:
After virtual fitting, in case of failure of the virtual fitting algorithm, or disagreement of the operator with the algorithm result, there should be a way to manually override the automatically virtually fitted transducer pose.

PR #422, opened 2025-06-18
Clean up virtual fit button logic (#357)

ebrahimebrahim:
The button enabled/disabled logic was quite scattered and it had become impossible to fix certain bugs where buttons were enabled when they shouldn't be.

This moves the button enabled/disabled/tooltip/text adjustements all to one function `updateVirtualfitButtons`.


OpenLIFUPrePlanning/OpenLIFUPrePlanning.py
```diff
@@ -534,40 +534,72 @@ def onLockClicked(self):
...
-
     def updateInputOptions(self):
         """Update the algorithm input options"""
 
         self._input_update_in_progress = True
         self.algorithm_input_widget.update()
         self._input_update_in_progress = False  # Prevents repeated function calls due to combo box index changed signals
 
-        self.updateVirtualfitButtonEnabled()
+        self.updateVirtualfitButtons()
```
sadhana-r, line 544: I don't think this is needed here since it is called in `self.updateVirtualFitResultsTable`. I tested with this commented out and it didn't seem to break anything.

sadhana-r (approved):
Thank you so much for figuring this out! Its much cleaner now and easier to follow the enabledness logic. 
I tested out a few different scenarios and didn't notice any issues!

1. No Target
2. Multiple targets and switching back and forth with different selections
3. Modifying an approved VF result
4. Modifying an approved TT result
5. Approving/revoking approval using the approve button.

PR #439, opened 2025-06-27
Hide pnp rendering from photoscan preview and tracking wizard

### LINKED ISSUE #430: Peak Negative Pressure replaces Photoscan in Preview and Tracking
When you have "Render Peak Negative Pressure" checked, and generate a new photoscan, the "Preview Photoscan" button pulls up this: 

Image

Also shows up in the transducer tracking wizard, rendering it unusable.

If I hide the peak negative pressure, the photoscan comes back and works fine.

sadhana-r:
Closes #430 

The issue occured when a photoscan was previewed before a sonication solution was computed and rendered. Subsequent previews of the same photoscan incorrectly displayed the PNP rendering. 
By default, the pnp rendering display node includes the view node associated with the preview/tracking window.  This PR updates the way display nodes are hidden from certain views, by explicitly accounting for the case where a display node is associated with a non-empty list of view node IDs. 

For review:
- Confirm that the issue is fixed when following the sequence of actions described above.
- The second commit's fix is not essential for this issue, but I thought it might be better to exlude the wizard view nodes when `view_node.SetRaycastTechnique(slicer.vtkMRMLViewNode.MaximumIntensityProjection)` is called just incase..

### DIFF
--- a/OpenLIFULib/OpenLIFULib/transducer_tracking_wizard_utils.py
+++ b/OpenLIFULib/OpenLIFULib/transducer_tracking_wizard_utils.py
@@ -60,6 +60,7 @@ def create_threeD_photoscan_view_node(photoscan_id: str):
         viewNode.SetLayoutColor(layoutColor)
         viewNode.SetName(f'view-{photoscan_id}')
         viewNode.SetAndObserveParentLayoutNodeID(viewOwnerNode.GetID())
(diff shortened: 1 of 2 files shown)


ebrahimebrahim (approved):
It worked for me, and thanks for the extra clean up as well with the maximum intensity projection!

PR #450, opened 2025-06-30
Improve live transducer hardware connectivity and feedback

### LINKED ISSUE #449: Integration gaps with live transducer hardware
Before the recent branch work, the hardware integration in main was incomplete and caused several usability issues that were not noticed until usage on live hardware.

- Stopping a sonication once started was not fully functional (only Abort was functional).
- Run and Abort buttons worked, but status feedback was not present.
- Device connection and real-time updates were only triggered on UI refresh, not via live callback. UI did not properly reflect oscillator state or sonication completion

This issue links the hardware-related fixes; a separate issue will track the asyncio refactor concerns.

arhowe00:
Closes #449 

This PR (most commits were fully authored by @duvitech-llc / @georgevigelette ) addresses several gaps in the main branch related to hardware integration and runtime status handling:

- Fix real-time data callbacks to process hardware status updates during sonication, which were previously missing.
> Note: `on_lifu_data_received` has an implementation that forces some data processing to happen from within the sonication control module. For instance, in addition to the pulse train percent for progress bar updates, we also receive `STOPPED` via the interface signal and then set the hardware status of the interface. This is not preferred, as it exposes Slicer to processing that should happen below the interface abstraction. We should just be able to poll the status via `get_status` and not have to call `set_status` externally.
- Implemented proper stop behavior at the end of sonication, ensuring the interface halts transmission and updates the UI accordingly.
- Restored connect/disconnect signals and UI updates, which were only firing on manual UI interaction before.
- Established UI progress and connection indicators based on actual hardware signals (e.g., trigger pulses detected on oscilloscope now propagate to the UI, and voltage rails are set correctly on “Run”).
- **Added a warning for re-initializing the interface in test mode. Lots of the logic around hardware mocking was removed and not replaced in some of the changes, so it is unlikely to work.** 

## For review:

@peterhollender , make sure that if you have a transducer device that things work as expected. @ebrahimebrahim or @sadhana-r it would be good to have a pass over the code as well.

## Acknowledgement

Most commits in this PR were authored by @duvitech-llc  / @georgevigelette

### DIFF
--- a/OpenLIFULib/OpenLIFULib/Resources/python-requirements.txt
+++ b/OpenLIFULib/OpenLIFULib/Resources/python-requirements.txt
@@ -1,2 +1,2 @@
-git+https://github.com/OpenwaterHealth/OpenLIFU-python.git@a5f725a11bde2dea657343bbf7554f7dd113ebf3
+openlifu==v0.8.0
 bcrypt
--- a/OpenLIFULib/OpenLIFULib/Resources/python-requirements.txt
+++ b/OpenLIFULib/OpenLIFULib/Resources/python-requirements.txt
@@ -1,2 +1,2 @@
-openlifu==v0.7.1
+git+https://github.com/OpenwaterHealth/OpenLIFU-python.git@a5f725a11bde2dea657343bbf7554f7dd113ebf3
 bcrypt
--- a/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
+++ b/OpenLIFUSonicationControl/OpenLIFUSonicationControl.py
@@ -438,6 +438,7 @@ class OpenLIFUSonicationControlWidget(ScriptedLoadableModuleWidget, VTKObservati
             returncode, run_parameters = runCompleteDialog.customexec_()
             if returncode:
                 self.logic.create_openlifu_run(run_parameters)
+        self.updateAllButtonsEnabled()
 
     @display_errors
     def onDeviceConnected(self):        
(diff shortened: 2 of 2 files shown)


ebrahimebrahim:
Does this work have a connection to the work that is being done at https://github.com/OpenwaterHealth/OpenLIFU-python/pull/352 ?

georgevigelette:
> > Does this work have a connection to the work that is being done at [OpenwaterHealth/OpenLIFU-python#352](https://github.com/OpenwaterHealth/OpenLIFU-python/pull/352) ?
> > (I see the OpenLIFU python hash used here is from the current main branch)
> 
> That would depend on if @georgevigelette was testing it on that python branch or intended for this to be coupled with those changes. @peterhollender

Yes I am testing with  [OpenwaterHealth/OpenLIFU-python#352] so they are linked though I believe control of the hardware would still work with previous versions.

ebrahimebrahim (approved):

georgevigelette:
Looks good from my end.

ebrahimebrahim:
Here is an extension package for hardware testing of this PR  
[33241-linux-amd64-SlicerOpenLIFU-git0f03575-2025-06-30.tar.gz](https://github.com/user-attachments/files/20987414/33241-linux-amd64-SlicerOpenLIFU-git0f03575-2025-06-30.tar.gz)

arhowe00:
I merged this. There wasn't as much scrutiny on the code side, but we should probably open some issues about restoring the mock interface to the hardware, as without it this module can't be run without the transducer.

PR #460, opened 2025-07-25
Clean up scene when deleting solution analysis on target modified

### LINKED ISSUE #438: Hide pnp rendering when a solution is deleted
When a solution is deleted, the p_min and p_intensity images remain in the scene. I think that's deliberate and nice to have for the extension, but if the render pnp option is checked, the beam is still rendered in the scene. If you then go and re-compute a solution (with the new/modified transducer position), two renderings are now visible in the scene. I think for the custom app, its important to hide the associated rendering when a solution is deleted. Also loading a different subject/session doesn't clear the solution analysis images, and so they remain rendered in the scene.

sadhana-r:
Closes #438 

Changes the default setting to clean up the scene when a solution is deleted on target modfication. 

For Review:

- With the example subject/sesssion - I get the notify message multiple times when a solution is deleted after I move the target after computing a solution. I didn't see this happen with the Soren example. Could you confirm if this happens for you?

### DIFF
--- a/OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
+++ b/OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
@@ -344,7 +344,7 @@ class OpenLIFUSonicationPlannerWidget(ScriptedLoadableModuleWidget, VTKObservati
         """
         data_logic : "OpenLIFUDataLogic" = slicer.util.getModuleLogic('OpenLIFUData')
         if self.logic.solution_analysis_exists():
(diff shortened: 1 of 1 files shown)


ebrahimebrahim:
> With the example subject/sesssion - I get the notify message multiple times when a solution is deleted after I move the target after computing a solution. I didn't see this happen with the Soren example. Could you confirm if this happens for you?

I do get multiple "Solution deleted: The target was modified." notifications, but I think it's not about which session but rather it seems to happen when you create a solution, get it deleted, and then create another solution. Opening a separate issue for that as it's unrelated to this PR.

PR #468, opened 2025-08-01
Fix transducer tracking outside of session context

### LINKED ISSUE #457: Transducer Tracking doesn't work outside Session context
I've used the Data module to manually load Transducer, Volume, Protocol, and Photoscan objects into the scene. I can Preview the Photoscan and run transducer tracking, but it throws errors related to approvals, and doesn't complete correctly. 

Here is the error text that (non-blocking) appears during the tracking wizard:

```
  File "C:/Users/pjh7/AppData/Local/slicer.org/Slicer5.8.1/slicer.org/Extensions-33241/SlicerOpenLIFU/lib/Slicer-5.8/qt-scripted-modules/OpenLIFUTransducerTracker.py", line 1158, in onFinish
    self.photoscan_to_volume_transform_node, self.transducer_to_volume_transform_node = self._logic.add_transducer_tracking_result(
  File "C:/Users/pjh7/AppData/Local/slicer.org/Slicer5.8.1/slicer.org/Extensions-33241/SlicerOpenLIFU/lib/Slicer-5.8/qt-scripted-modules/OpenLIFUTransducerTracker.py", line 2771, in add_transducer_tracking_result
    approved_photoscan_id = session.get_transducer_tracking_approvals()
AttributeError: 'NoneType' object has no attribute 'get_transducer_tracking_approvals'
  File "C:/Users/pjh7/AppData/Local/slicer.org/Slicer5.8.1/slicer.org/Extensions-33241/SlicerOpenLIFU/lib/Slicer-5.8/qt-scripted-modules/OpenLIFUTransducerTracker.py", line 2075, in onRunTrackingClicked
    raise RuntimeError("Transducer tracking wizard was completed without generating valid transducer tracking transforms")
RuntimeError: Transducer tracking wizard was completed without generating valid transducer tracking transforms
```

This check and updating of the approvals probably shouldn't happen if there isn't a session loaded.

sadhana-r:
Closes #457 

### For Review

Confirm that tt workflow functions without errors outside of the session context. You can do this by loading the individual protocol/transducer/volume and photoscan in the data module (I used the objects affiliated with Soren). 

Similar to a session-based workflow, during the manual workflow, if the user loads two photoscans, tracking can only be approved for one photoscan at a time. 

Note: While testing with multiple photoscans, I discovered #469. This is not yet resolved.

### DIFF
--- a/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
+++ b/OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
@@ -1083,7 +1083,8 @@ class TransducerTrackingWizard(qt.QWizard):
             self.photoscan.model_node.GetDisplayNode().SetVisibility(False)
             self.transducer_surface.GetDisplayNode().SetVisibility(False)
             self.transducer_body.GetDisplayNode().SetVisibility(False)
-            self.transducer.cloned_virtual_fit_model.GetDisplayNode().SetVisibility(False)
+            if self.transducer.cloned_virtual_fit_model:
+                self.transducer.cloned_virtual_fit_model.GetDisplayNode().SetVisibility(False)
 
             if self.photoscanMarkupPage.facial_landmarks_fiducial_node:
                 self.photoscanMarkupPage.facial_landmarks_fiducial_node.GetDisplayNode().SetVisibility(False)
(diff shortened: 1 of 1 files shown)


ebrahimebrahim (approved):
This worked really well!!

PR #474, opened 2025-08-19
Speed up virtual fitting

ebrahimebrahim:
This is work towards https://github.com/OpenwaterHealth/OpenLIFU-python/issues/367

---

- Incorporate embree speedups from the work on https://github.com/OpenwaterHealth/OpenLIFU-python/issues/367
- Cap BLAS threads, which seems to help the  pos search when not using embree. I tried this on ubuntu 22.04.

---

Before merging
- [x] replace  hash by versioned OpenLIFU-python requirement

### DIFF
--- a/OpenLIFULib/OpenLIFULib/lazyimport.py
+++ b/OpenLIFULib/OpenLIFULib/lazyimport.py
@@ -18,7 +18,12 @@ def install_python_requirements() -> None:
 
 def python_requirements_exist() -> bool:
     """Check and return whether python requirements are installed."""
-    return importlib.util.find_spec('openlifu') is not None
+    try:
+        import threadpoolctl
+        import bcrypt
+    except ModuleNotFoundError:
+        return False
+    return importlib.util.find_spec('openlifu') is not None # openlifu import causes a delay so we check for it without actually importing yet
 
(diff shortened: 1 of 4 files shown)


OpenLIFULib/OpenLIFULib/lazyimport.py
```diff
@@ -18,7 +18,12 @@ def install_python_requirements() -> None:
 
 def python_requirements_exist() -> bool:
     """Check and return whether python requirements are installed."""
-    return importlib.util.find_spec('openlifu') is not None
+    try:
+        import threadpoolctl
+        import bcrypt
+    except ModuleNotFoundError:
+        return False
+    return importlib.util.find_spec('openlifu') is not None # openlifu import causes a delay so we check for it without actually importing yet
```
ebrahimebrahim, line 26: we should have added this `try` long ago when bcrypt was added to `python-requirements.txt`.

so here I include bcrypt in addition to threadpoolctl

ebrahimebrahim:
@sadhana-r for review: Please test that virtual fitting works for you on windows. Also, can you let me know roughly how fast does it run for you? A couple of seconds? Or minutes?

sadhana-r (approved):

PR #475, opened 2025-08-22
Add peak negative pressure window level / opacity threshold controls

### LINKED ISSUE #433: PNP/Intensity Color Controls
When rendering Peak Negative Pressure, it would be good to have some controls for setting the color and opacity bounds. These can already be controlled through the `Volumes` and `Volume Rendering` modules now, but some of those basic controls should be available on the Sonication Planning page. 

## Color 
- Double Handle (Manual Min/Max) Slider
- Slider Minimum = 0
- Slider Maximum = Target Pressure*1.5
- Default to (0, Target Pressure)
- Sets Window/Level

## Opacity
~- Double Handle (Manual Min/Max) Slider~
~- Slider Minimum = 0~
~- Slider Maximum = Target Pressure~
~- Default to (0.1•Target Pressure, 0.5•Target Pressure)~
~- Bounds set the 0 and 100% opacity limits in volume rendering, as well as the lower bound for thresholding of slices~
- Single Handle (Manual) Slider
- Slider Bounds: Same as Color
- Default to 0
- Sets the min of Threshold (max is always the max value of the array to prevent clipping of large values)

arhowe00:
Closes #433 

The automatic rendering of the Peak Negative Pressure (p_min) volume can obscure the true pressure field distribution due to suboptimal defaults.

This change allows to manually defining the window/level for the color and opacity. This allows for a more precise and accurate visualization of the underlying simulation data.

### DIFF
--- a/OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
+++ b/OpenLIFUSonicationPlanner/OpenLIFUSonicationPlanner.py
@@ -419,12 +419,9 @@ class OpenLIFUSonicationPlannerWidget(ScriptedLoadableModuleWidget, VTKObservati
 
     def onPnpColorSliderChanged(self, new_min_val: float, new_max_val: float) -> None:
         """Called when the PNP color slider values are changed."""
-        slicer.app.processEvents() # Ensures slider remains responsive
-
         data_parameter_node = get_openlifu_data_parameter_node()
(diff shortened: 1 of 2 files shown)


ebrahimebrahim (changes requested):
I am getting the sliders to affect the slice views but now they are no longer affecting the volume rendering:

Peek 2025-08-25 15-28

ebrahimebrahim (approved):
Works well and code looks good. Thank you!

PR #492, opened 2025-11-26
Update for v0.5.10 of the 3D Scanner App

### LINKED ISSUE #489: Update for v0.5.10 of the 3D Scanner App
We recently have been using v0.5.10 of the 3D Scanner App, and when we try to transfer the images over to the application, an error message pops up even though the reference number is the same. It works fine with the older version of the photogrammetry app, but not with the newer version (we've changed the .jpeg to .jpg for the image files, but it doesn't seem to help), is this something we should be looking at on the photogrammetry app side? 

Image

arhowe00:
Closes #489 

Changes the Android image pull to optionally pull files related to the photoscan mesh and forgo the photocollection import. This also changes the path where adb searches to [new path specified in #489](https://github.com/OpenwaterHealth/SlicerOpenLIFU/issues/489#issuecomment-3564006638).

## For review

Please do a line-by-line pass over `on_transfer_photocollection_from_android_device_clicked`, specifically the branch taken when a photoscan is present. I tried to replicate all state change from `onStartPhotoscanGenerationButtonClicked` which was the previous way to generate photoscans. Since we don't have a mocking framework for the Android and I don't have a current version of the app with cloud capabilities, it would be great if the team at NVP could give this branch a try.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,6 @@ add it to your `PATH`.
 
 ## Meshroom Setup (Optional)
 
-This application is designed to work with the [OpenLIFU 3D Scanner Android app](https://github.com/OpenwaterHealth/OpenLIFU-3DScanner). With credits purchased in the app, computationally intensive tasks such as photogrammetric mesh reconstruction are performed in the cloud, eliminating the need for local Meshroom installation.
+This application is designed to work with the [OpenLIFU 3D Scanner Android app](https://github.com/OpenwaterHealth/OpenLIFU-3DScanner). With credits in the app, computationally intensive tasks such as photogrammetric mesh reconstruction are performed in the cloud, eliminating the need for local Meshroom installation.
 
 If you prefer to perform mesh reconstruction locally instead of using cloud processing, you will need to install Meshroom and add it to your system PATH. Follow the instructions [here](https://github.com/OpenwaterHealth/OpenLIFU-python?tab=readme-ov-file#installing-meshroom) to download and configure Meshroom for local photoscan generation.


arhowe00:
I can't tag @ipyanin

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -2666,39 +2711,45 @@ def __init__(self) -> None:
     def getParameterNode(self):
         return OpenLIFUTransducerTrackerParameterNode(super().getParameterNode())
 
-    def pull_photocollection_from_android(self, reference_number: str) -> List[str]:
+    def pull_photocollection_from_android(self, reference_number: str) -> tuple[str, List[str]]:
```
ebrahimebrahim: Since it can pull either a photoscan or a photocollection, this function should be renamed. Maybe something like `pull_photo_data_from_android`

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -2061,47 +2061,92 @@ def on_transfer_photocollection_from_android_device_clicked(self, checked:bool):
...
             )
-            return
 
-        # Below is done twice because session_photocollections stored in the
-        # data parameter node is not the same as those stored in
-        # SlicerOpenLIFUSession and both must be updated
-        if photocollection_dict["reference_number"] not in data_parameter_node.session_photocollections:
-            data_parameter_node.session_photocollections.append(photocollection_dict["reference_number"]) # automatically load as well
-        data_logic.update_photocollections_affiliated_with_loaded_session()
+        else:
```
ebrahimebrahim: suggest explicit `elif data_type=='photos'` here, ending with an `else:` that raises an error about an unhandled case or unexpected `data_type`.

OpenLIFUTransducerTracker/OpenLIFUTransducerTracker.py
```diff
@@ -2061,47 +2061,92 @@ def on_transfer_photocollection_from_android_device_clicked(self, checked:bool):
...
 
-        if not imported_filepaths:
-            slicer.util.errorDisplay(
-                text="Error importing files: No files found or import failed.",
-                windowTitle="Import Error"
+            self.updateInputOptions()
+            self.updateWorkflowControls()
+
+            slicer.util.infoDisplay(
+                text=f"Photoscan '{cur_reference_number}' has been successfully imported from the Android device.",
```
ebrahimebrahim: Here we should inform the user that since they now have the photoscan they don't need to proceed to "Generate a photoscan"

image


From the UI it looks like you have to now go to Step 2: Generate a photoscan.

The info message here could have additional text like "You do not need to generate a photoscan locally."

ebrahimebrahim (approved):
Great!

PR #493, opened 2025-12-01
Enable adding DICOM volumes

### LINKED ISSUE #490: Enable adding DICOM
Update SlicerOpenLIFU UI to support the option of adding volume from DICOM; see https://github.com/OpenwaterHealth/OpenLIFU-python/issues/396 for context on OpenLIFU-python supporting this.

Needs https://github.com/OpenwaterHealth/OpenLIFU-python/pull/398 to be merged.

arhowe00:
Closes #490 

This adds a file browser that allows selecting either individual volume files or DICOM directories, with basic validation of volume file types. For the DICOM directory case, there is logic within openlifu to check whether the directory is valid DICOM, and exceptions will be propagated here through the `display_errors` decorator. Was offloading this part to openlifu the correct approach?

## For review

Code review should be sufficient.

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -221,8 +221,6 @@ class AddNewVolumeDialog(qt.QDialog):
         self.setLayout(formLayout)
 
         self.volumeFilePath = ctk.ctkPathLineEdit()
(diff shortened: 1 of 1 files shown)


ebrahimebrahim:
(also I appreciate the commit message body explaining how the problem was solved!)

PR #517, opened 2026-01-16
Make texture files optional for loading and adding photoscans (#512)

### LINKED ISSUE #512: Ensure texture files are not required loading a photoscan
It was reported that an exr file did have to be provided. It should be okay to omit texture data.

sadhana-r:
Depends on https://github.com/OpenwaterHealth/OpenLIFU-python/pull/413
Closes #512 

Before merging this:

- [x]  Merge https://github.com/OpenwaterHealth/OpenLIFU-python/pull/413
- [x]  Update the python-requirements.txt openlifu commit hash

### For review

- Update the python requirements (```slicer.util.pip_install('git+https://github.com/OpenwaterHealth/OpenLIFU-python.git@98306bd455c020e094d2850c7345f04b30ce8d0b')``` )
- Confirm that you are able to load and view a photoscan with and without a texture file. Photoscans can be loaded manually via the Data module, as part of a session (upon Preview) or when added in the Transducer Localization module.

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -835,7 +835,7 @@ class LoadPhotoscanDialog(qt.QDialog):
             self.photoscanTextureFilePath = ctk.ctkPathLineEdit()
             self.photoscanTextureFilePath.filters = ctk.ctkPathLineEdit.Files
             # Allowable photoscan filetypes
-            self.photoscan_texture_extensions = ("Photoscan Texture" + " (*.jpg *. *.png *.tiff *.exr);;" +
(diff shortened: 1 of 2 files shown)


OpenLIFUData/OpenLIFUData.py
```diff
@@ -838,20 +838,21 @@ def updateDialog(self):
             self.photoscan_texture_extensions = ("Photoscan Texture" + " (*.jpg *. *.png *.tiff *.exr);;" +
```
ebrahimebrahim: (this is not part of this PR but noticed there's a stray "*." -- was that supposed to be "*.jpeg"? if so it could fixed in this PR)

PR #521, opened 2026-01-22
Add backward compatibility for legacy photogrammetry app v3.0 (#501)

### LINKED ISSUE #501: Backwards Compatibility
I noticed that OpenLIFU v1.5.2 isn't compatible with the older version of the photogrammetry app (v3.0) and is solely compatible with the newest version (v0.5.10, via NVP). I was wondering if we can have it be compatible with both the older and newer versions? Let me know what you think, since we're still in progress with updating/working on the newest version of the photogrammetry application.

arhowe00:
Closes #501 

Falls back to `/sdcard/DCIM/Camera/{ref}_*` when the new app location (`/sdcard/OpenLIFU-3DScanner/{ref}/`) is not found before giving error messages.

### DIFF
--- a/OpenLIFUTransducerLocalization/OpenLIFUTransducerLocalization.py
+++ b/OpenLIFUTransducerLocalization/OpenLIFUTransducerLocalization.py
@@ -2790,7 +2790,7 @@ class OpenLIFUTransducerLocalizationLogic(ScriptedLoadableModuleLogic):
                 raise FileNotFoundError(
                     f"No photos found with reference number '{reference_number}' "
                     f"on the android device. Please make sure you typed the correct "
-                    f"reference number into the 3D Open Water app."
+                    f"reference number into the OpenLIFU Android app."
                 )
 
         files = [f for f in result.stdout.strip().split('\n') if f]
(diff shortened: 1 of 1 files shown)


OpenLIFUTransducerLocalization/OpenLIFUTransducerLocalization.py
```diff
@@ -2774,34 +2842,26 @@ def pull_photo_data_from_android(self, reference_number: str) -> tuple[str, List
...
+                return self._pull_photos_from_legacy_location(reference_number, temp_path)
             else:
-                raise FileNotFoundError(
-                    f"No photos found with reference number '{reference_number}' "
-                    f"on the android device. Please make sure you typed the correct "
-                    f"reference number into the 3D Open Water app."
-                    )
+                # New app's base directory exists but this reference number wasn't found;
+                # fall back to legacy location
+                return self._pull_photos_from_legacy_location(reference_number, temp_path)
```
ebrahimebrahim: The if and else branches both do the same thing -- let's simplify this

ebrahimebrahim (approved):
Thank you!

PR #529, opened 2026-02-12
Add refresh button to update list of db photoscans 

### LINKED ISSUE #528: Add "refresh" button to Photoscans
We need to add a "refresh" button to the transducer tracking page that will make refresh the list of available photoscans. The cloud syncing functionality will push a new photoscan into the filesystem correctly, so we just need a way to refresh the list.

Image

Behind the scene, I'm assuming we just call

`data_logic.update_photoscans_affiliated_with_loaded_session()`

sadhana-r:
Closes #528

This was a little more complicated than expected due to the way the input widget is generated i.e. its generated at runtime using the `OpenLIFUAlgorithmInputWidget` class. 

## For Review

Look over the code to confirm that my approach makes sense. I tested it out by manully adding a photoscan to the database and confirmed that it shows up in the input list when I hit refresh.

### DIFF
--- a/OpenLIFUTransducerLocalization/OpenLIFUTransducerLocalization.py
+++ b/OpenLIFUTransducerLocalization/OpenLIFUTransducerLocalization.py
@@ -2042,6 +2042,7 @@ class OpenLIFUTransducerLocalizationWidget(ScriptedLoadableModuleWidget, VTKObse
         self.addObserver(slicer.mrmlScene, slicer.vtkMRMLScene.NodeRemovedEvent, self.onNodeRemoved)
 
         # ---- Photoscan generation connections ----
+        self.ui.referenceNumberRefreshButton.setIcon(slicer.app.style().standardIcon(qt.QStyle.SP_BrowserReload))
         self.ui.referenceNumberRefreshButton.clicked.connect(self.on_reference_number_refresh_clicked)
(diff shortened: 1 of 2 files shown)


OpenLIFUTransducerLocalization/OpenLIFUTransducerLocalization.py
```diff
@@ -2062,6 +2062,7 @@ def setup(self) -> None:
         replace_widget(self.ui.algorithmInputWidgetPlaceholder, self.algorithm_input_widget, self.ui)
         self.updateInputOptions()
         self.algorithm_input_widget.connect_combobox_indexchanged_signal(self.updateInputRelatedWidgets)
+        self.algorithm_input_widget.inputs_dict["Photoscan"].refresh_button.clicked.connect(self.refreshPhotoscanList)
```
ebrahimebrahim: Instead of this `connect` line here that reaches deeeep into algorithm input widget, we could create a 

```
OpenLIFUAlgorithmInputWidget.connect_refresh_signal(input_name: str, function_call: Callable)
```

similar to the existing 

```
OpenLIFUAlgorithmInputWidget.connect_combobox_indexchanged_signal(self, function_call: Callable)
```

ebrahimebrahim (approved):
Thanks for this!

I think I understand why it had to be done this way.

I have two minor suggestions but only if they don't take too much time, so feel free to address what you agree with and then merge

PR #561, opened 2026-03-13
Add openlifu version check and installation button in login module

### LINKED ISSUE #544: Add openlifu Python library installer/updater to Login module
With #541 and #542 adding "application setup" convenience features to the Login module, I think it also makes sense to add something for the `openlifu` Python library.

Similarly an indicator that checks whether `openlifu` is available, with a button that can be used to kick off an install if not.

However here we can easily do better than the button that exists in the Home module: _check also whether the available openlifu version matches the required version in python-requirements.txt_. The indicator can let the user know if they need to update their `openlifu`. Currently the need to update `openlifu` is hidden and because our checks only check for its presence and not for a version requirement -- a bad user experience that often leads to cryptic error messages.

sadhana-r:
Closes #544 

- The install python requirements button is always enabled. 
- The button text updates to indicate whether openlifu is installed, and what version is required
- If openlifu is not installed, entering the login module will prompt the user automatically to install openlifu since it is required for initializing default user accounts.
- If openlifu is installed, but is not the latest version, the user is not automatically prompted to update openlifu, but the button and status icon indicate that the latest version is not installed.

Todo:

- [x] Confirm that the changes to the user account settings don't break things in the custom app.

### DIFF
--- a/OpenLIFUHome/OpenLIFUHome.py
+++ b/OpenLIFUHome/OpenLIFUHome.py
@@ -20,10 +20,6 @@ from OpenLIFULib.util import (
 )
 
 from OpenLIFULib.guided_mode_util import set_guided_mode_state, Workflow
-from OpenLIFULib.lazyimport import (
-    check_and_install_python_requirements,
-    python_requirements_exist,
-)
 
 from OpenLIFUCloudSync import getCloudSyncLogic
 #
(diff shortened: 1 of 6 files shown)


ebrahimebrahim (commented):
This looks great to me! It works very well 

I agree that we should check that nothing broke int he custom app before merging.

BTW did the login module always import openlifu on entry? That creates a situation where you can never validly get to the install button in the first place if you didn't already have some version of openlifu installed.

ebrahimebrahim:
> The claude suggested fix works well for handling the default users, but I realized the other issue is that `OpenLIFUSonicationControlLogic`'s `init` function calls `openlifu_lz()` (for lifu interface related functions) and the login module currently initializes all of the moduels in ` self.cacheAllLoginRelatedWidgets()`.
> 
> So we need to either make changes to the sonication control module to avoid openlifu dependency at startup or the login module should not cache the sonication control module if openlifu isn't installed. perhaps better to do this in a separate PR?

Yes, agreed, a separate PR. We can put it off indefinitely for now ... it is good that we have a clear explanation  of the problem and possible solutions recorded here

ebrahimebrahim (changes requested):
Looks great! One small change and we should be good to merge

ebrahimebrahim (approved):

PR #568, opened 2026-05-28
Add starter database initialization options

### LINKED ISSUE #543: Add the option to start with a database that is pre-loaded with transducers
The initial database currently is an [`empty_db`](https://github.com/OpenwaterHealth/SlicerOpenLIFU/tree/1477a659ada4b92bfeb0aa606bbf3a2c8bbd9a04/OpenLIFUDatabase/Resources/openlifu-database/empty_db), but we should also add an option to initialize to a more complete starting db that has transducers in it.

It could be called `starter_db` or something like that

When prompted as to whether you want to make a empty db you should additionally have the choice to go with starter db instead. Currently we provide the option to start with `empty_db` here:

https://github.com/OpenwaterHealth/SlicerOpenLIFU/blob/1477a659ada4b92bfeb0aa606bbf3a2c8bbd9a04/OpenLIFUDatabase/OpenLIFUDatabase.py#L174-L182

The starter db has large files, because there are transducer models, so we would not commit it to the repo. Instead we can pull it from some special data repo that openwater has made, or use git lfs. Discuss with Openwater where the data should be and how we should pull it from the extension. One option that was considered also is to leverage Slicer's SampleData module.

ebrahimebrahim:
Closes #543 

- We use the [openlifu-sample-database](https://github.com/OpenwaterHealth/openlifu-sample-database) repo with its git LFS tracked data to download sample data.
- A **sample database** is obtained from an openlifu-sample-database tag which is hardcoded in sample_data.py: it should always be set to be a tag that consists of data that is compatible with the version of openlifu-python that SlicerOpenLIFU uses.
- A **starter** database is another hard-coded tag, which we should create as a version of the sample database tag that has all the subjects stripped out.

image

### For review

Try it out: In the database module, put in a writable path where there isn't already a database. Try connecting and then try "Use Sample Data". It is a large download.

Try again with "Create Starter datatabase". It is a medium sized download.

### DIFF
--- a/OpenLIFULib/OpenLIFULib/sample_data.py
+++ b/OpenLIFULib/OpenLIFULib/sample_data.py
@@ -1,0 +1,323 @@
+# Standard library imports
+import os
+import tempfile
+import urllib.request
+import zipfile
+from pathlib import Path
+from typing import Callable, List, Optional
+
+SAMPLE_DATABASE_REPOSITORY_URL = "https://github.com/OpenwaterHealth/openlifu-sample-database"
+SAMPLE_DATABASE_TAG = "openlifu-v0.20.0"
+STARTER_DATABASE_TAG = "openlifu-v0.20.0-no-subjects"
+
+
+def archive_url_for_tag(tag: str) -> str:
+    return f"{SAMPLE_DATABASE_REPOSITORY_URL}/archive/refs/tags/{tag}.zip"
+
+
+def readme_url_for_tag(tag: str) -> str:
+    return f"{SAMPLE_DATABASE_REPOSITORY_URL}/blob/{tag}/README.md"
+
+
+SAMPLE_DATABASE_ARCHIVE_URL = archive_url_for_tag(SAMPLE_DATABASE_TAG)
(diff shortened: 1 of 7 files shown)


OpenLIFULib/OpenLIFULib/sample_data.py
```diff
@@ -0,0 +1,323 @@
...
+import os
+import tempfile
+import urllib.request
+import zipfile
+from pathlib import Path
+from typing import Callable, List, Optional
+
+SAMPLE_DATABASE_REPOSITORY_URL = "https://github.com/OpenwaterHealth/openlifu-sample-database"
+SAMPLE_DATABASE_TAG = "openlifu-v0.20.0"
+STARTER_DATABASE_TAG = "openlifu-v0.20.0-no-subjects"
```
ebrahimebrahim, line 11: These tags should be updated as the sample data evolves. We should consider updating them roughly whenever the openlifu-python version is updated, to continue providing compatible sample/starter data.

sadhana-r (approved):
Tested both versions and it works well! It might be helpful to include some info on the expected size of the sample database before download, since its quite a large download. Incase there are any storage considerations for the user.

PR #573, opened 2026-06-08
Fix ADB PATH setting on windows machines with missing registry key

sadhana-r:
Closes #572


ebrahimebrahim (approved):
Tested on windows that this works. I don't have a missing registry key issue so I could not confirm that it fixes the original issue, but at least I can confirm it doesn't break the existing functionality. TY!

PR #576, opened 2026-06-15
Replace lazy imports with enter-time dependency checks

ebrahimebrahim:
- Rename `lazyimport.py` to `dependency_utils.py` and remove the `*_lz()` import helpers.
- Add `ensure_python_requirements_for_module_enter()` and use it from module `enter()` methods to run dependency checks/install prompts before OpenLIFU-dependent behavior.
- Warn once per session when the installed openlifu version does not match the pinned requirements version.
  - (In a separate commit, also fix a bug in the version checker where it trips up over the brackets in openlifu[app])
- Replace `*_lz()` call sites with direct local imports for openlifu, openlifu_sdk, and related packages.
- Defer setup-time OpenLIFU-dependent initialization where needed:
  - `OpenLIFUProtocolConfig`: build dynamic protocol editor widgets after dependencies are available on module enter.
  - `OpenLIFUSonicationControl`: create LIFUInterface after dependencies are available on module enter.
  - `OpenLIFULogin`: initialize default users after dependencies are available instead of during setup.
  - `OpenLIFUCloudSync`: move cloud SDK/request imports into dependency-checked runtime paths.

I tested the behavior of different modules when openlifu is installed, isn't installed, or is installed at the wrong version.

@sadhana-r For review either a quick test of workflow or a look over the code is sufficient, no need for both. I've tested behavior and looked over the code carefully.

@peterhollender adding you for visibility since this is something you were requesting, but also it is somehting you did in a separate branch. It is re-done more completely here and may clobber some of what was in your other branch, but I'm guessing not more than it was already clobbered by the import overhaul we just merged. Hopefully this is okay and along the lines of what you were looking to get, but let me know and we can be flexible.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -110,10 +110,7 @@ UI is defined in `Resources/UI/<ModuleName>.ui` (Qt Designer XML) and loaded in 
 
 ### OpenLIFULib Key Patterns
 
-**Lazy importing** (`lazyimport.py`): The `openlifu` library and its dependencies are heavy and may not be installed yet. All imports go through lazy-import functions:
-- `openlifu_lz()` — returns the `openlifu` module, installing it first if needed
-- `xarray_lz()`, `bcrypt_lz()`, `threadpoolctl_lz()` — same pattern
-- For IDE type-checking, real imports are guarded under `if TYPE_CHECKING:`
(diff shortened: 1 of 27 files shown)


sadhana-r (approved):
I looked over the code. This approach and the changes look good to me!

PR #577, opened 2026-06-17
Fix hardware test mode using SimulatedLIFUInterface

### LINKED ISSUE #451: Fix openlifu transducer hardware interface mock in SlicerOpenLIFU
The LIFU interface for controlling the transducer should have a more thorough test_mode implementation. Currently, there is a warning displayed in the application when attempting to run in test_mode, but that warning should be removed. Furthermore, there shouldn't be any monkeypatching or mocking of the hardware in SlicerOpenLIFU, so this issue is mainly to track more steps to be done with the test_mode initialization in the sonication control module.

If there's any dead code around previous monkeypatches or hardware mocks, they should be removed from SlicerOpenLIFU.

ebrahimebrahim:
Fixes hardware test mode support in Sonication Control by replacing the old broken SDK `_test_mode` with the newer `openlifu_sdk.ui.SimulatedLIFUInterface`.

image

Main changes:
- Adds a `Test without hardware` checkbox to Sonication Control.
- Removes the old “Test Mode Not Supported” warning and broken private `_test_mode` toggling.

As for the hardware-connected path:
- (No real change) Real hardware initialization still creates `openlifu_sdk.LIFUInterface(run_async=True)`. The removed `TX_test_mode=False` and `HV_test_mode=False` arguments were redundant because those are the SDK defaults.
- (No real change) Real hardware monitoring still uses the existing background asyncio monitor thread path.
- CHANGED initialization flow! The old implicit test-mode toggle is replaced with an explicit `Test without hardware` checkbox plus an initialize/reinitialize button. We no longer auto-initialize on module entry!

Testing update:
- Sonication control finally joins the workflow integration test! We add an end-to-end Sonication Control workflow test that sends a solution to simulated hardware, starts/stops a simulated run, accepts the run completion dialog, and verifies the saved run.

Limitation of `SimulatedLIFUInterface` for testing:
- Passing simulated hardware tests does not prove the real hardware pathway is bug-free. The simulator has a different monitoring model: `SimulatedLIFUInterface.start_monitoring()` creates Qt timers and must run on the GUI thread, while real hardware monitoring runs through the background asyncio monitor thread. This means the test covers the Sonication Control workflow and UI state transitions, but not every threading/device-communication behavior of real hardware.

### DIFF
--- a/OpenLIFULib/OpenLIFULib/dependency_utils.py
+++ b/OpenLIFULib/OpenLIFULib/dependency_utils.py
@@ -30,6 +30,7 @@ def python_requirements_exist() -> bool:
     return (
         importlib.util.find_spec('openlifu') is not None
         and importlib.util.find_spec('openlifu_sdk') is not None
+        and importlib.util.find_spec('openlifu_sdk.ui.simulated_interface') is not None
     )
 
 def check_and_install_python_requirements(prompt_if_found = False) -> None:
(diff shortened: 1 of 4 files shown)


sadhana-r (approved):
Tested it out and it works well! And its great that it is now included as part of the integration tests!

PR #578, opened 2026-06-17
Display RMS distance after ICP registration (#554)

### LINKED ISSUE #554: Add an additional metric that defines the alignment between the 3D mesh and transducer CAD model
Add RMS error output after ICP is run for transducer registration between photogrammetry mesh and transducer model

sadhana-r:
Closes #554 

After ICP registration, the pointwise distance is computed between every point on the moving mesh and the closest point on the fixed reference mesh. In addition to the minimim and maximum distance, we now also show the RMS distance. 

The logic for computing the RMS distance was already added - I just added it to the label display for both skin-segmentation-photoscan and photoscan-transducer registration.


ebrahimebrahim (approved):
Looks great, ty!

PR #580, opened 2026-06-22
Add button to check and install Meshroom dependency

### LINKED ISSUE #542: Add a Meshroom installer to the Login module
Meshroom is an external dependency of SlicerOpenLIFU that has its own separate install process.
We want to more smoothly incorporate it into the OpenLIFU app and extension setup.

The login module is a natural place for this to go, since setting the database path is a part of "initial application setup", just like installing these external dependencies is

We should add an indicator that checks whether Meshroom is available, with a button that can be used to kick off an install if not.

The way to actually do the install needs some investigation. It could be subprocess call or it could involve NSIS scripting. The primary target is to support a smooth experience on _Windows_; on linux or mac it may be okay to direct the user to some instructions or just give them a command they can paste into a terminal.

Special consideration for Meshroom: We may want to first resolve the discussion on whether we want to support 2023, 2025, or both.

sadhana-r:
Addresses #542 

 Adds a button to the login module's "Optional" dependencies section for checking and installing Meshroom 2025.1.0.

Detects Meshroom by looking for meshroom_batch on PATH (the executable required by openlifu). Since meshroom_batch has no --version flag, the version is inferred from the parent directory name (e.g., Meshroom-2025.1.0).

The Meshroom archive is large (~10 GB on Windows, ~13 GB on Linux), so the download runs in a background QProcess with a progress dialog and cancel support - the same approach used for the sample database download. After extraction, the Meshroom directory is added to PATH using the same registry-based approach used for Android Platform Tools. On Windows, a post-install dialog provides step-by-step instructions for enabling NVIDIA GPU acceleration, based on the instructions provided here: https://github.com/OpenwaterHealth/OpenLIFU-python?tab=readme-ov-file#enable-gpu-acceleration

I also added some unit tests to the Login module. 

## For Review

Test it out, and review the overall design. The specifics of the progress bar/download code is mostly borrowed from the sample data functionality

### DIFF
--- a/OpenLIFULib/OpenLIFULib/meshroom_install_cli.py
+++ b/OpenLIFULib/OpenLIFULib/meshroom_install_cli.py
@@ -1,0 +1,151 @@
+import argparse
+import json
+import os
+import sys
+import tarfile
+import urllib.request
+import zipfile
+from pathlib import Path
+
+PROGRESS_LINE_PREFIX = "OPENLIFU_MESHROOM_INSTALL_PROGRESS "
+BYTES_PER_MB = 1024 * 1024
+DOWNLOAD_SOCKET_TIMEOUT_SECONDS = 45
+
+
+class MeshroomInstallCanceled(Exception):
+    pass
+
+
+def _emit_progress_event(event: dict) -> None:
+    print(PROGRESS_LINE_PREFIX + json.dumps(event), flush=True)
+
+
+def _progress_callback(message: str, value: int, maximum: int) -> None:
+    _emit_progress_event({"message": message, "value": value, "maximum": maximum})
+
+
+def _raise_if_canceled(cancel_callback):
+    if cancel_callback is not None and cancel_callback():
+        raise MeshroomInstallCanceled("Meshroom installation canceled.")
+
+
+def _download_archive(url: str, archive_path: Path, cancel_callback) -> None:
+    _raise_if_canceled(cancel_callback)
+    _progress_callback("Downloading Meshroom. This is a large download and may take several minutes.", 0, 0)
+
+    with urllib.request.urlopen(url, timeout=DOWNLOAD_SOCKET_TIMEOUT_SECONDS) as response:
+        total_size = int(response.headers.get("Content-Length") or 0)
+        total_mb = max(1, total_size // BYTES_PER_MB) if total_size else 0
+        downloaded_size = 0
+        last_reported_mb = -1
+
+        with archive_path.open("wb") as f:
+            while True:
+                _raise_if_canceled(cancel_callback)
+                chunk = response.read(BYTES_PER_MB)
+                if not chunk:
+                    break
+                _raise_if_canceled(cancel_callback)
+                f.write(chunk)
+                downloaded_size += len(chunk)
+                downloaded_mb = downloaded_size // BYTES_PER_MB
+                if downloaded_mb == last_reported_mb:
+                    continue
+                if total_mb:
+                    reported_mb = min(total_mb, downloaded_mb)
+                    _progress_callback(
+                        f"Downloading Meshroom ({reported_mb} of {total_mb} MB)...",
+                        reported_mb,
+                        total_mb,
+                    )
+                else:
+                    _progress_callback(
+                        f"Downloading Meshroom ({downloaded_mb} MB downloaded)...",
+                        0,
+                        0,
+                    )
+                last_reported_mb = downloaded_mb
+
+
+def _extract_zip(archive_path: Path, extraction_dir: Path, cancel_callback) -> None:
+    _raise_if_canceled(cancel_callback)
+    _progress_callback("Extracting Meshroom...", 0, 0)
+    extraction_dir_resolved = extraction_dir.resolve()
+    with zipfile.ZipFile(archive_path) as archive:
+        members = archive.infolist()
+        total_members = len(members)
+        report_every = max(1, total_members // 100)
+        for i, member in enumerate(members, start=1):
+            _raise_if_canceled(cancel_callback)
+            destination_path = (extraction_dir / member.filename).resolve()
+            if os.path.commonpath([str(extraction_dir_resolved), str(destination_path)]) != str(extraction_dir_resolved):
+                raise RuntimeError(f"Archive contains an unsafe path: {member.filename}")
+            archive.extract(member, extraction_dir)
+            if i == total_members or i % report_every == 0:
+                _progress_callback(
+                    f"Extracting Meshroom ({i} of {total_members} files)...",
+                    i,
+                    total_members,
+                )
+
+
+def _extract_tarball(archive_path: Path, extraction_dir: Path, cancel_callback) -> None:
+    _raise_if_canceled(cancel_callback)
+    _progress_callback("Extracting Meshroom...", 0, 0)
+    extraction_dir_resolved = extraction_dir.resolve()
+    with tarfile.open(archive_path, "r:gz") as archive:
+        members = archive.getmembers()
+        total_members = len(members)
+        report_every = max(1, total_members // 100)
+        for i, member in enumerate(members, start=1):
+            _raise_if_canceled(cancel_callback)
+            destination_path = (extraction_dir / member.name).resolve()
+            if os.path.commonpath([str(extraction_dir_resolved), str(destination_path)]) != str(extraction_dir_resolved):
+                raise RuntimeError(f"Archive contains an unsafe path: {member.name}")
+            archive.extract(member, extraction_dir, filter="data")
+            if i == total_members or i % report_every == 0:
+                _progress_callback(
+                    f"Extracting Meshroom ({i} of {total_members} files)...",
+                    i,
+                    total_members,
+                )
+
+
+def main(argv=None) -> int:
+    parser = argparse.ArgumentParser(description="Download and extract Meshroom.")
+    parser.add_argument("--destination", required=True, help="Directory where Meshroom will be extracted.")
+    parser.add_argument("--work-dir", required=True, help="Temporary work directory for the archive download.")
+    parser.add_argument("--archive-url", required=True, help="URL of the Meshroom archive to download.")
+    parser.add_argument("--cancel-file", default="", help="Sentinel file; if it exists, installation exits as canceled.")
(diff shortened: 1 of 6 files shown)


OpenLIFULib/OpenLIFULib/meshroom_install_gui.py
```diff
@@ -0,0 +1,345 @@
...
+import slicer
+
+from OpenLIFULib.meshroom_install_cli import PROGRESS_LINE_PREFIX
+
+logger = logging.getLogger(__name__)
+
+MESHROOM_VERSION = "2025.1.0"
+MESHROOM_WINDOWS_URL = "https://zenodo.org/records/16887472/files/Meshroom-2025.1.0-Windows.zip"
+MESHROOM_LINUX_URL = "https://zenodo.org/records/16887472/files/Meshroom-2025.1.0-Linux.tar.gz"
+MESHROOM_EXTRACTED_DIR_NAME = "Meshroom-2025.1.0"
```
sadhana-r, line 19: The meshroom version is defined by the `MESHROOM_VERSION` variable and these download links. So if updating meshroom in the future, only these variables would need to be updated.

OpenLIFULogin/OpenLIFULogin.py
```diff
@@ -1144,6 +1149,132 @@ def _installADBLinux(self) -> None:
...
+
+        if meshroom_path:
+            meshroom_dir = Path(meshroom_path).parent
+            dir_name = meshroom_dir.name
+            version_str = dir_name[len("Meshroom-"):] if dir_name.startswith("Meshroom-") else "installed"
+            self.ui.installMeshroomPushButton.setEnabled(False)
+            self.ui.installMeshroomPushButton.setText(f"Meshroom {version_str}")
+        else:
+            self.ui.installMeshroomPushButton.setEnabled(True)
+            self.ui.installMeshroomPushButton.setText("Install Meshroom 2025")
```
sadhana-r: The button text would also need to be updated if updating meshroom.

OpenLIFULogin/OpenLIFULogin.py
```diff
@@ -1144,6 +1149,132 @@ def _installADBLinux(self) -> None:
...
+                    "To enable GPU acceleration for Meshroom:\n\n"
+                    "1. Open the NVIDIA Control Panel from the Start menu or system tray.\n"
+                    "2. In the left sidebar under '3D Settings', click 'Manage 3D settings'.\n"
+                    "3. Go to the 'Program Settings' tab.\n"
+                    "4. Click 'Add', then browse to and select meshroom_batch.exe\n"
+                    f"   from the folder where Meshroom was installed ({meshroom_bin_path}).\n"
+                    "5. Under 'Select the preferred graphics processor for this program',\n"
+                    "   choose 'High-performance NVIDIA processor'.\n"
+                    "6. Click 'Apply'."
+                )
```
ebrahimebrahim, line 1282: Do these instructions work on modern windows? My AI reviewer found [this nvidia doc](https://www.nvidia.com/content/Control-Panel-Help/vLatest/en-us/mergedProjects/3D%20Settings/Setting_the_Preferred_Graphics_Processor.htm) stating the following:

> Beginning with Windows 10 20H1, the Windows OS assigns the graphics processor to use for various applications, and overrides any settings made in the NVIDIA Control Panel > Manage 3D Settings > Preferred graphics processor setting for that application.
> 
> - You can change the GPU preference that the OS will use for an application from the Windows Graphics Settings.
> - If the OS does not have a GPU preference for an application, then you can specify the preferred graphics processor to use from the NVIDIA Control Panel > Manage 3D Settings > Preferred graphics processor setting.

So maybe the instructions need to first direct the user to do something with Windows Graphics Settings, and then if those settings don't exist at the OS level the NVIDIA control panel becomes the appropriate fallback

OpenLIFULogin/OpenLIFULogin.py
```diff
@@ -1144,6 +1149,132 @@ def _installADBLinux(self) -> None:
...
+        pixmap = slicer.app.style().standardIcon(icon_name).pixmap(qt.QSize(16, 16))
+        self.ui.meshroomStatusIcon.setPixmap(pixmap)
+        self.ui.meshroomStatusIcon.setText("")
+
+        if meshroom_path:
+            meshroom_dir = Path(meshroom_path).parent
+            dir_name = meshroom_dir.name
+            version_str = dir_name[len("Meshroom-"):] if dir_name.startswith("Meshroom-") else "installed"
+            self.ui.installMeshroomPushButton.setEnabled(False)
+            self.ui.installMeshroomPushButton.setText(f"Meshroom {version_str}")
```
ebrahimebrahim: Since you are getting the version_str here, and you define the required version str in `meshroom_install_gui.MESHROOM_VERSION`, you could compare them and not disable the install meshroom button in the event that the version is not the required one. Right now if the meshroom_batch exists but the installed meshroom is the wrong version it shows green and disables the button.

OpenLIFULogin/OpenLIFULogin.py
ebrahimebrahim: Here is a suggested set of post-install instructions for modern windows (by AI); it needs testing still:

> 1. Open **Settings → System → Display → Graphics**. On some Windows 10 versions this is called **Graphics settings**.
> 2. Under **Custom options for apps** or **Choose an app to set preference**, add a **Desktop app**.
> 3. Browse to the Meshroom installation directory and select `meshroom_batch.exe`, since this is the executable OpenLIFU launches.
> 4. Select the newly added application and click **Options**.
> 5. Choose **High performance**, confirming that the listed GPU is the NVIDIA GPU, then click **Save**.
> 6. Restart OpenLIFU before running a reconstruction.

On older Windows versions, or if Windows has no GPU preference for the application, the **NVIDIA Control Panel → Manage 3D settings → Program Settings** approach is a good fallback.

OpenLIFULib/OpenLIFULib/meshroom_install_cli.py
```diff
@@ -0,0 +1,151 @@
...
+    with tarfile.open(archive_path, "r:gz") as archive:
+        members = archive.getmembers()
+        total_members = len(members)
+        report_every = max(1, total_members // 100)
+        for i, member in enumerate(members, start=1):
+            _raise_if_canceled(cancel_callback)
+            destination_path = (extraction_dir / member.name).resolve()
+            if os.path.commonpath([str(extraction_dir_resolved), str(destination_path)]) != str(extraction_dir_resolved):
+                raise RuntimeError(f"Archive contains an unsafe path: {member.name}")
+            archive.extract(member, extraction_dir, filter="data")
```
ebrahimebrahim: I got an error on linux:

Image

The problem seems to be something in the meshroom linux archive. The archive contains an absolute symbolic link related to the "mrSegmentation" component of Meshroom (something OpenLIFU doesn't use), and the tarfile data extractor is rejecting it (the default behavior of the extractor is to reject absolute symbolic links because a malicious archive can use those to make software unknowingly interact with sensitive files on someone's system). In our case we are already trusting meshroom: we can assume it is not a malicious archive. A way to bypass that security measure is:

```suggestion
            archive.extract(member, extraction_dir, filter="tar")
```

See https://docs.python.org/3/library/tarfile.html#tarfile-extraction-filter on the extraction filter options.

I tested this out myself and it did resolve the issue.

OpenLIFULogin/OpenLIFULogin.py
```diff
@@ -1144,6 +1149,132 @@ def _installADBLinux(self) -> None:
...
+                    "5. Under 'Select the preferred graphics processor for this program',\n"
+                    "   choose 'High-performance NVIDIA processor'.\n"
+                    "6. Click 'Apply'."
+                )
+            else:
+                slicer.util.infoDisplay(
+                    f"Meshroom {MESHROOM_VERSION} installed successfully.\n\n"
+                    f"To persist across sessions, add the following to your shell profile:\n"
+                    f"    export PATH=\"$PATH:{meshroom_bin_path}\""
+                )
```
ebrahimebrahim, line 1286: Instead of having the user update the PATH, why don't we just save the installed path to a QSettings. The settings key could be `"OpenLIFU/meshroomExecutablePath"`:

```python
MESHROOM_EXECUTABLE_SETTINGS_KEY = "OpenLIFU/meshroomExecutablePath"

qt.QSettings().setValue(
    MESHROOM_EXECUTABLE_SETTINGS_KEY,
    str(install_dir / MESHROOM_EXTRACTED_DIR_NAME / "meshroom_batch"),
)
```

You could do this for windows too probably!

Then you could have a utility function

```python
def restore_meshroom_path() -> Path | None:
    settings = qt.QSettings()
    saved_path = settings.value(MESHROOM_EXECUTABLE_SETTINGS_KEY, "")
    if not saved_path:
        return None

    executable = Path(str(saved_path))
    if not executable.is_file():
        settings.remove(MESHROOM_EXECUTABLE_SETTINGS_KEY)
        return None

    meshroom_dir = str(executable.parent)
    path_entries = os.environ.get("PATH", "").split(os.pathsep)

    if meshroom_dir not in path_entries:
        os.environ["PATH"] = os.pathsep.join([meshroom_dir, *path_entries])

    return executable
```

which you call in `_checkMeshroomStatus` (before the `shutil.which()`)
and also in the tranducer localization module setup (in case the login module wasn't visited before going to the module that needs meshroom)

ebrahimebrahim (changes requested):
The progress bars and general process work really smoothly!

Did you test the installation yourself on Windows? From a non-installed state?

Did you try out the instructions to enable GPU acceleration? What version of windows?

OpenLIFULogin/OpenLIFULogin.py
sadhana-r: Good suggestion! I have added the save/restore from QSettings functionality for both windows and linux/mac. So `_checkMeshroomStatus` first checks the QSettings, then the PATH variable (if not found in QSettings), and if the meshroom path is not specified in either location, the user is then prompted to reinstall. 

After meshroom is installed, the path is saved to QSettings. For Windows, I still write to the windows registry so the meshroom instalation is added to PATH. I removed the recommendation to add to path on Linux.

ebrahimebrahim (approved):

PR #581, opened 2026-07-08
Document our new release process

ebrahimebrahim:
This formalizes the process extension releases.

This is a change from how we used to do releases off of the main line!
After this, there will be release branches.
Typically we would not create release candidates.
The extension is developed quickly and patched quickly, and there is no dependency freeze period for it like we are planning to do with the desktop application.


peterhollender (approved):
Looks good and consistent with our plan

PR #645, opened 2026-08-08
Fix custom app page lifecycle guards (#627)

HussainAther:
Fixes #627.

## Summary

This fixes the custom-app page lifecycle regression where `dataChanged` handlers could short-circuit even while a page was visible.

In vanilla Slicer, page widgets can rely on `self.parent.isEntered`, but in the custom desktop app pages are reparented into a plain `QWidget`, so that attribute is unavailable. The previous guard therefore evaluated false and prevented page refresh logic from running.

### Changes

- add a shared `page_is_entered(widget)` helper in `OpenLIFULib.util`
- use `parent.isEntered` when available in vanilla Slicer
- fall back to each widget's `_entered` state in the custom app
- set `_entered` in affected widgets during `enter()` / `exit()`
- replace direct `self.parent.isEntered` and `getattr(..., False)` lifecycle checks with the shared helper
- update `dataChanged` and scene-close guards across the affected pages

## Validation

- `git diff --check` passes
- targeted Python compilation passes for the modified modules that are compilable on the current base
- verified there are no remaining uses of:
  - `self.parent.isEntered`
  - `getattr(self.parent, "isEntered", False)`

`OpenLIFUTransducerLocalization.py` contains a pre-existing unrelated f-string syntax error in the current base snapshot, so full-file compilation for that module is not possible. The changes in this PR do not touch the failing line.

## Scope

This PR is intentionally limited to the page lifecycle / visibility guard regression described in #627. Follow-up behavior around PNP rendering, opacity/window-level state, and other solution-state issues can be handled separately if they remain reproducible after this fix.

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -51,6 +51,7 @@ from OpenLIFULib.util import (
     display_errors,
     ensure_list,
     replace_widget,
+    page_is_entered,
 )
 from OpenLIFULib.volume_thresholding import load_volume_and_threshold_background
 from OpenLIFULib.virtual_fit_results import (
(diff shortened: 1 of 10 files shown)


ebrahimebrahim:

PR #646, opened 2026-08-09
Refactor clear_session teardown ordering (#597)

HussainAther:
gh pr create \
  --repo OpenwaterHealth/SlicerOpenLIFU \
  --base main \
  --head HussainAther:fix/597-clear-session-snapshot \
  --title "Refactor clear_session teardown ordering (#597)" \
  --body "$(cat <<'EOF'
Fixes #597.

## Summary

- snapshot session metadata and scene-node references before detaching `loaded_session`
- clear `loaded_session` before performing session teardown
- avoid accessing the detached parameter-pack-backed session during cleanup
- guarantee `session_loading_unloading_in_progress` is reset with `try/finally`
- add regression coverage to the existing full workflow verifying that the session remains detached after cleanup

## Validation

- `python -m py_compile OpenLIFUData/OpenLIFUData.py OpenLIFUHome/OpenLIFUHome.py`
- `git diff --check`
- regression check runs after the existing OpenLIFU full workflow

The teardown now operates only on values snapshotted before `loaded_session` is cleared, preventing cleanup callbacks from interacting with the detached session wrapper.
EOF
)"

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1678,7 +1678,7 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
         for f in self._on_subject_changed_callbacks:
             f(self._subject)
 
-    def clear_session(self, clean_up_scene:bool = True) -> None:
+    def clear_session(self, clean_up_scene: bool = True) -> None:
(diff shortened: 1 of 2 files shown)


ebrahimebrahim:

PR #649, opened 2026-08-09
Avoid loading all users during login (#596)

HussainAther:
Addresses part of #596.

## Summary

- replace `load_all_users()` in the interactive login path with a targeted `load_user(user_id)`
- preserve the same unsuccessful-login behavior for unknown users and incorrect passwords
- catch the documented `FileNotFoundError` raised when a user ID does not exist

## Why

The login path only needs to authenticate a single requested user, but it currently calls `load_all_users()`, which deserializes every user in the database and emits a `Loaded User ...` log entry for each account.

Using `load_user(user_id)` avoids that unnecessary work and reduces log noise during ordinary login.

## Validation

- `python -m py_compile OpenLIFULogin/OpenLIFULogin.py`
- `git diff --check`
- verified `Database.load_user()` raises `FileNotFoundError` when the requested user does not exist
- remaining `load_all_users()` calls are intentionally left unchanged because they serve different multi-user UI/state paths

## Behavior

- valid username + valid password: login succeeds
- valid username + invalid password: unsuccessful login
- unknown username: unsuccessful login

### DIFF
--- a/OpenLIFULogin/OpenLIFULogin.py
+++ b/OpenLIFULogin/OpenLIFULogin.py
@@ -778,12 +778,21 @@ class OpenLIFULoginWidget(ScriptedLoadableModuleWidget, VTKObservationMixin, Gui
         if not returncode:
             return
 
-        users = get_cur_db().load_all_users()
         import bcrypt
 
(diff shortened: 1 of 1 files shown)


ebrahimebrahim:

PR #650, opened 2026-08-09
Enable 2D visibility for transducer models (#607)

HussainAther:
Fixes #607.

## Summary

- enable 2D slice visibility for the live transducer model
- enable 2D slice visibility for the transducer body model
- enable 2D slice visibility for the registration surface model
- enable 2D slice visibility for the cloned virtual-fit preview model

## Why

Transducer-related model nodes were visible in the 3D view but their slice intersections were disabled by default.

Enabling `Visibility2D` allows the transducer geometry to appear as outlines in the Red, Green, and Yellow slice views during target placement, virtual fitting, and transducer-localization review.

## Validation

- `python -m py_compile OpenLIFULib/OpenLIFULib/transducer.py`
- `git diff --check`
- change is limited to the four transducer-related model display nodes

## Manual regression

1. Load a session with a transducer.
2. Open the Red / Green / Yellow slice views.
3. Confirm the live transducer model appears where it intersects each slice.
4. Confirm body / registration-surface geometry also appears when available.
5. Display a virtual-fit preview and confirm its cloned model also appears in the slice views.
6. Confirm normal 3D visibility behavior is unchanged.


ebrahimebrahim:

PR #651, opened 2026-08-09
Allow VF-only workflow outside guided mode (#620)

HussainAther:
Fixes #620.

## Summary

- allow the Transducer Localization workflow to proceed without an approved TT result when running outside guided mode
- detect whether the loaded session has at least one approved virtual-fit result
- preserve the existing TT requirement in guided mode

## Why

The Sonication Planner supports a VF-only pre-solution workflow, but Transducer Localization previously blocked Next unless an approved transducer-tracking result existed.

That prevented users in non-guided mode from reaching the pre-solution path even when the session already had an approved virtual-fit result.

## Behavior

- approved TT present → proceed as before
- no approved TT + approved VF + non-guided mode → proceed to compute a pre-solution
- no approved TT + guided mode → remain blocked
- no approved TT and no approved VF → remain blocked

## Validation

- `git diff --check`
- change is limited to `OpenLIFUTransducerLocalization/OpenLIFUTransducerLocalization.py`
- existing guided-mode TT requirement is preserved

## Note

Whole-file `py_compile` is currently blocked by a pre-existing malformed f-string elsewhere in `OpenLIFUTransducerLocalization.py`, unrelated to this change.

### DIFF
--- a/OpenLIFUTransducerLocalization/OpenLIFUTransducerLocalization.py
+++ b/OpenLIFUTransducerLocalization/OpenLIFUTransducerLocalization.py
@@ -73,7 +73,11 @@ from OpenLIFULib.transducer_tracking_wizard_utils import (
 from OpenLIFULib.user_account_mode_util import UserAccountBanner
 from OpenLIFULib.util import add_slicer_log_handler, BusyCursor, get_cloned_node, replace_widget, display_errors
 from OpenLIFULib.notifications import notify
-from OpenLIFULib.virtual_fit_results import get_virtual_fit_approval_for_target, get_approval_from_virtual_fit_result_node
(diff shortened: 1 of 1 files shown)


ebrahimebrahim:
#620 and many other issues were created during Peter's work on a separate branch working towards a c2 of the application with lots of workflow changes. That branch already has fixes and for many of these issues and just needs to be rebased and reworked. So we cannot integrate this PR at this time, as it will unnecessarily complicate the upcoming rebase of a lot of work that has already been done. Sorry for any wasted effort time! Many of these issues were just not up for grabs, and we did not clearly indicate that. In the future it will be better to check first whether an issue is available and have it assigned to you, as there's often a lot of work going on at Openwater that isn't always outwardly tracked by github assignments.

PR #652, opened 2026-08-09
Clear empty session photoscan affiliations (#619)

HussainAther:
Fixes #619.

## Summary

- always refresh the loaded session's affiliated photoscan state from the database
- explicitly set the session's affiliated photoscans to an empty dictionary when the database reports none
- prevent photoscan affiliation state from surviving a switch to a session with no photoscans

## Root cause

`update_photoscans_affiliated_with_loaded_session()` only called
`set_affiliated_photoscans()` when the database returned a non-empty
photoscan dictionary.

When switching from a session with affiliated photoscans to a session with
none, the empty result therefore did not update the session wrapper. This
could leave stale photoscan affiliation state visible to Transducer
Localization.

`set_affiliated_photoscans({})` already correctly clears the wrapper's
affiliated photoscan dictionary, so the refresh should always call it.

## Validation

- `python -m py_compile OpenLIFUData/OpenLIFUData.py`
- `git diff --check`
- change limited to `OpenLIFUData/OpenLIFUData.py`

## Manual regression

1. Load session A with an affiliated photoscan.
2. Exit session A.
3. Load session B with no affiliated photoscans.
4. Navigate to Transducer Localization.
5. Confirm session A's photoscan is no longer presented as affiliated with session B.

### DIFF
--- a/OpenLIFUData/OpenLIFUData.py
+++ b/OpenLIFUData/OpenLIFUData.py
@@ -1882,8 +1882,7 @@ class OpenLIFUDataLogic(ScriptedLoadableModuleLogic):
         
         # Keep track of any photocollections associated with the session
         affiliated_photocollections = get_cur_db().get_photocollection_reference_numbers(subject_id, session_id)
-        if affiliated_photocollections:
-            loaded_session.set_affiliated_photocollections(affiliated_photocollections)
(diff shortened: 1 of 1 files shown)


ebrahimebrahim:

PR #654, opened 2026-08-10
Fix stale photoscan preview cache

ebrahimebrahim:
Closes #653

For review: I tested this before and after the fix, and it did fix the issue. So a quick look at the code suffices


sadhana-r (approved):
lgtm!

